PR #113, opened 2025-12-19
feat: add transaction override support for replay command

Troublor:
## Summary
- Add CLI flags to override transaction fields when replaying transactions from RPC
- Supports overriding: gas limit, value, and input data
- Input can be provided as hex string or loaded from file

## New CLI Options
```
Transaction Override Options:
  --override.gas-limit <GAS>      Override transaction gas limit
  --override.value <WEI>          Override transaction value (in wei)
  --override.input <HEX>          Override transaction input data (hex string)
  --override.input-file <FILE>    Override transaction input data from file
```

### DIFF
--- a/bin/mega-evme/README.md
+++ b/bin/mega-evme/README.md
@@ -203,6 +203,28 @@ mega-evme replay 0x1234...txhash --trace
 | ------------- | --------------------- | --------------------------------- |
 | `--rpc <URL>` | http://localhost:8545 | RPC URL to fetch transaction from |
 
+#### Transaction Override Options
+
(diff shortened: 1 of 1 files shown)


bin/mega-evme/src/common/tx_override.rs
```diff
@@ -0,0 +1,142 @@
+//! Transaction override support for mega-evme replay command.
+//!
+//! This module provides the ability to override transaction fields when replaying
+//! transactions from RPC.
```
megakabi: 新引入的这个参数文档里没有，感觉应该加一下说明

Troublor: Done

PR #131, opened 2026-01-13
fix: avoid panic when log file cannot be created

kinderjoypresents:
## Summary
This PR prevents mega-evme from panicking when --log.file points to an invalid/unwritable path.

## Changes
- Replace expect() on std::fs::File::create with graceful error handling.
- On failure, print a short message to stderr and fall back to stderr logging.

## Motivation
Passing an invalid --log.file path should not crash the CLI; it should continue running with stderr logging.


Troublor:
@kinderjoypresents The CI workflow is currently misconfigured for external PRs. We will fix that soon.

PR #151, opened 2026-01-29
fix(t8n): implement signature recovery for transactions

huahuayu:
This PR implements signature recovery for transactions in the `t8n` tool. Previously, it only supported transactions with a `secret_key` provided in the input. Now, it correctly recovers the sender address from the `v, r, s` signature components using `MegaTxEnvelope::recover_signer`.

Changes:
- Added `MegaTxEnvelope` to imports in `cmd.rs`.
- Implemented recovery logic for Legacy, EIP-2930, EIP-1559, and EIP-7702 transactions.


Troublor:
Thanks, `t8n` was meant to be used to run ethereum-tests, but it has been broken for months. I will revisit this next month

PR #180, opened 2026-03-04
feat: Rex4 extend per-frame budgets to compute/data/kv limits

RealiCZ:
### Summary

  This PR extends Rex4 per-frame budget enforcement from state growth to all additional-limit dimensions: compute gas, data size, and KV updates.

### Key changes:

  - Enable Rex4 per-frame budget checks in `ComputeGasTracker`, `DataSizeTracker`, and `KVUpdateTracker `using the same remaining * 98 / 100 forwarding rule.
  - Keep compute gas semantics consistent with existing design:
      - Per-frame exceed in Rex4 is frame-local (Revert) as an early-termination guardrail.
      - Compute gas remains persistent across reverts.
      - Gas detention remains TX-level.
  - Fix frame budget forwarding overflow risk in FrameLimitTracker by using safe intermediate arithmetic for the
    98/100 calculation.
  - Update limit-related inline documentation/comments in `limit.rs` to reflect current runtime behavior.
 
### Test Plan
New test file `tests/rex4/frame_limits`.rs with 30+ tests covering:

  - Basic per-frame budget allocation for DataSize / KVUpdate / ComputeGas (98/100 rule, boundary cases)
  - Parent-usage-aware forwarding (remaining * 98 / 100) and budget propagation correctness
  - Frame-local absorption semantics in Rex4 (revert not halt, parent continues)
  - Resource-specific revert effects:
      - DataSize / KVUpdate discard child frame usage on revert
      - ComputeGas remains persistent across child revert
  - Revert data encoding and ABI decoding (MegaLimitExceeded(kind, limit)) for kinds 0/1/2
  - Nested frame scenarios (grandchild progressive reduction, sibling independence, child exceed then sibling success)
  - Top-level exceed behavior in Rex4 for DataSize/KVUpdate (frame-local revert path)
  - Backward compatibility checks (REX3 has no per-frame limits for these dimensions, TX-level exceed halts)
  - Explicit Rex4 compute TX-level exceed halt case (separate from frame-local revert path)

### DIFF
--- a/crates/mega-evm/src/limit/compute_gas.rs
+++ b/crates/mega-evm/src/limit/compute_gas.rs
@@ -121,6 +121,11 @@ impl TxRuntimeLimit for ComputeGasTracker {
             if frame_check.exceeded_limit() {
                 return frame_check;
             }
+            // Do not early-return on frame WithinLimit:
+            // 1) pre-frame intrinsic compute gas is recorded in `tx_entry`, outside current frame
+            //    budget;
+            // 2) `detained_limit` can be lowered at runtime by volatile-data access.
+            // So TX-level detained check must still run even when frame check is within limit.
         }
         // TX-level detained check (all specs): total usage vs effective limit (min of tx/detained).
         let limit = self.tx_limit();


crates/mega-evm/src/limit/compute_gas.rs
```diff
@@ -92,10 +106,18 @@ impl TxRuntimeLimit for ComputeGasTracker {
...
+    /// In Rex4+, checks the per-frame budget first (`frame_local`: true on exceed), then falls
+    /// through to the TX-level detained check. Gas detention is always TX-level (`frame_local`:
+    /// false) across all specs — accessing volatile data caps the whole transaction.
     #[inline]
     fn check_limit(&self) -> LimitCheck {
+        if self.rex4_enabled {
+            let frame_check = self.frame_tracker.exceeds_current_frame_limit(LimitKind::ComputeGas);
+            if frame_check.exceeded_limit() {
+                return frame_check;
+            }
```
Troublor: When `REX4` is enabled, we should not check tx-level limits, similar to state growth limit. Please change this logic.

Troublor (commented):
some comments below

crates/mega-evm/tests/rex4/frame_limits.rs
```diff
@@ -0,0 +1,941 @@
...
+    assert!(result.result.is_halt(), "REX3 should halt on TX-level KV exceed");
+    assert!(matches!(
+        result.result,
+        ExecutionResult::Halt { reason: MegaHaltReason::KVUpdateLimitExceeded { .. }, .. }
+    ));
+}
+
+// ============================================================================
+// COMPUTE GAS PER-FRAME LIMITS
+// ============================================================================
```
Troublor: `data_size` and `kv_update` both have `test_*_top_level_exceed_is_frame_local_revert` tests. Compute gas doesn't have an equivalent. Since compute gas is "always persistent," a top-level frame exceed would raise the TX-level check (which halts rather than reverts). If that's the expected behavior, consider adding a comment here explaining why the top-level frame test is absent to keep the suite symmetric.

Troublor (commented):
Two minor suggestions for test suite consistency.

crates/mega-evm/src/limit/compute_gas.rs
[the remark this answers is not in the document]
RealiCZ: For `compute_gas`, we cannot early-return on frame `WithinLimit` for two reasons:

  1. initial/pre-frame compute gas is recorded in `tx_entry`, which is outside the current frame budget;
  2. `detained_limit` can be lowered dynamically at runtime, while frame budgets are propagated from parent-frame remaining budget, so they cannot replace the TX-level detained check.

PR #181, opened 2026-03-05
feat: Rex4 MegaLimitControl system contract

RealiCZ:
### Summary
  - Introduce the `MegaLimitControl` system contract at `0x6342…0005` (Rex4), exposing `remainingComputeGas()` for contracts and direct transaction calls.
  - Implement system-call interception for `CALL/STATICCALL` to `LIMIT_CONTROL_ADDRESS` in Rex4 and return ABI-encoded uint64.
  - Define `remainingComputeGas()` as **current-call remaining compute gas** in Rex4, not transaction-level detained remaining.
 -  Reject non-zero ETH value for this view-style system method with `NonZeroTransfer()`.
 - Align interception scope with existing system-contract behavior: `DELEGATECALL` and `CALLCODE` are intentionally not intercepted.
 - Add Solidity interface and implementation (`IMegaLimitControl.sol`, `MegaLimitControl.sol`),
- Update `specs/Rex4.md` to document contract address, current-call remaining semantics, interception scope, and expected` DELEGATECALL/CALLCODE` non-interception behavior.
---
  ### Test plan

- [x] `cargo check -p mega-evm passes`
- [x] `cargo test -p mega-evm --tests` — all 485 tests pass
- [x] `cargo fmt --all --check passes`
- [x] `cargo clippy --workspace passes`
- [x] `cargo sort --check --workspace passes`

### DIFF
--- a/crates/mega-evm/tests/rex4/limit_control.rs
+++ b/crates/mega-evm/tests/rex4/limit_control.rs
@@ -340,21 +340,21 @@ fn test_remaining_compute_gas_sequential_queries_decrease() {
     let code = BytecodeBuilder::default()
         // First query
         .mstore(0x0, REMAINING_COMPUTE_GAS_SELECTOR)
-        .push_number(32_u64)   // retSize
+        .push_number(32_u64) // retSize
         .push_number(0x20_u64) // retOffset — store first result at 0x20
-        .push_number(4_u64)    // argsSize
-        .push_number(0_u64)    // argsOffset
-        .push_number(0_u64)    // value
+        .push_number(4_u64) // argsSize
+        .push_number(0_u64) // argsOffset
+        .push_number(0_u64) // value
         .push_address(LIMIT_CONTROL_ADDRESS)
         .push_number(100_000_u64)
         .append(CALL)
         .append(POP)
         // Second query (selector still at 0x00)
-        .push_number(32_u64)   // retSize
+        .push_number(32_u64) // retSize
         .push_number(0x40_u64) // retOffset — store second result at 0x40
-        .push_number(4_u64)    // argsSize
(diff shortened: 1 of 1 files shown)


crates/mega-evm/src/limit/compute_gas.rs
```diff
@@ -66,6 +66,24 @@ impl ComputeGasTracker {
...
+    /// If no frame exists yet (direct TX → system contract), returns the TX-level
+    /// remaining which accounts for intrinsic compute gas.
+    /// In pre-Rex4, falls back to TX-level remaining compute gas.
+    ///
+    /// Called during system contract interception, before the callee's frame is pushed.
+    /// At that point `frame_stack.last()` is the caller's frame, so
+    /// `current_frame_remaining()` gives the caller's remaining compute gas.
+    pub(crate) fn current_call_remaining(&self) -> u64 {
+        if self.rex4_enabled {
+            self.frame_tracker.current_frame_remaining()
```
Troublor: `current_call_remaining()` returns `frame_tracker.current_frame_remaining()` which ignores `detained_limit`. But execution halts when `tx_usage() > min(tx_limit, detained_limit)`, so the effective remaining is `min(frame_remaining, detained_remaining)`.

Since this API is meant to tell contracts how much compute gas they can actually use, it should cap with the TX-level detained remaining. Otherwise, after a TIMESTAMP access triggers gas detention, the returned value will overstate what the contract can actually consume before halting.

```suggestion
            self.frame_tracker.current_frame_remaining()
                .min(self.tx_limit().saturating_sub(self.tx_usage()))
```

The test `test_remaining_compute_gas_not_clamped_by_detention_limit` asserts the current (uncapped) behavior — it would need to be updated to verify the capped behavior instead.

PR #184, opened 2026-03-09
feat: Rex4 relative gas detention cap for volatile data access

RealiCZ:
## Summary

Rex4 changes gas detention from an absolute cap to a relative cap.
- Previously, when a transaction accessed volatile data (block env, beneficiary, oracle), the compute gas limit was capped at an absolute value (e.g., 20M).
- If the transaction had already consumed >20M compute gas before the access, it halted immediately — even though the intent is to limit post-access computation, not penalize prior work.
- Now in Rex4+, the detention cap is relative to current usage at the access point:` effective_limit = usage_at_access + cap`. A transaction that used 25M gas before accessing TIMESTAMP gets an effective limit of 45M (25M + 20M), allowing 20M more gas after access. Pre-Rex4 specs retain the absolute-cap behavior for backward compatibility.

### Changes

  - `ComputeGasTracker::set_detained_limit()`: Computes relative cap (`tx_usage + cap`) for REX4+, absolute cap for pre-REX4.
  - `execution_result():` Detects detention-caused halts by comparing against detained_limit instead of the raw cap. Reports detained_limit in `VolatileDataAccessOutOfGas`.limit so actual > limit always holds.

### Test plan

  - `cargo test -p mega-evm -- gas_detention` — 5 new REX4 relative cap tests pass
  - `cargo test -p mega-evm — full` test suite (488 tests) passes

### DIFF
--- a/crates/mega-evm/src/evm/execution.rs
+++ b/crates/mega-evm/src/evm/execution.rs
@@ -35,7 +35,7 @@ use revm::{
 use crate::{
     constants, dispatch_system_contract_interceptors, is_mega_system_transaction,
     sent_from_mega_system_address, ExternalEnvTypes, HostExt, MegaContext, MegaEvm, MegaHaltReason,
-    MegaInstructions, MegaSpecId, MegaTransactionError, TxRuntimeLimit, MEGA_SYSTEM_ADDRESS,
+    MegaInstructions, MegaSpecId, MegaTransactionError, MEGA_SYSTEM_ADDRESS,
     MEGA_SYSTEM_TRANSACTION_SOURCE_HASH,
 };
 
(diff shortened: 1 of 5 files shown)


crates/mega-evm/src/evm/execution.rs
```diff
@@ -438,12 +438,17 @@ where
...
+                    if actual > detained_limit &&
+                        detained_limit < additional_limit.compute_gas.frame_tx_limit()
+                    {
                         return MegaHaltReason::VolatileDataAccessOutOfGas {
                             access_type,
-                            limit: volatile_compute_gas_limit,
+                            limit: detained_limit,
                             actual,
                         };
                     }
```
Troublor: The detention halt detection here feels architecturally split: `ComputeGasTracker` already knows both `detained_limit` and the base TX limit internally, yet we expose them via `detained_limit()` / `frame_tx_limit()` so that `execution_result()` can re-derive whether detention was the binding constraint.

What do you think about encapsulating this entirely in `ComputeGasTracker::check_limit()`? For example, `LimitCheck::ExceedsLimit` could carry an extra field (or use a new variant) indicating that detention was the cause. Then this block would just map the check result to the appropriate `MegaHaltReason` — no need to reach into the tracker's internals.

Benefits:
- `detained_limit()` and `frame_tx_limit()` don't need to be exposed
- The `detained_limit < frame_tx_limit` guard (which is tricky to reason about) moves into the tracker where the context is clearer
- `execution_result()` becomes a straightforward mapping from limit check → halt reason

We also have to consider how to pass `access_type` into the `LimitCheck::ExceedsLimit` so that it can be converted to `MegaHaltReason::VolatileDataAccessOutOfGas`.

Troublor (approved):

PR #185, opened 2026-03-11
feat: Rex4 keyless deploy sandbox shares parent salt and oracle env

RealiCZ:
### Summary

  Rex4 fixes the keyless deploy sandbox to share the parent transaction's SALT env and oracle env, ensuring correct dynamic gas pricing for storage operations in keyless deploy constructors.

  Prior to this change, the sandbox always used `EmptyExternalEnv`, causing all storage operations (`SSTORE`, `CREATE`, new accounts) to use a 1x gas multiplier regardless of actual bucket capacities, and oracle hints emitted in constructors to be silently dropped.

  ### Changes

  - Add `salt_env: Rc<ExtEnvs::SaltEnv>` field to MegaContext and SaltEnv for `Rc<T>` blanket impl, enabling the sandbox to share the parent's SALT snapshot while keeping its own `DynamicGasCost` cache isolated
  - Add `new_with_shared_ext_envs` constructor for sandbox context creation with shared external envs
  - Gate the new behavior behind `MegaSpecId::REX4`; pre-REX4 retains `EmptyExternalEnv` for backward compatibility
  - Extract `process_sandbox_transact_result` helper to deduplicate result processing across REX4/pre-REX4 branches
  - Fix potential underflow in new_with_context (- 1 → saturating_sub(1))

  ### Test plan

  - `test_rex4_keyless_deploy_uses_salt_env` — elevated bucket capacity increases sandbox gas
  - `test_rex3_keyless_deploy_ignores_salt_env` — bucket capacity has no effect on pre-REX4 sandbox gas
  - `test_rex4_keyless_deploy_constructor_uses_oracle_env` — constructor hints forwarded to parent
  - `test_rex3_keyless_deploy_constructor_ignores_oracle_env` — pre-REX4 sandbox drops hints
  - `test_shared_salt_env_keeps_dynamic_gas_cache_isolated` — parent bucket cache unaffected by sandbox

### DIFF
--- a/crates/mega-evm/src/external/salt.rs
+++ b/crates/mega-evm/src/external/salt.rs
@@ -2,13 +2,10 @@
 //! gas pricing. Storage slots and accounts are organized into buckets, and the gas cost scales
 //! with bucket capacity to incentivize efficient resource allocation.
 
-#[cfg(not(feature = "std"))]
-use alloc as std;
 use core::{
     convert::Infallible,
     fmt::{Debug, Display},
 };
-use std::rc::Rc;
 
 use alloy_primitives::{Address, U256};
 use auto_impl::auto_impl;
@@ -45,7 +42,7 @@ pub const MIN_BUCKET_SIZE: usize = 1 << MIN_BUCKET_SIZE_BITS;
 /// The trait provides default methods [`bucket_id_for_account`](SaltEnv::bucket_id_for_account)
 /// and [`bucket_id_for_slot`](SaltEnv::bucket_id_for_slot) that can be overridden by
 /// implementations to customize bucket assignment logic.
-#[auto_impl(&, Box, Arc)]
+#[auto_impl(&, Box, Rc, Arc)]
 pub trait SaltEnv: Debug + Unpin {
     /// Error type returned when bucket capacity cannot be retrieved.
     type Error: Display;
@@ -110,22 +107,5 @@ impl SaltEnv for EmptyExternalEnv {
 
     fn bucket_id_for_slot(_address: Address, _key: U256) -> BucketId {
         0 as BucketId
-    }
-}
-
-/// Delegates SALT bucket queries through a shared `Rc` handle without cloning the underlying env.
-impl<T: SaltEnv + ?Sized> SaltEnv for Rc<T> {
-    type Error = T::Error;
-
-    fn get_bucket_capacity(&self, bucket_id: BucketId) -> Result<u64, Self::Error> {
-        self.as_ref().get_bucket_capacity(bucket_id)
-    }
-
-    fn bucket_id_for_account(account: Address) -> BucketId {
-        T::bucket_id_for_account(account)
-    }
-
-    fn bucket_id_for_slot(address: Address, key: U256) -> BucketId {
-        T::bucket_id_for_slot(address, key)
     }
 }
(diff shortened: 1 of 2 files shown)


crates/mega-evm/src/evm/context.rs
```diff
@@ -53,8 +53,11 @@ pub struct MegaContext<DB: Database, ExtEnvs: ExternalEnvTypes> {
     /// Additional limits for the EVM.
     pub additional_limit: Rc<RefCell<AdditionalLimit>>,
 
+    /// Shared SALT environment handle.
+    pub(crate) salt_env: Rc<ExtEnvs::SaltEnv>,
+
     /// Calculator for dynamic gas costs during transaction execution.
-    pub dynamic_storage_gas_cost: Rc<RefCell<DynamicGasCost<ExtEnvs::SaltEnv>>>,
+    pub dynamic_storage_gas_cost: Rc<RefCell<DynamicGasCost<Rc<ExtEnvs::SaltEnv>>>>,
```
Troublor: Consider adding `Clone` as a supertrait bound on `SaltEnv` instead of wrapping it in `Rc` here.

Since `SaltEnv` is read-only (`get_bucket_capacity` takes `&self`), `Clone` and `Rc` sharing are semantically equivalent. But `Clone` keeps the types much simpler:
- `DynamicGasCost<ExtEnvs::SaltEnv>` instead of `DynamicGasCost<Rc<ExtEnvs::SaltEnv>>`
- No need for the `salt_env: Rc<...>` field on `MegaContext`
- No need for the `impl SaltEnv for Rc<T>` blanket impl

Production implementations likely already hold a shared DB handle internally, so `Clone` would be trivial.

This would also let the sandbox reuse the existing `with_external_envs()` builder — just clone the parent's salt/oracle envs into an `ExternalEnvs` and call `with_external_envs()`, rather than needing the new `new_with_shared_ext_envs()` constructor.

crates/mega-evm/src/external/salt.rs
```diff
@@ -109,3 +112,20 @@ impl SaltEnv for EmptyExternalEnv {
         0 as BucketId
     }
 }
+
+/// Delegates SALT bucket queries through a shared `Rc` handle without cloning the underlying env.
+impl<T: SaltEnv + ?Sized> SaltEnv for Rc<T> {
```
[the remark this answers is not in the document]
RealiCZ: Good catch. I went back to verify this against the `auto_impl` docs for our pinned version, and `Rc` is indeed supported.

 I also tried the change locally and it compiles fine with `#[auto_impl(&, Box, Rc, Arc)]`, so I will just use that and remove the manual `Rc` impl.

crates/mega-evm/src/evm/context.rs
RealiCZ: I thought about the `SaltEnv: Clone` direction, but I am not sure it is a net win here.

My main concern is that `Rc` makes the sharing semantics explicit in the sandbox path, while `Clone` pushes that into each `SaltEnv` impl.

  Also, even if we did that for `SaltEnv`, we **still need real shared ownership for `OracleEnv`** because sandbox-emitted hints need to flow back to the parent context. So I don't think this would actually let us get rid of the shared-env constructor entirely.

  My current preference is to keep the explicit shared-handle approach in this PR, and treat a possible `Clone`-based cleanup as a separate refactor if we want to revisit the external env ownership model more broadly.

PR #187, opened 2026-03-13
test: add Rex4 system contract deployment and idempotence tests

viktorcrypt:
## Summary

Adds deployment wiring tests for the two Rex4 system contracts:
`MegaAccessControl` (0x...0004) and `MegaLimitControl` (0x...0005).

## What is tested

- **Activation**: verifies both contracts are deployed at the correct
  addresses after `apply_pre_execution_changes()` on a Rex4 block
- **Idempotence**: verifies that calling `apply_pre_execution_changes()`
  twice does not panic or corrupt state
- **Spec boundary**: verifies that on Rex3, neither contract is deployed

## Why

The interception behavior of these contracts was already covered,
but the hardfork-gated deployment wiring in `block/executor.rs` had
no test coverage. These tests close that gap for the Rex4 unstable frontier.

### DIFF
--- a/crates/mega-evm/tests/rex4/deployment.rs
+++ b/crates/mega-evm/tests/rex4/deployment.rs
@@ -1,0 +1,192 @@
+//! Tests for Rex4 system contract deployment wiring.
+
+use alloy_evm::{block::BlockExecutor, Database, Evm, EvmEnv, EvmFactory};
+use alloy_hardforks::ForkCondition;
+use alloy_op_evm::block::receipt_builder::OpAlloyReceiptBuilder;
+use alloy_primitives::{Address, B256, Bytes};
+use mega_evm::{
+    test_utils::MemoryDatabase, BlockLimits, MegaBlockExecutionCtx, MegaBlockExecutor,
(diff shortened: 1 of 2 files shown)


crates/mega-evm/tests/rex4/deployment.rs
```diff
@@ -0,0 +1,192 @@
...
+    );
+    let receipt_builder = OpAlloyReceiptBuilder::default();
+    let mut executor = MegaBlockExecutor::new(evm, block_ctx, rex3_chain_spec(), receipt_builder);
+
+    executor.apply_pre_execution_changes().expect("pre-execution changes should succeed");
+
+    let db_ref = executor.evm_mut().db_mut();
+    assert_contract_not_deployed(db_ref, ACCESS_CONTROL_ADDRESS, "MegaAccessControl");
+    assert_contract_not_deployed(db_ref, LIMIT_CONTROL_ADDRESS, "MegaLimitControl");
+}
```
Troublor: Significant setup duplication across all three test functions (~20 lines of identical boilerplate each). The existing test suite typically extracts setup into helper functions. Consider a helper like `fn make_executor(spec: MegaSpecId, hardforks: MegaHardforkConfig) -> MegaBlockExecutor<...>` to reduce repetition.

PR #202, opened 2026-03-20
docs: add MegaEVM book docs and refactor spec docs structure

RealiCZ:
## Summary

- Add GitBook-based documentation structure under `book/`, including EVM, system contracts, and network upgrade pages.
- Add `specs/STYLE-GUIDE.md` to define normative/informative doc conventions and writing rules.
- Add per-spec informative companion docs under `specs/impl/` (Behavior Details + Implementation References).
- Keep this PR documentation/spec-text focused, without runtime code changes.

## Test plan

- Verified changed files are documentation/spec files only (`book/`, `specs/`, `.gitbook.yaml`).
- Manually reviewed cross-links and document structure consistency (`book/src/SUMMARY.md`, spec references).
- No runtime behavior validation needed for this docs/spec-only PR.


RealiCZ:
@claude Please post a final review summary confirming the current state of the PR.

book/src/evm/README.md
```diff
@@ -0,0 +1,186 @@
...
+### Dynamic SALT Multiplier
+
+Storage gas costs for SSTORE, account creation, and contract creation scale dynamically based on SALT bucket capacity.
+
+**Formula**: `multiplier = bucket_capacity / MIN_BUCKET_SIZE`
+
+- At `multiplier = 1` (minimum bucket size): **zero storage gas** — fresh storage is free.
+- At `multiplier > 1`: linear scaling makes crowded state regions more expensive.
+
+See [Dual Gas Model](dual-gas-model.md) for details.
```
Troublor: After reading the SALT multiplier section, a developer's immediate question is: "How do I estimate gas for my transactions?"

With dynamic SALT multipliers, storage gas, gas detention, and multidimensional resource limits all affecting the final cost, a developer cannot manually compute gas costs from the formulas alone — the multiplier depends on on-chain state they may not have access to.

Suggestion: Add a brief practical note here, e.g.:

> **Gas estimation**: Call `eth_estimateGas` on a MegaETH RPC endpoint to get precise gas estimates. The endpoint accounts for SALT multipliers, storage gas, and all resource dimensions. Do not attempt to compute gas costs manually from the formulas above — the dynamic multiplier depends on on-chain SALT bucket state.

specs/MiniRex.md
```diff
@@ -1,307 +1,214 @@
...
+- **Storage gas**: Additional costs for operations that impose persistent storage burden on nodes.
 
-**Increased Maximum Initcode Size:**
-536,576 bytes (512 KB + 24 KB)
+Storage gas MUST be charged for the following operations:
 
-- In comparison, Standard EVM (EIP-3860): 49,152 bytes (48 KB)
+| Operation | Storage Gas | Condition |
+| --- | --- | --- |
+| **SSTORE (0→non-0)** | 2,000,000 × multiplier | Only zero-to-non-zero transitions |
```
Troublor: "Only zero-to-non-zero transitions" is ambiguous for a reimplementer.

Verified against code (`instructions.rs`), the exact condition is:
```
original_value == 0 AND present_value == 0 AND new_value != 0
```

This is the "clean zero → nonzero" case only. Critically, if `original == 0` but `present != 0` (the slot was already written to non-zero earlier in the same tx), **no storage gas is charged** on re-write — even though the transition from the caller's perspective looks like a zero-to-non-zero write.

A reimplementer reading "zero-to-non-zero" could reasonably implement `present == 0 && new != 0` (missing the `original == 0` guard), which would overcharge in the `original != 0, present == 0, new != 0` case (slot cleared then re-set within the same tx).

Suggestion: State the exact triple condition:
> Only when `original_value == 0 AND present_value == 0 AND new_value != 0` (using EIP-2200 terminology: original is the value at transaction start, present is the current value in the journal).

RealiCZ:
@Troublor 

1. `.claude/REVIEW.md` shouldn't be committed to the repo — I've removed it.
2. `book/src/hardfork-spec.md` has 8 broken relative links — the file sits at `book/src/` root
   level but all links use `../`prefix  - Fixed

Everything else LGTM.

PR #209, opened 2026-03-27
feat: add REX4 storage gas stipend for value transfers

RealiCZ:
## Summary

  Rex4 introduces `STORAGE_CALL_STIPEND` (23,000 gas) for value-transferring `CALL`/`CALLCODE`, fixing LOG events in `receive()` / `fallback()` that broke
  under MegaETH's 10× storage gas multiplier.

  Prior to this change, the EVM's `CALL_STIPEND` (2,300 gas) was the only gas available to callees receiving ETH via `transfer()` or `send()`.
  A single `LOG1` costs 4,500 gas on MegaETH (750 compute + 3,750 storage), exceeding the stipend and causing any event-emitting `receive()` to revert.
  Affected contracts include WETH9, Gnosis Safe, and smart contract wallets.

  ## Changes

- Add `STORAGE_CALL_STIPEND = 23_000` constant in `constants::rex4`
- Add `StorageCallStipendTracker` in `limit/storage_call_stipend.rs`, following the same pattern as other trackers (`ComputeGasTracker`, `DataSizeTracker`, etc.)
- Detect qualifying value-transferring `CALL`/`CALLCODE` (`depth > 0`) in `AdditionalLimit::before_frame_init` and inflate `gas_limit`; `forward_gas_ext` and `execution.rs` are unaware of the stipend
- Two-phase frame init: `before_frame_init` detects and inflates `gas_limit` (before other trackers push), `after_frame_init` pushes stipend to burn stack and caps per-frame compute gas at the original `forwarded_gas + CALL_STIPEND` (after compute frame exists)
- `before_frame_return_result` burns unused stipend by capping returned gas at `gas_limit - stipend`, preventing gas leakage; burn applies identically on success and revert (no discardable usage)
- System contract and inspector interception paths push a zero-stipend empty frame for stack alignment without granting or burning any stipend
- Stipend amount derived from spec via `StorageCallStipendTracker::stipend_for_spec`, making it easy to adjust per spec in the future
- Add `SystemContractInterceptor` doc comment noting that intercepted calls do not receive `STORAGE_CALL_STIPEND`

  ## Test plan

  - `cargo fmt --all --check` - passed
  - `cargo clippy --workspace --lib --examples --tests --benches --all-features --locked` - passed
  - `cargo test -p mega-evm --tests` — 160 passed

### DIFF
--- a/crates/mega-evm/src/limit/limit.rs
+++ b/crates/mega-evm/src/limit/limit.rs
@@ -315,6 +315,10 @@ impl AdditionalLimit {
     /// afterwards.
     pub(crate) fn rescue_gas(&mut self, gas: &Gas) {
         let stipend = self.storage_call_stipend.current_frame_stipend();
+        debug_assert_eq!(
+            stipend, 0,
+            "rescue_gas called with active stipend {stipend} — inflated gas would leak"
+        );
         let effective_remaining = if stipend > 0 {
             let original_limit = gas.limit().saturating_sub(stipend);
             gas.remaining().min(original_limit)


docs/upgrades/rex4.md
```diff
@@ -120,6 +122,33 @@ interface IMegaLimitControl {
...
+- On return, unused storage gas stipend is **burned** — it is never returned to the caller.
+- `DELEGATECALL` and `STATICCALL` are unaffected (no value transfer, no stipend).
+- The compute gas cap ensures the callee cannot perform state-modifying operations (SSTORE, CALL with value, CREATE) with the extra gas, preserving the reentrancy protection properties of the original `CALL_STIPEND`.
+
+| Event type | Compute gas | Storage gas | Total gas | Fits in 25,300? |
+| ---------- | ----------- | ----------- | --------- | --------------- |
+| LOG1 (1 topic, 0 bytes) | 750 | 3,750 | 4,500 | ✅ |
+| LOG2 (SafeReceived) | 1,125 | 7,500 | 8,625 | ✅ |
+| LOG3 (Transfer) | 1,500 | 11,250 | 12,750 | ✅ |
+| LOG4 + 32 bytes data | ~2,075 | 17,560 | ~19,635 | ✅ |
```
Troublor: Nit: compute gas for LOG4 + 32 bytes should be 375 + 4×375 + 32×8 = **2,131**, not ~2,075. That makes the total ~19,691.

```suggestion
| LOG4 + 32 bytes data | ~2,131 | 17,560 | ~19,691 | ✅ |
```

crates/mega-evm/src/constants.rs
```diff
@@ -114,6 +114,20 @@ pub mod rex4 {
...
+    /// 4,500 gas (750 compute + 3,750 storage), exceeding the EVM's `CALL_STIPEND` of 2,300.
+    /// This breaks `transfer()` / `send()` to contracts that emit events in `receive()`.
+    ///
+    /// When a CALL or CALLCODE transfers value (value > 0), the callee receives this extra gas
+    /// on top of the standard `CALL_STIPEND`.
+    /// The callee's compute gas limit remains at the original level (`forwarded_gas +
+    /// CALL_STIPEND`), so the extra gas can only be consumed by storage gas operations (e.g.,
+    /// the 10x LOG topic/data costs).
+    /// On return, unused `STORAGE_GAS_STIPEND` is burned — it is never returned to the caller.
+    pub const STORAGE_GAS_STIPEND: u64 = 23_000;
```
Troublor: Consider renaming to `STORAGE_CALL_STIPEND` — it keeps the same `_CALL_STIPEND` suffix as revm's `CALL_STIPEND`, making the relationship between the two immediately obvious. They're roughly the same concept (extra gas granted to the callee on value transfer), just for different gas dimensions.

crates/mega-evm/src/evm/instructions.rs
```diff
@@ -609,11 +613,34 @@ pub mod forward_gas_ext {
...
+                            };
+
                         //  Return the gas to the parent call.
                         context.interpreter.gas.erase_cost(gas_to_return as u64);
 
-                        // Set the child call gas limit to the capped value.
-                        call_inputs.gas_limit = new_child_gas as u64;
+                        // Set the child call gas limit to the capped value plus any
+                        // storage gas stipend.
+                        call_inputs.gas_limit = (new_child_gas + storage_stipend) as u64;
```
Troublor: Consider moving the storage gas stipend logic out of `wrap_gas_cap!` and into `AdditionalLimit` instead.

**How**: Change `before_frame_init` to take `&mut FrameInit` (it's an internal API, not a trait). Then `before_frame_init` can detect value-transferring CALL/CALLCODE itself (via `call_inputs.scheme` + `call_inputs.transfers_value()`), inflate `gas_limit` directly, push to the stipend stack, and cap compute gas — all in one place.

**Why this is better**:
1. **Eliminates the "pending" two-step pattern** (`pending_storage_gas_stipend` / `pending_compute_gas_cap` fields, set here, consumed later in `before_frame_init`/`push_empty_frame`). This pattern is fragile — if the pending values are ever set but not consumed, or consumed without being set, you get silent bugs. Consolidation removes that class of bugs entirely.
2. **Single responsibility**: `wrap_gas_cap!` stays focused on the 98/100 forwarding rule. `AdditionalLimit` owns all stipend lifecycle (grant → track → burn), making the invariants easier to audit.
3. **Removes cross-module coupling**: Currently the instruction layer must know about `STORAGE_GAS_STIPEND` and `set_pending_storage_gas_stipend`. After consolidation, instructions don't need to know about the stipend at all.

A short comment can be left here in `wrap_gas_cap!` noting that the storage gas stipend is applied later in `AdditionalLimit::before_frame_init`, so readers know to look there.

PR #251, opened 2026-04-14
feat: add replay cache-file mode

RealiCZ:
## Summary

  This PR adds `--rpc.cache-file` support to mega-evme replay.
  It introduces a transport-level RPC capture/replay path that can persist full JSON-RPC request/response fixtures and replay them offline.
  It also threads captured external env data through replay fixtures so offline replay can reuse bucket-capacity inputs without extra CLI flags.
  Along the way, it refactors provider construction to separate standard RPC, capture mode, and replay mode, and updates mega-evme tests and fixtures to cover the new behavior.

  ## Test plan

  - `cargo test -p mega-evme provider -- --nocapture`
  - `cargo test -p mega-evme state -- --nocapture`
  - `cargo test -p mega-evme replay -- --nocapture`
  - `cargo fmt --all --check`
  - `cargo clippy --workspace --lib --examples --tests --benches --all-features --locked`


RealiCZ:
@claude review

RealiCZ:
- **Documentation updates deferred**:  Will be done in a follow-up commit to keep the code diff focused.

bin/mega-evme/src/common/provider.rs
```diff
@@ -56,36 +47,40 @@ pub struct BuildProviderOutput {
...
-    pub rpc_url: String,
+    pub rpc_url: Option<String>,
+
+    /// Single-file RPC cache and fixture.
+    /// Combined with --rpc: load if present, fetch missing entries, persist on clean exit.
+    /// Without --rpc: read-only replay; any RPC miss is a hard error; file is never written.
+    /// Cannot be used with --rpc.cache-dir, --rpc.clear-cache,
+    /// --rpc.no-cache-file, or --rpc.cache-size.
+    #[arg(
+        long = "rpc.cache-file",
```
Troublor: The name `--rpc.cache-file` is confusingly similar to `--rpc.cache-dir`. Both sound like they control "RPC caching" with slightly different storage layouts, but they serve fundamentally different purposes:

- `--rpc.cache-dir` is a **performance cache** (provider-level LRU for ~8 methods)
- `--rpc.cache-file` is a **test fixture capture/replay** mechanism (transport-level, captures every RPC call)

Consider renaming to something that makes the distinction explicit, e.g. `--rpc.fixture-file`, `--rpc.recording`, or `--rpc.capture-file`. The current name suggests "same thing, different granularity" when the intent and behavior are quite different.

bin/mega-evme/src/common/provider.rs
```diff
@@ -1,35 +1,24 @@
 //! RPC provider factory and on-disk cache store for mega-evme.
```
Troublor: This file is now 1000+ lines with three distinct concerns: CLI args & provider builders, cache persistence (store + envelope I/O), and transport layer (CachingTransport / ReplayTransport / TransportCache). Consider splitting into a module:

```
common/provider/
  mod.rs          — RpcArgs, build_* methods, BuildProviderOutput
  cache_store.rs  — RpcCacheStore, save_cache_atomic, envelope I/O
  transport.rs    — TransportCache, CachingTransport, ReplayTransport
```

This would keep each file under ~300 lines and make it easier to review transport-layer changes independently from CLI/persistence logic.

bin/mega-evme/src/common/provider.rs
```diff
@@ -56,36 +47,40 @@ pub struct BuildProviderOutput {
...
     #[arg(
         long = "rpc",
         visible_aliases = ["rpc-url"],
         alias = "fork.rpc",
-        env = "RPC_URL",
-        default_value = "http://localhost:8545"
     )]
-    pub rpc_url: String,
+    pub rpc_url: Option<String>,
+
```
Troublor: The capture vs. replay mode distinction is determined implicitly by whether `--rpc` is also present, but this isn't obvious from the CLI help. A user reading `--rpc.cache-file` alone can't tell it has two completely different behaviors:

- `--rpc.cache-file foo.json` → read-only offline replay
- `--rpc <URL> --rpc.cache-file foo.json` → live capture with persistence

Consider making the mode explicit, e.g. two separate flags like `--rpc.capture <file>` and `--rpc.replay <file>`, or a subcommand/mode selector. The current design requires reading the docs to understand the modal behavior.

bin/mega-evme/src/common/provider.rs
```diff
@@ -418,183 +483,401 @@ impl RpcCacheStore {
...
+}
+
+/// Snapshot of mega-evm external environment inputs not derivable from RPC.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct ExternalEnvSnapshot {
+    /// SALT bucket capacity pairs `(bucket_id, capacity)`.
+    #[serde(default)]
+    pub bucket_capacities: Vec<(u32, u64)>,
+}
+
```
Troublor: `load_envelope` and `save_envelope` are natural methods on `CacheFileEnvelope`:

```rust
impl CacheFileEnvelope {
    fn load(path: &Path) -> Result<Self> { ... }
    fn save(&self, path: &Path) -> Result<()> { ... }
    fn new(cache: &TransportCache, chain_id: u64, external_env: Option<&ExternalEnvSnapshot>) -> Self { ... }
}
```

This collocates the I/O with the type, cleans up the 4-parameter `save_envelope` signature, and makes the construction + serialization responsibilities clearer.

bin/mega-evme/src/common/provider.rs
```diff
@@ -418,183 +483,401 @@ impl RpcCacheStore {
...
-        self.inner.as_ref().map(|inner| &inner.cache)
+        match &self.inner {
+            Some(RpcCacheStoreInner::Raw { cache, .. }) => Some(cache),
+            _ => None,
+        }
     }
 
     /// Returns the resolved cache file path, or `None` for a no-op store.
     #[cfg(any(test, feature = "test-utils"))]
     pub fn cache_path(&self) -> Option<&Path> {
```
Troublor: The variant names `Raw` and `Envelope` describe the serialization format, not the functionality. A reader seeing `Raw` vs `Envelope` can't tell what each does without reading the doc comments. Consider names that reflect the workflow:

```rust
enum RpcCacheStoreInner {
    /// Provider-level performance cache (`--rpc.cache-dir`).
    ProviderCache { cache: SharedCache, path: PathBuf },
    /// Transport-level fixture capture (`--rpc.cache-file`).
    FixtureCapture { cache: TransportCache, path: PathBuf, chain_id: u64 },
}
```

bin/mega-evme/src/replay/cmd.rs
```diff
@@ -75,217 +89,236 @@ pub(super) struct ReplayOutcome {
     pub receipt: OpTxReceipt,
```
Troublor, line 89: The refactoring removed most `debug!`/`trace!` log lines, leaving only the final execution result (`info!`/`warn!` at lines 337-340). The replay pipeline is now opaque between "provider built" and "execution succeeded/reverted/halted" — no visibility into provider mode selection, transaction fetch, block determination, state fork, chain/spec config, preceding tx progress, or override application.

Consider keeping at least key `info!` checkpoints so users with `RUST_LOG=info` can follow the pipeline:
- Which provider mode was selected (replay/capture/online)
- Transaction found at block N
- Forking state from parent block
- Executing N preceding transactions
- Applying transaction overrides (when present)

Troublor, line 89: To clarify the previous comment: the removed `debug!` and `trace!` lines should also be kept, not just `info!`. They were there for a reason — `debug!` for block number determination, transaction lookups, and chain config; `trace!` for per-transaction outcomes and receipt details. Stripping all verbose logging as part of a refactor makes future debugging harder. The refactor should preserve the existing observability at all levels.

bin/mega-evme/src/common/provider.rs
```diff
@@ -418,183 +483,401 @@ impl RpcCacheStore {
...
+            if let Some(fut) = self.cache.try_serve(&key, r.id().clone()) {
+                return fut;
+            }
+
+            // Cache miss: forward to inner, cache successful responses.
+            let cache = self.cache.clone();
+            let fut = self.inner.call(req);
+            return Box::pin(async move {
+                let response = fut.await?;
+                if let ResponsePacket::Single(ref resp) = response {
```
Troublor: The comment on line 801 says "cache successful responses" but the code caches all `ResponsePacket::Single` responses, including JSON-RPC error responses (e.g., `-32000`, `-32603`). If the endpoint returns a transient error as a JSON-RPC error body (rather than an HTTP error), it gets baked into the cache and replayed forever.

Low risk for a developer-driven fixture workflow, but the code should match its comment:

```suggestion
                if let ResponsePacket::Single(ref resp) = response {
                    if resp.is_error() {
                        tracing::warn!(method = %r.method(), "Skipping cache for JSON-RPC error response");
                    } else if let Ok(serialized) = serde_json::to_string(resp) {
                        cache.put(key, serialized);
                    }
```

bin/mega-evme/src/common/provider.rs
```diff
@@ -274,15 +223,128 @@ impl RpcArgs {
...
+        })
+    }
+
+    /// Build the provider in replay mode (`--rpc.cache-file` without `--rpc`).
+    ///
+    /// Loads the envelope's transport-level cache and builds the provider over
+    /// [`ReplayTransport`], which serves cached responses directly and returns
+    /// a descriptive error on cache miss. No `CacheLayer` is used — all caching
+    /// is at the transport level so every RPC method is covered.
+    pub async fn build_replay_provider(&self) -> Result<BuildProviderOutput> {
```
Troublor: The new provider builders (`build_replay_provider`, `build_capture_provider`) only log a single `info!` at the end. The intermediate steps lack observability — consider adding `debug!` logs for:
- Envelope loaded: number of cache entries, whether `external_env` is present
- Capture mode: existing envelope found/not found, chain ID validation passed, number of entries merged

This matches the level of detail that `build_provider` has for its cache path (e.g., logging cache dir creation, clear-cache, load).

vincent-k2026 (approved):
LGTM. The provider split and capture/replay workflow are solid. Please make sure docs are updated to reflect the new flags and replay-mode constraints as noted in comments.

vincent-k2026:
Looks great overall. One final note before merge: please sync the docs with the new CLI semantics (`--rpc.capture-file` / `--rpc.replay-file`) and add a short migration note from the old cache-file workflow. In particular, it would help to document offline replay behavior and the `--bucket-capacity` constraint in replay mode. Once that is in, this change is in great shape.

Cargo.toml
```diff
@@ -135,6 +135,7 @@ alloy-consensus = { version = "1.0.23", default-features = false }
 alloy-eips = { version = "1.0.23", default-features = false }
 alloy-evm = { version = "0.15.0", default-features = false }
 alloy-hardforks = { version = "0.2.7", default-features = false }
+alloy-json-rpc = { version = "1.0.41", default-features = false }
```
flyq, line 138:
```suggestion
alloy-json-rpc = { version = "1.0.23", default-features = false }
```
建议 alloy 相关依赖的版本统一，具体参考 mega-reth

Cargo.toml
```diff
@@ -146,6 +147,7 @@ alloy-rpc-types-eth = { version = "1.0.24", default-features = false, features =
 alloy-rpc-types-trace = { version = "1.0.22", default-features = false }
 alloy-serde = { version = "1.0.23", default-features = false }
 alloy-sol-types = { version = "1.0.23", default-features = false }
+alloy-transport = { version = "1.0.41", default-features = false }
 alloy-transport-http = { version = "1.0.41", default-features = false }
```
flyq, line 151: 不过这里之前的 alloy-transport-http 和  alloy-transport 版本已经不同了，我也不确定 @Troublor

bin/mega-evme/src/common/provider/transport.rs
```diff
@@ -0,0 +1,295 @@
...
+    }
+}
+
+/// Compute a deterministic cache key from an RPC method and params.
+/// The request ID is deliberately excluded so the same logical request
+/// produces the same key across runs.
+fn transport_cache_key(method: &str, params: Option<&serde_json::value::RawValue>) -> B256 {
+    let params_str = params.map_or("null", |p| p.get());
+    keccak256(format!("{method}\x00{params_str}"))
+}
```
flyq, line 129: - **`transport_cache_key` params canonicalization (transport.rs:126–129)**: The cache key is `keccak256(method + "\x00" + params_raw_json)`. `params.get()` returns the raw unparsed JSON from `serde_json::value::RawValue`. If the alloy client serializes "equivalent" params with slightly different byte form across runs (whitespace, key ordering, numeric formatting like `1` vs `1.0`), the hash changes and replay misses. In practice alloy uses a deterministic `serde_json::to_string` so this should be stable within a single alloy version — but it's implicitly coupled to alloy's serialization output. A cross-alloy-version fixture would miss. **Suggestion**: consider either (a) documenting this alloy-version coupling prominently in `transport.rs`, or (b) normalizing via `serde_json::from_str::<serde_json::Value>(raw).unwrap()` and re-serializing through a canonical form before hashing. Option (b) costs a parse per request.

Cargo.toml
Troublor: mega-evm和mega-reth没必要一定使用同一个alloy版本的，没关系。

Troublor (approved):

PR #260, opened 2026-04-20
feat: add SequencerRegistry and dynamic system address

RealiCZ:
## Summary

  Introduce `SequencerRegistry` system contract at `0x634200...0006` (REX5) with two independent roles: **system address**  and **sequencer** .

  - Each role has its own `current*()`, `*At()`, `schedule*Change()`, rotation history, and independent pre-block apply.
  - Initial storage (system address, sequencer, admin, `initialFromBlock`) is seeded at deploy time from `SequencerRegistryConfig` on the hardfork config. No constructor execution.
  - A single `applyPendingChanges()` pre-block system call applies both roles atomically, gated by a Rust pre-check that reads 4 committed storage slots.
  - `resolve_system_address()` reads `_currentSystemAddress` (slot 0) after commit. Rotating the sequencer does NOT affect Oracle authority, and vice versa.
  - Oracle upgraded in-place to v2.0.0 — `onlySystemAddress` now reads `SequencerRegistry.currentSystemAddress()` instead of constructor `immutable`.

  **Note:** `mega-evme`'s `deploy_system_contracts()` is temporarily aligned in this PR (REX5 contracts added). A
  follow-up PR will refactor this to unify the system contract deployment path.

  ## Test plan

  - `cargo fmt --all --check` passed
  - `cargo clippy --workspace --lib --examples --tests --benches --all-features --locked` passed
  - `cargo test -p mega-evm` — 667 passed
  - `cargo bench -p mega-evm --bench block_bench -- rex5_pre_block` -- bootstrap ~17µs, no_rotation ~7µs
  - `forge test` — 67 passed (45 SequencerRegistry + 22 Oracle)
  - `npx prettier --check 'docs/**/*.md'` passed


flyq (commented):
考虑引入 `SystemContractCall` 错误变体，或者将 `BlockHashContractCall` 重命名得更通用一些——目前多个系统合约都在使用误导性的名称来重用它。

RealiCZ:
`BlockHashContractCall`来源于上游仓库，目前不太好做扩展。
https://github.com/alloy-rs/evm/blob/v0.15.0/crates/evm/src/block/error.rs
对等的还有一个分类是 `Internal`，但我觉得也不太好放在 `Internal` 里

@flyq

flyq (approved):

crates/mega-evm/src/system/sequencer_registry.rs
```diff
@@ -0,0 +1,545 @@
...
+pub use mega_system_contracts::sequencer_registry::ISequencerRegistry;
+
+/// The storage slot index for `_currentSequencer` (slot 0 in the Solidity layout).
+const SLOT_CURRENT_SEQUENCER: U256 = U256::ZERO;
+
+/// The storage slot index for `_pendingSequencer` (slot 2 in the Solidity layout).
+const SLOT_PENDING_SEQUENCER: U256 = U256::from_limbs([2, 0, 0, 0]);
+
+/// The storage slot index for `_activationBlock` (slot 3 in the Solidity layout).
+const SLOT_ACTIVATION_BLOCK: U256 = U256::from_limbs([3, 0, 0, 0]);
```
Troublor: ### Storage layout: make it explicit on the Solidity side, not hardcoded here

These `SLOT_*` constants encode Solidity's *implicit* field-declaration order (slot 0 = `_currentSequencer`, slot 1 = `_admin`, slot 2 = `_pendingSequencer`, etc.). Any field reorder in `SequencerRegistry.sol` silently corrupts these Rust reads — there's no compile-time coupling, and the only signal is a failing integration test if one happens to exercise the affected field. Once shipped in a spec, the layout is consensus-critical, which makes this fragility a real risk.

Suggestion: make the layout an explicit part of the contract via **ERC-7201 namespaced storage**. The contract declares a named struct at a deterministic base slot derived from a documented namespace string; Rust re-derives the same base and references per-field offsets against it. The struct declaration then becomes the single, auditable source of truth for the layout.

```solidity
/// @custom:storage-location erc7201:megaeth.storage.SequencerRegistry
struct SequencerRegistryStorage {
    address currentSequencer;              // offset 0
    address admin;                         // offset 1
    address pendingSequencer;              // offset 2
    uint256 activationBlock;               // offset 3
    RotationRecord[] rotations;            // offset 4
    address[] systemTxSenders;             // offset 5  (whitelist refactor)
    mapping(address => uint256) senderIndex; // offset 6
}

/// keccak256(abi.encode(uint256(keccak256("megaeth.storage.SequencerRegistry")) - 1))
/// & ~bytes32(uint256(0xff))
/// Precomputed; consensus-critical; documented in the spec.
bytes32 private constant SEQUENCER_REGISTRY_STORAGE_LOCATION = 0x...;

function _getStorage() private pure returns (SequencerRegistryStorage storage $) {
    assembly { $.slot := SEQUENCER_REGISTRY_STORAGE_LOCATION }
}
```

All reads/writes in Solidity go through `_getStorage()`. Field addition (e.g., the whitelist fields) extends the struct without touching existing offsets.

Rust side:

```rust
/// Matches SEQUENCER_REGISTRY_STORAGE_LOCATION in SequencerRegistry.sol.
/// Derived from keccak256("megaeth.storage.SequencerRegistry") - 1, low byte cleared.
const SEQUENCER_REGISTRY_STORAGE_LOCATION: U256 = ...; // same precomputed value

const SLOT_CURRENT_SEQUENCER: U256 = SEQUENCER_REGISTRY_STORAGE_LOCATION;
const SLOT_ADMIN: U256 = add_u256(SEQUENCER_REGISTRY_STORAGE_LOCATION, 1);
const SLOT_PENDING_SEQUENCER: U256 = add_u256(SEQUENCER_REGISTRY_STORAGE_LOCATION, 2);
// ...
```

Add a Rust/Solidity parity test that asserts `SEQUENCER_REGISTRY_STORAGE_LOCATION` matches across both sides (e.g., deploy the bytecode, call a view function that returns the base-slot constant, compare), and that reading back a freshly-seeded field through the computed slot yields the same value as calling the Solidity getter.

Payoff:

- Field order changes require touching a struct in one file; Rust can't drift silently.
- Spec can reference the namespace string as the definitional source for the layout.
- Future versioned layouts are cleanly expressed via `megaeth.storage.SequencerRegistry.v2` without disturbing v1 storage.

This applies naturally to the whitelist refactor (new fields extend the struct at offsets 5/6) and should extend to any other system contract whose storage is read directly by Rust.

crates/system-contracts/contracts/SequencerRegistry.sol
```diff
@@ -0,0 +1,161 @@
...
+///      Deployed by mega-evm without constructor execution — initial state comes from constants.
+contract SequencerRegistry is ISemver, ISequencerRegistry {
+    // =========================================================================
+    // Protocol constants — must match the Rust-side REX5 constants.
+    // These are compiled into bytecode and verified at build time.
+    // =========================================================================
+
+    /// @notice The initial sequencer address, used when no rotation has occurred.
+    /// @dev Compile-time constant — must match the Rust-side REX5 protocol constant.
+    address public constant INITIAL_SEQUENCER = address(0xA887dCB9D5f39Ef79272801d05Abdf707CFBbD1d);
```
Troublor: ### Architectural suggestion: split into two single-address roles, each with pending/apply

`INITIAL_SEQUENCER = 0xA887…BD1d` in this PR only works because no rotation has occurred. Today `0xA887…BD1d` is the Oracle-update sender (used as `MEGA_SYSTEM_ADDRESS`), while the sequencer signs mini-blocks with a *different* key. The first real rotation of the signing key would silently revoke Oracle authority from the production Oracle sender — chain break on first rotation.

#### Proposal

Track **two independent single-address roles** in the registry, each with its own pending + pre-block apply mechanism (same shape as the existing sequencer rotation):

- **`systemAddress`** — the authorized system-tx sender and Oracle authority. Initial = `0xA887…BD1d` (= today's `MEGA_SYSTEM_ADDRESS`, preserving production behavior).
- **`sequencer`** — the mini-block signing-key address. Initial = actual production signing-key address (to supply).

Neither role's rotation affects the other. One shared admin for now (can split later if needed).

#### Solidity shape (`SequencerRegistry.sol`)

Storage layout (consensus-critical — see the ERC-7201 comment below for how to express this explicitly instead of by implicit field order):

| Slot offset | Name | Type |
| ----------- | ---- | ---- |
| 0 | `currentSystemAddress` | `address` |
| 1 | `currentSequencer` | `address` |
| 2 | `admin` | `address` |
| 3 | `pendingSystemAddress` | `address` |
| 4 | `systemAddressActivationBlock` | `uint256` |
| 5 | `pendingSequencer` | `address` |
| 6 | `sequencerActivationBlock` | `uint256` |
| 7 | `systemAddressRotations` | `RotationRecord[]` |
| 8 | `sequencerRotations` | `RotationRecord[]` |

```solidity
/// Preserves today's Oracle sender — used when no rotation has occurred.
/// TODO(future PR): move this (and INITIAL_SEQUENCER / INITIAL_ADMIN) to a chain-params
/// system so a single bytecode artifact supports mainnet / testnet / devnet.
address public constant INITIAL_SYSTEM_ADDRESS = 0xA887dCB9D5f39Ef79272801d05Abdf707CFBbD1d;

// -- System address (Oracle-sender / system-tx sender) --
function currentSystemAddress() external view returns (address);
function systemAddressAt(uint256 blockNumber) external view returns (address);
function scheduleNextSystemAddressChange(address newSystemAddress, uint256 activationBlock) external; // onlyAdmin
function applyPendingSystemAddressChange() external; // permissionless

// -- Sequencer (mini-block signing) --  (already in the PR; unchanged semantics, rename only if needed)
function currentSequencer() external view returns (address);
function sequencerAt(uint256 blockNumber) external view returns (address);
function scheduleNextSequencerChange(address newSequencer, uint256 activationBlock) external; // onlyAdmin
function applyPendingSequencerChange() external; // permissionless

// -- Admin --
function admin() external view returns (address);
function transferAdmin(address newAdmin) external;

event SystemAddressChangeScheduled(address indexed oldSystemAddress, address indexed newSystemAddress, uint256 activationBlock);
event SequencerChangeScheduled(address indexed oldSequencer, address indexed newSequencer, uint256 activationBlock);
event AdminTransferred(address indexed oldAdmin, address indexed newAdmin);
error FutureBlock();
error NotAdmin();
error ZeroAddress();
error InvalidActivationBlock();
```

Same bootstrap-fallback pattern as the PR uses for `_currentSequencer`: `currentSystemAddress()` returns `_currentSystemAddress` if non-zero, otherwise `INITIAL_SYSTEM_ADDRESS`.

#### Oracle v2.0.0

```solidity
modifier onlySystemAddress() {
    _;
    if (msg.sender != SEQUENCER_REGISTRY.currentSystemAddress()) revert NotSystemAddress();
}
```

On bootstrap block this returns `0xA887…BD1d` → Oracle authority unchanged vs. production.

#### Rust side

- `system/sequencer_registry.rs`: add `REX5_INITIAL_SYSTEM_ADDRESS = MEGA_SYSTEM_ADDRESS`; add slot constants, `is_system_address_rotation_due`, `transact_apply_pending_system_address_change_if_due`, and `resolve_system_address` helpers (mirroring the existing sequencer-rotation quartet). **Also: update `INITIAL_SEQUENCER` from `0xA887…BD1d` to the actual production signing-key address.**
- `evm/context.rs`: two fields — `system_address: Address` (existing, now clearly the Oracle-sender role) and `current_sequencer: Address` (new). Both default to `MEGA_SYSTEM_ADDRESS` pre-Rex5. Two getters, two `pub(crate)` setters.
- `block/executor.rs`: `pre_execution_changes` runs two apply-if-due calls (system address, sequencer). After `commit_system_call_outcomes`, resolve both and set both context fields.
- `evm/host.rs`, `evm/execution.rs`, `system/tx.rs`: **unchanged from the current PR** — they correctly use `ctx.system_address` for the Oracle-sender role, and the field now genuinely names a single concept (no longer conflated with the sequencer).
- Extend `test_rust_constants_match_bytecode_constants` to call both `currentSystemAddress()` and `currentSequencer()` on a fresh deploy and assert both against their Rust constants.

#### Tests to add

- Bootstrap block resolves both roles to their constants.
- Rotating system address doesn't change `currentSequencer()`; rotating sequencer doesn't change `currentSystemAddress()`.
- Oracle authority after sequencer-only rotation: unchanged (old system address still works).
- Oracle authority after system-address-only rotation: new address gains authority, old loses.
- Block-stability of each role via existing pending/apply semantics.

#### Future PR (out of scope here)

Moving `INITIAL_SEQUENCER` / `INITIAL_ADMIN` / `INITIAL_SYSTEM_ADDRESS` off hardcoded constants onto a Rex5-chain-params system (Rust supplies per-network values; `SequencerRegistry` storage seeded at deploy) will let mainnet / testnet / devnet share one bytecode artifact.

#### mega-reth2 follow-ups (separate PR, not blocking)

`txpool/validate.rs:70-76` and `payload_executor.rs:526` swap `== MEGA_SYSTEM_ADDRESS` for `== currentSystemAddress()` (cached at block start). Until that lands, production is unaffected because the initial value is exactly `MEGA_SYSTEM_ADDRESS`.

#### Open questions

1. **Actual sequencer signing-key address** for `INITIAL_SEQUENCER` (the address `get_secret_key(&sequencer_secret())` derives to in production)?
2. Initial admin — keep `0xA887…BD1d` or a dedicated operator key? Admin and Oracle-sender don't have to be the same.
3. Naming: `currentSystemAddress` vs. alternatives like `systemTxSender`? First matches existing `SYSTEM_ADDRESS` docs terminology.

crates/mega-evm/src/system/sequencer_registry.rs
RealiCZ: In the current version, the slot constants are centralized in mega-system-contracts, and I added a parity test against the Solidity storage layout, so field reordering will fail tests instead of drifting silently.

crates/system-contracts/contracts/SequencerRegistry.sol
```diff
@@ -0,0 +1,221 @@
...
+    uint256 private _systemAddressActivationBlock;
+
+    /// @notice The next sequencer waiting to be applied.
+    address private _pendingSequencer;
+
+    /// @notice The block at which the pending sequencer becomes active.
+    uint256 private _sequencerActivationBlock;
+
+    /// @dev Packed: uint96 fromBlock + address addr fit in one 32-byte slot.
+    struct RotationRecord {
```
Troublor: ### Naming: drop "rotation" in favor of "change" for consistency with the external API

The external ABI already uses "change" consistently: `scheduleNextSystemAddressChange`, `scheduleNextSequencerChange`, `applyPendingChanges`, `SystemAddressChangeScheduled`, `SequencerChangeScheduled`. Internal storage, identifiers, and docs still lean on "rotation" — which made sense when a single sequencer role was rotating, but with two independent role changes, "rotation" now suggests cyclical/periodic semantics that don't exist here. Keeping both terms in play costs readers mental energy every time they cross the API/internal boundary.

Suggested rename sweep:

**Solidity** (`SequencerRegistry.sol`, `ISequencerRegistry.sol`):

- `struct RotationRecord` → `struct ChangeRecord` (alternatives: `RoleChange`, `HistoryEntry`).
- `_systemAddressRotations` → `_systemAddressHistory`.
- `_sequencerRotations` → `_sequencerHistory`.
- `ActivationBlockTooLarge` NatSpec: `(packed RotationRecord)` → `(packed ChangeRecord)`.
- Doc strings throughout: "rotation history" → "change history"; "rotation lifecycle" → "change lifecycle"; "first sequencer rotation" → "first sequencer change"; "ordered by activation block" stays.

**Rust** (`crates/mega-evm/src/system/sequencer_registry.rs`):

- `fn is_rotation_due(...)` → `fn is_role_change_due(...)` (internal helper).
- Module and function doc strings: same "rotation → change" sweep.

**Docs** (`docs/spec/system-contracts/sequencer-registry.md`, `docs/spec/upgrades/rex5.md`):

- Heading "Rotation Scheduling" → "Change Scheduling".
- Heading "Pre-Block Role Rotation" → "Pre-Block Role Change".
- Table header `RotationRecord[]` → `ChangeRecord[]`.
- Prose: "rotation lifecycle" / "role rotation" / "rotation history" → "change" equivalents.

Purely mechanical rename — no behavioral changes. Payoff is single terminology throughout the stack so readers (and future specs) don't need to translate.

crates/mega-evm/src/block/hardfork.rs
```diff
@@ -62,6 +62,11 @@ pub trait MegaHardforks: OpHardforks {
     /// [`ForkCondition::Never`].
     fn mega_fork_activation(&self, fork: MegaHardfork) -> ForkCondition;
 
+    /// Returns the bootstrap configuration for `SequencerRegistry`.
+    fn sequencer_registry_config(&self) -> SequencerRegistryConfig {
+        SequencerRegistryConfig::default()
+    }
+
```
Troublor: ### Trait scope: don't leak spec-specific config into `MegaHardforks`; co-locate params with the fork entry

Adding `sequencer_registry_config()` here means the trait now carries a method for one specific spec's one specific contract. Every future spec that needs deploy-time parameters would either add another named method (trait grows monotonically with each spec) or overload an existing one (confusing). It also leaves the params disjoint from the fork entry they activate alongside, losing the natural "a fork activates with these parameters" invariant.

Suggested reshape: embed params directly into each `ForkEntry`, type-check per fork via a trait that pins each params type to a single hardfork variant.

```rust
use core::any::{Any, TypeId};

/// Marks a type as params for a specific hardfork. Implementors pin themselves to one variant.
pub trait HardforkParams: Any + core::fmt::Debug + Send + Sync {
    const FORK: MegaHardfork;
}

/// Parameters for Rex5 system-contract deployment.
#[derive(Debug, Clone, Default, PartialEq, Eq)]
pub struct Rex5Params {
    pub initial_system_address: Address,
    pub initial_sequencer: Address,
    pub initial_admin: Address,
}
impl HardforkParams for Rex5Params {
    const FORK: MegaHardfork = MegaHardfork::Rex5;
}

/// One fork entry: identity, activation condition, and (optionally) its params.
pub struct ForkEntry {
    pub fork: Box<dyn Hardfork>,
    pub condition: ForkCondition,
    pub params: Option<Box<dyn Any + Send + Sync>>,
}

pub struct MegaHardforkConfig {
    hardforks: Vec<ForkEntry>,
}

impl MegaHardforkConfig {
    /// Register a hardfork (no params).
    pub fn with(mut self, fork: impl Hardfork + 'static, cond: ForkCondition) -> Self { ... }

    /// Attach params to the fork identified by `P::FORK`. Compile-time-safe.
    pub fn with_params<P: HardforkParams>(mut self, params: P) -> Self { ... }
}

pub trait MegaHardforks: OpHardforks {
    fn mega_fork_activation(&self, fork: MegaHardfork) -> ForkCondition;

    /// Erased params accessor — implementors override this.
    fn fork_params_any(&self, fork: MegaHardfork) -> Option<&(dyn Any + Send + Sync)>;

    /// Typed params accessor. `P::FORK` identifies the fork.
    fn fork_params<P: HardforkParams>(&self) -> Option<&P> {
        self.fork_params_any(P::FORK)?.downcast_ref::<P>()
    }
}
```

Call sites:

```rust
// Registration — fork + condition + params in the same builder chain:
let cfg = MegaHardforkConfig::default()
    .with(MegaHardfork::Rex5, ForkCondition::Timestamp(1_700_000_000))
    .with_params(Rex5Params { initial_system_address, initial_sequencer, initial_admin });

// Read — one typed call:
let rex5 = hardforks.fork_params::<Rex5Params>().expect("Rex5 params configured");
let addr = rex5.initial_system_address;
```

Payoffs:

- **Co-located.** Fork, `ForkCondition`, and params live together in `ForkEntry`. Reading a single entry answers "which fork, when, with what parameters".
- **Per-fork type safety.** `fork_params::<Rex5Params>()` returns `Option<&Rex5Params>`. Asking for the wrong type for Rex5 is impossible — the type itself pins the fork via `const FORK`. Missing registration is a `None`, wrong type is a compile error.
- **Trait scope bounded.** `MegaHardforks` stops growing with each spec; implementors only provide the erased accessor, the typed one is a default impl. No Ext-trait split since nothing in-tree uses `dyn MegaHardforks`.
- **Naming.** `HardforkParams` / `Rex5Params` / `fork_params` — no collision with `MegaHardforkConfig`.
- **Adding a future spec.** New `FooParams` struct + `impl HardforkParams` + builder call. No edits elsewhere.

Cost: `Box<dyn Any>` + `downcast_ref` at the accessor. Runtime-cheap, unsafe-free.

Notes:

- Most forks carry no params — their `ForkEntry::params` is simply `None`.
- If `dyn MegaHardforks` ever becomes necessary, split the generic `fork_params` into an Ext trait. Not worth paying that complexity up front.

crates/mega-evm/src/system/sequencer_registry.rs
```diff
@@ -0,0 +1,777 @@
...
+impl Default for SequencerRegistryConfig {
+    fn default() -> Self {
+        Self {
+            // Preserve the legacy single-address behavior unless the chain spec overrides it.
+            initial_system_address: MEGA_SYSTEM_ADDRESS,
+            initial_sequencer: MEGA_SYSTEM_ADDRESS,
+            initial_admin: MEGA_SYSTEM_ADDRESS,
+        }
+    }
+}
```
Troublor: ### Remove `Default` for `SequencerRegistryConfig`

This `Default` impl silently substitutes `MEGA_SYSTEM_ADDRESS` for all three roles — which is the exact "sequencer == admin == Oracle-sender" collapse this PR is meant to fix. Any chain that forgets to pass a `SequencerRegistryConfig` activates Rex5 with all three roles pointing at the same address, no error, wrong state committed to the bootstrap block.

Why it exists today:

- `MegaHardforks::sequencer_registry_config()` has a default trait impl that needs a value to return.
- `MegaHardforkConfig::default()` transitively needs it.
- Tests use it for convenience.

Under the `ForkEntry { params: Option<...> }` reshape in the other thread on `hardfork.rs`, none of those callers need a `Default` anymore:

- A fork entry with no params → `fork_params::<Rex5Params>()` returns `None` → callers must handle the absence deliberately (panic with a clear message, or fail closed inside the EVM).
- Tests explicitly construct `Rex5Params { ... }` with test-specific addresses.
- Production chains always supply real values; there's no silent-wrong-behavior path.

Concrete changes:

- Drop `impl Default for SequencerRegistryConfig` (renamed `Rex5Params` under the reshape).
- If default-construction is still useful for test ergonomics, expose it as a named non-trait fn like `Rex5Params::for_tests()` so the intent is explicit at call sites.

The three addresses are *required data* for a Rex5 chain, not optional values with a meaningful default. Anything that looks like a default here hides misconfiguration.

crates/mega-evm/src/system/sequencer_registry.rs
```diff
@@ -0,0 +1,777 @@
...
+    if acc.account_info().is_none() {
+        return Ok(false);
+    }
+
+    if is_rotation_due(db, PENDING_SYSTEM_ADDRESS, SYSTEM_ADDRESS_ACTIVATION_BLOCK, block_number)? {
+        return Ok(true);
+    }
+
+    is_rotation_due(db, PENDING_SEQUENCER, SEQUENCER_ACTIVATION_BLOCK, block_number)
+}
```
Troublor, line 215: ### Reads escape the stateless witness on the "nothing due" path

This function does up to 5 storage-layer reads that never flow into an outcome:

- `db.load_cache_account(SEQUENCER_REGISTRY_ADDRESS)` — account info.
- `read_registry_storage(db, PENDING_SYSTEM_ADDRESS)`.
- `read_registry_storage(db, SYSTEM_ADDRESS_ACTIVATION_BLOCK)` (when pending nonzero).
- `read_registry_storage(db, PENDING_SEQUENCER)`.
- `read_registry_storage(db, SEQUENCER_ACTIVATION_BLOCK)` (when pending nonzero).

Since the function returns `bool`, the reads never reach `system_caller.on_state` and don't enter the stateless witness. On the **"due"** branch this is masked because the subsequent `transact_apply_pending_changes` EVM call re-reads the same slots and captures them in its `ResultAndState`. On the **"not due"** branch — the common case, every block without a pending rotation — the reads are silent.

A stateless verifier needs to see `_pendingSystemAddress == 0` (and/or `_pendingSequencer == 0`) in its witness to agree that skipping `applyPendingChanges()` was correct. Without those reads in an outcome, the witness is incomplete.

Fix shape: return an `EvmState` alongside the decision, carrying a read-only account entry with a no-change slot record for every slot read:

```rust
pub(crate) fn is_apply_pending_changes_due<DB: Database>(
    db: &mut State<DB>,
    block_number: u64,
) -> Result<(bool, Option<EvmState>), BlockExecutionError> {
    let acc = db.load_cache_account(SEQUENCER_REGISTRY_ADDRESS)
        .map_err(BlockExecutionError::other)?;
    let Some(info) = acc.account_info() else {
        return Ok((false, None)); // registry not deployed; no reads to record
    };

    let mut account = Account { info, ..Default::default() };

    let pending_sys = read_registry_storage(db, PENDING_SYSTEM_ADDRESS)?;
    account.storage.insert(
        PENDING_SYSTEM_ADDRESS,
        EvmStorageSlot::new_changed(pending_sys, pending_sys, 0),
    );
    let mut due = false;
    if !pending_sys.is_zero() {
        let activation = read_registry_storage(db, SYSTEM_ADDRESS_ACTIVATION_BLOCK)?;
        account.storage.insert(
            SYSTEM_ADDRESS_ACTIVATION_BLOCK,
            EvmStorageSlot::new_changed(activation, activation, 0),
        );
        due |= block_number >= activation.saturating_to::<u64>();
    }

    let pending_seq = read_registry_storage(db, PENDING_SEQUENCER)?;
    account.storage.insert(
        PENDING_SEQUENCER,
        EvmStorageSlot::new_changed(pending_seq, pending_seq, 0),
    );
    if !pending_seq.is_zero() {
        let activation = read_registry_storage(db, SEQUENCER_ACTIVATION_BLOCK)?;
        account.storage.insert(
            SEQUENCER_ACTIVATION_BLOCK,
            EvmStorageSlot::new_changed(activation, activation, 0),
        );
        due |= block_number >= activation.saturating_to::<u64>();
    }

    Ok((due, Some(EvmState::from_iter([(SEQUENCER_REGISTRY_ADDRESS, account)]))))
}
```

Executor pushes the returned EvmState as an outcome regardless of `due`. If `due` is true, the EVM call's own `ResultAndState` will re-read the same slots — redundant but harmless; the witness tolerates duplicate records.

Alternative: move the pre-check into the executor and always contribute an outcome, so the helper never does uncaptured reads. Either shape satisfies the `PRE-BLOCK STATE CHANGE CONTRACT` in `system/AGENTS.md`.

crates/mega-evm/src/system/sequencer_registry.rs
```diff
@@ -0,0 +1,777 @@
...
+
+    if value.is_zero() {
+        return Err(BlockValidationError::BlockHashContractCall {
+            message: "SequencerRegistry deployed but _currentSystemAddress is zero".into(),
+        }
+        .into());
+    }
+
+    Ok(Address::from_word(value.into()))
+}
```
Troublor, line 316: ### Reads escape the stateless witness on every Rex5 block

This function is called at `block/executor.rs:553-555` on **every Rex5 block** (after `commit_system_call_outcomes`). It reads:

- `db.load_cache_account(SEQUENCER_REGISTRY_ADDRESS)` — account info.
- `read_registry_storage(db, CURRENT_SYSTEM_ADDRESS)` — slot 0.

The returned `Address` is stored in `MegaContext.system_address`, but the *read records* never flow through any outcome / `on_state` hook. A stateless verifier replaying the block needs to see the slot-0 SLOAD result to agree with the resolved `system_address`; today the witness is missing that read on every Rex5 block.

Fix shape: return the reads as an EvmState alongside the resolved address, and let the executor push it as a post-commit outcome:

```rust
pub fn resolve_system_address<DB: Database>(
    hardforks: impl MegaHardforks,
    spec: crate::MegaSpecId,
    db: &mut State<DB>,
) -> Result<(Address, Option<EvmState>), BlockExecutionError> {
    if !spec.is_enabled(crate::MegaSpecId::REX5) {
        return Ok((MEGA_SYSTEM_ADDRESS, None));
    }

    let bootstrap = hardforks.sequencer_registry_config().initial_system_address;
    let acc = db.load_cache_account(SEQUENCER_REGISTRY_ADDRESS)
        .map_err(BlockExecutionError::other)?;

    let Some(info) = acc.account_info() else {
        return Ok((bootstrap, None)); // bootstrap phase, no concrete read
    };

    // ... existing code-hash check ...

    let value = read_registry_storage(db, CURRENT_SYSTEM_ADDRESS)?;
    if value.is_zero() { return Err(/* existing error */); }

    let mut account = Account { info, ..Default::default() };
    account.storage.insert(
        CURRENT_SYSTEM_ADDRESS,
        EvmStorageSlot::new_changed(value, value, 0),
    );
    let state = EvmState::from_iter([(SEQUENCER_REGISTRY_ADDRESS, account)]);

    Ok((Address::from_word(value.into()), Some(state)))
}
```

Executor call-site:

```rust
let (system_address, read_state) =
    resolve_system_address(&self.hardforks, spec, self.evm.db_mut())?;
if let Some(state) = read_state {
    self.commit_system_call_outcomes(vec![MegaSystemCallOutcome {
        source: StateChangeSource::PreBlock(/* dedicated source */),
        state,
    }])?;
}
self.evm.ctx_mut().set_system_address(system_address);
```

The slot's old and new values are equal so `commit` is a no-op for state; only the `on_state` witness record is load-bearing.

(Pairs with the `is_apply_pending_changes_due` comment — same class of issue, same fix pattern.)

crates/system-contracts/contracts/SequencerRegistry.sol
```diff
@@ -0,0 +1,221 @@
...
+        uint96 fromBlock;
+        address addr;
+    }
```
Troublor: ### Move `RotationRecord` to the interface

The struct is a public data type — any caller decoding `_systemAddressRotations` / `_sequencerRotations` via the auto-generated array getters (or any future method that returns rotation records) needs this layout. Keeping it in the implementation contract forces external consumers to either redefine it or import the implementation just to get the type.

Move the declaration to `ISequencerRegistry.sol`:

```solidity
// ISequencerRegistry.sol
interface ISequencerRegistry {
    /// Historical role-change record: packed into one slot.
    struct RotationRecord {
        uint96 fromBlock;
        address addr;
    }

    // ... existing methods, events, errors ...
}
```

In `SequencerRegistry.sol`, drop the duplicate struct and reference `ISequencerRegistry.RotationRecord` where it's used (e.g. `ISequencerRegistry.RotationRecord[] private _systemAddressRotations`).

Composes cleanly with the separate "rotation → change" naming suggestion — if that lands, the struct moves to the interface as `ChangeRecord`.

docs/spec/evm/gas-detention.md
```diff
@@ -82,7 +82,7 @@ The following rules MUST apply:
 - `STATICCALL` to the oracle contract address alone MUST NOT trigger oracle detention.
 - Oracle detention is triggered by storage reads, not by message-call targeting alone.
 - `DELEGATECALL` to the oracle contract MUST NOT trigger oracle detention solely by virtue of targeting the oracle address, because `SLOAD` in a `DELEGATECALL` context reads the caller's storage, not the oracle contract's storage.
-- If the transaction sender is [`MEGA_SYSTEM_ADDRESS`](../system-contracts/system-tx.md), oracle gas detention MUST NOT be applied.
+- If the transaction sender is [`SYSTEM_ADDRESS`](../system-contracts/system-tx.md), oracle gas detention MUST NOT be applied.
```
Troublor: ### Keep `MEGA_SYSTEM_ADDRESS` as the single name; add it to the glossary

The rename to `SYSTEM_ADDRESS` across `gas-detention.md`, `oracle.md`, `overview.md`, `sequencer-registry.md`, `rex5.md`, and `system-tx.md § System Address` splits the documentation into two parallel terms where one suffices. `MEGA_SYSTEM_ADDRESS` already names "the authorized protocol-maintenance sender"; Rex5 changes how that value is *resolved* per block, not what it means. A single term with a per-spec resolution rule is clearer than two terms bound by a cross-spec policy — and avoids forward references (Rex3 pages referencing a Rex5-introduced term).

Concrete asks:

**1. Revert the rename.** Everywhere this PR swapped `MEGA_SYSTEM_ADDRESS` → `SYSTEM_ADDRESS`, restore the original name. Files: `gas-detention.md`, `oracle.md`, `overview.md`, `sequencer-registry.md`, `rex5.md`, and the new `§ System Address` prose in `system-tx.md`.

**2. Document the Rex5 resolution in one place** (`system-tx.md`). Keep the Constants table entry unchanged (the literal fallback value). Replace the `§ System Address` section with a `§ Resolution` subsection that uses `MEGA_SYSTEM_ADDRESS` throughout:

> From [Rex5](../upgrades/rex5.md) onward, `MEGA_SYSTEM_ADDRESS` is resolved per block from [`SequencerRegistry.currentSystemAddress()`](sequencer-registry.md) after all pre-block changes are committed. Pre-Rex5 blocks use the fixed constant in the table above.

**3. Add a glossary entry** in `docs/spec/glossary.md`, and bump its front-matter `spec:` from `Rex4` to `Rex5`:

````markdown
## `MEGA_SYSTEM_ADDRESS`

The authorized sender for [Mega System Transactions](system-contracts/system-tx.md).

- Pre-[Rex5](upgrades/rex5.md): the fixed constant `0xA887…BD1d` (defined in [`system-tx.md § Constants`](system-contracts/system-tx.md#constants)).
- Rex5 onward: resolved per block from [`SequencerRegistry.currentSystemAddress()`](system-contracts/sequencer-registry.md).

Exempt from oracle [gas detention](evm/gas-detention.md); carries [Oracle](system-contracts/oracle.md) write authority.
````

**4. Proper reference links.** Raw-code-font occurrences of `MEGA_SYSTEM_ADDRESS` across the spec today mostly don't link anywhere. Once the glossary entry exists, update references to link to it (or to `system-tx.md § Constants`, whichever is canonical — pick one and use consistently).

Net: one name, one authoritative definition, one place explaining the per-spec resolution, consistent linking. The `SYSTEM_ADDRESS` umbrella — which today lives only in prose and has no Constants / glossary entry — goes away.

File: docs/spec/system-contracts/oracle.md [hunk 10]
```diff
@@ -48,6 +49,16 @@ Deployed bytecode:
...
 ```
 
+#### Version 2.0.0
+
+Since: [Rex5](../upgrades/rex5.md)
```
Troublor, line 56: ### Mark Rex5 as unstable inline on every page that documents Rex5 content

Today only [`upgrades/rex5.md`](../upgrades/rex5.md) carries the "Unstable — under active development" blockquote. Every other page that documents Rex5 behavior bumps its front-matter `spec:` to `Rex5` without any inline signal to the reader.

Affected pages:

- [`system-contracts/oracle.md`](oracle.md) — mixes stable Version 1.0.0 (MiniRex) and Version 1.1.0 (Rex2) with Version 2.0.0 (Rex5). A reader arriving directly at `#### Version 2.0.0` has no indication that its semantics may still change.
- [`system-contracts/sequencer-registry.md`](sequencer-registry.md) — entire contract is Rex5-new. Front-matter says `spec: Rex5`; no inline marker.
- [`system-contracts/system-tx.md`](system-tx.md) — front-matter bumped to Rex5; the new prose on dynamic resolution is Rex5 behavior.
- [`system-contracts/overview.md`](overview.md) — adds the `SequencerRegistry` row + `SEQUENCER_REGISTRY_ADDRESS` constant with no Rex5 marker.
- [`evm/gas-detention.md`](../evm/gas-detention.md) — the rename to `SYSTEM_ADDRESS` introduces Rex5 terminology into a stable-spec page.

Suggested fix: reuse the blockquote pattern already in `rex5.md`:

```markdown
> **Unstable** — Rex5 semantics described here are under active development and may change before network activation.
```

For pages that mix stable and Rex5 content (e.g., `oracle.md`), place the blockquote directly under the Rex5-specific subsection heading (right below `#### Version 2.0.0`, below the SequencerRegistry table row in `overview.md`, etc.) rather than at the page top — so stable content isn't visually downgraded while Rex5 content is clearly flagged.

For wholly-Rex5 pages (`sequencer-registry.md`), top of page is fine.

Without these inline markers, a reader arriving at a deep anchor or following a `SUMMARY.md` link has no way to tell they're looking at in-development content versus ratified spec.

Troublor (commented):
8 comments across three themes.

- **Witness correctness** (2). `is_apply_pending_changes_due` and `resolve_system_address` read registry storage without surfacing the reads in any outcome, so the stateless witness is incomplete — the first on the common "no rotation due" path, the second on every Rex5 block. Both threads include concrete fix sketches that return the reads as an `EvmState` for the executor to push through `on_state`.
- **Architecture / API** (5). Per-fork typed params on `MegaHardforks` via a `HardforkParams` trait + `ForkEntry`-embedded config (instead of spec-specific trait methods + a floating default); move `RotationRecord` into `ISequencerRegistry` so external consumers can decode rotation arrays; drop `Default` for `SequencerRegistryConfig` since the defaults silently collapse all three roles to `MEGA_SYSTEM_ADDRESS`; revert the `MEGA_SYSTEM_ADDRESS → SYSTEM_ADDRESS` rename, formalize the Rex5 resolution in one place, and add a glossary entry with proper reference links; rename `rotation → change` internally for consistency with the external ABI (`scheduleNext*Change`, `applyPendingChanges`, `*ChangeScheduled` events).
- **Docs** (1). Mark Rex5 as unstable inline on every page that documents Rex5 content (oracle.md § v2.0.0, sequencer-registry.md, system-tx.md, overview.md row, gas-detention.md rename). The existing blockquote in `upgrades/rex5.md` doesn't propagate to pages a reader reaches via deep anchors or `SUMMARY.md`.

Session 1 status: the two-role + pending/apply architectural suggestion landed and went further (bootstrap-seeded chain params) — thread resolved. The ERC-7201 storage-layout suggestion is partially addressed (slot constants moved to a named module), but the Solidity layout is still implicit and the stronger Rust/Solidity parity test was removed; keeping that thread open.

Troublor (commented):
Session 3 follow-up on `6861dbe` and `d39b6d0`: all 8 comments from my previous round are addressed and resolved.

- **Witness recording** (`is_apply_pending_changes_due`, `resolve_system_address`) — both refactored to return `EvmState` alongside their result; reads now flow through `system_caller.on_state` as suggested.
- **`HardforkParams` + `ForkEntry`** — clean landing of the proposed shape (per-fork typed params via `const FORK`, `with_params<P>` builder, default-impl typed accessor, no Ext-trait split).
- **Naming and docs** — `rotation → change` sweep, `ChangeRecord` moved to the interface, `MEGA_SYSTEM_ADDRESS` rename revert with a new glossary entry and proper links, inline `Unstable — Rex5` markers added to every affected page.
- **`SequencerRegistryConfig`** — `Default` removed; Rex5 activation now fails closed when params aren't configured.

No new blockers. Nice follow-through.

Troublor (approved):

PR #265, opened 2026-04-21
docs(mega-evme): document --rpc.capture-file and --rpc.replay-file

### LINKED ISSUE #262: docs(mega-evme): document `--rpc.cache-file` replay mode
## Gap

`mega-evme replay` gained a `--rpc.cache-file` flag in megaeth-labs/mega-evm#251 (merged 2026-04-20), but `docs/mega-evme/commands/replay.md` does not mention it. Users have no way to discover the feature or learn capture vs. replay semantics from the docs.

## What changed in #251

- New transport-level RPC capture/replay path that persists full JSON-RPC request/response fixtures to a file.
- Captured fixtures include external env data (bucket-capacity inputs), so offline replay works without extra CLI flags.
- Provider construction was refactored to separate standard RPC, capture mode, and replay mode.

## What "done" looks like

Update `docs/mega-evme/commands/replay.md` (and cross-link from the RPC cache / retry section if appropriate) to cover:

- `--rpc.cache-file <path>` flag syntax and where it fits in the options list.
- **Capture mode** — first run against a live RPC endpoint saves every JSON-RPC request/response to `<path>`.
- **Replay mode** — subsequent runs read from `<path>` and execute fully offline (no network access required).
- That env data (bucket-capacity inputs) is threaded through the fixture, so offline replay does not require `--bucket-capacity` to be re-specified.
- At least one example invocation for each mode.

## References

- PR: megaeth-labs/mega-evm#251 — `feat: add replay cache-file mode`
- Upstream audit: megaeth-labs/auto-audits#11 (finding F-001, P2)
- Target file: [`docs/mega-evme/commands/replay.md`](https://github.com/megaeth-labs/mega-evm/blob/main/docs/mega-evme/commands/replay.md)

Troublor:
Adds documentation for the transport-level JSON-RPC fixture mechanism introduced in #251.

- documents `--rpc.capture-file` capture mode
- documents `--rpc.replay-file` offline replay mode
- explains how bucket-capacity env data is embedded in the fixture
- adds example invocations for both modes
- updates the options cross-reference and full help output

Closes #262

### DIFF
--- a/docs/mega-evme/commands/replay.md
+++ b/docs/mega-evme/commands/replay.md
@@ -141,14 +141,9 @@ See the linked pages for full details.
 ## Examples
 
 In all examples below, replace `0xabc123...` with a real transaction hash from MegaETH mainnet.
+Every `replay` invocation requires one of `--rpc <URL>`, `--rpc.capture-file <PATH>`, or `--rpc.replay-file <PATH>`.
 
-**Basic replay**
-
-```bash
-mega-evme replay 0xabc123...
-```
-
-**Replay against a custom RPC**
+**Replay against a live RPC**
 
 ```bash
 mega-evme replay --rpc https://mainnet.megaeth.com/rpc 0xabc123...
@@ -157,25 +152,25 @@ mega-evme replay --rpc https://mainnet.megaeth.com/rpc 0xabc123...
 **Replay with call tracing**
 
 ```bash
-mega-evme replay --trace --tracer call 0xabc123...
+mega-evme replay --rpc https://mainnet.megaeth.com/rpc --trace --tracer call 0xabc123...
 ```
 
 **Replay with a tighter gas limit**
 
 ```bash
-mega-evme replay --override.gas-limit 50000 0xabc123...
+mega-evme replay --rpc https://mainnet.megaeth.com/rpc --override.gas-limit 50000 0xabc123...
 ```
 
 **Replay with different calldata**
 
 ```bash
-mega-evme replay --override.input 0xdeadbeef 0xabc123...
+mega-evme replay --rpc https://mainnet.megaeth.com/rpc --override.input 0xdeadbeef 0xabc123...
 ```
 
 **Replay under a specific spec**
 
 ```bash
-mega-evme replay --override.spec Rex2 0xabc123...
+mega-evme replay --rpc https://mainnet.megaeth.com/rpc --override.spec Rex2 0xabc123...
 ```
 
 ## See Also
--- a/docs/mega-evme/commands/replay.md
+++ b/docs/mega-evme/commands/replay.md
@@ -4,7 +4,9 @@ description: Fetch and re-execute an on-chain transaction with optional override
 
 # replay
 
-Re-execute a historical transaction locally using state and block context fetched from an RPC endpoint.
+Re-execute a historical transaction locally using an RPC endpoint or a previously captured fixture file.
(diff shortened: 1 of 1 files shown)


flyq (approved):
LGTM

docs/mega-evme/commands/replay.md
```diff
@@ -34,6 +34,57 @@ Any standard Ethereum JSON-RPC provider works here.
 mega-evme replay --rpc https://mainnet.megaeth.com/rpc <TX_HASH>
 ```
 
+## RPC Cache File
+
+`mega-evme replay` supports a transport-level JSON-RPC fixture mechanism that records every request/response pair to a single file and serves them back on later runs without touching the network.
+It is useful for pinning a reproducible replay (e.g. for regression tests, debugging sessions, or offline review) and for running `replay` in environments that cannot reach the RPC endpoint.
+
+Unlike the generic [RPC Cache](../configuration/state-management.md#rpc-cache-and-retry), which is keyed on a small allow-list of cacheable methods and stored per chain under the platform cache directory, the cache file covers **every** JSON-RPC call issued during the replay and lives at a user-chosen path.
```
RealiCZ: This overstates the implementation a bit. `--rpc.capture-file` only records single JSON-RPC requests; batch requests are explicitly forwarded without caching. Could we say `every single (non-batch) JSON-RPC call` here, or just drop `every`?

docs/mega-evme/commands/replay.md
```diff
@@ -34,6 +34,57 @@ Any standard Ethereum JSON-RPC provider works here.
...
+
+The mechanism has two modes, selected by two mutually exclusive flags.
+
+### `--rpc.capture-file <PATH>`
+
+Capture mode. Requires `--rpc`.
+
+On the first run, every JSON-RPC request and response issued while serving the replay is appended to `<PATH>`.
```
RealiCZ: `appended to <PATH>` sounds like we stream entries incrementally to disk, but that is not what the implementation does. The requests/responses are accumulated in memory and the full envelope is persisted atomically on clean exit. I think `recorded` / `captured` would be more accurate than `appended`.

docs/mega-evme/commands/replay.md
RealiCZ: The top-level description still reads as if `replay` always fetches data from a live RPC endpoint, but `--rpc.replay-file` is a fully offline mode. Could we reword this section so it covers both RPC-backed replay and offline replay-file mode?

docs/mega-evme/commands/replay.md
```diff
@@ -140,6 +193,21 @@ Options:
           RPC URL to fetch transaction from
           [env: RPC_URL=] [default: http://localhost:8545] [aliases: --rpc-url]
```
RealiCZ: This help block looks stale relative to the current CLI. `--rpc` no longer has an `RPC_URL` env var or a default `http://localhost:8545`, and the full `replay --help` output now includes additional RPC options as well. I think this block should be regenerated from the actual current help output instead of updated manually.

RealiCZ (changes requested):

docs/mega-evme/commands/replay.md
RealiCZ: These examples look stale relative to the current command behavior. `mega-evme replay` now requires one of `--rpc <URL>`, `--rpc.capture-file <PATH>`, or `--rpc.replay-file <PATH>`, so the bare examples below are not runnable as written.

RealiCZ (approved):

PR #267, opened 2026-04-21
feat: pack `RotationRecord` in one storage slot

davidperk:
## Summary

Packs each `RotationRecord` in `SequencerRegistry` into a single storage word (`uint96 fromBlock` + `address sequencer`) so `sequencerAt` loads one slot per history step instead of two.

## Motivation

Rotation history is walked in reverse on every `sequencerAt` query. Unpacked `uint256` + `address` used two slots per element, doubling storage reads along the hot path. Packing cuts that cost without changing observable behavior for valid schedules.

## Changes

- **`SequencerRegistry`**: `RotationRecord` now uses `uint96 fromBlock` and `address sequencer` in one slot; `applyPendingChange` still stores `_activationBlock` as `uint256` and casts when appending history.
- **Scheduling rule**: For non-cancel calls, `activationBlock` must be `<= type(uint96).max`; otherwise the contract reverts with `ActivationBlockTooLarge()`. Cancel via `activationBlock == type(uint256).max` and `newSequencer == address(0)` is unchanged.
- **`ISequencerRegistry`**: Documents the new error and the `uint96` bound on schedules.
- **Tests**: `SequencerRegistry.t.sol` covers the too-large activation revert; existing rotation and `sequencerAt` tests still pass.
- **Artifacts**: Regenerated `SequencerRegistry` deployed bytecode and code hash (`SequencerRegistry-1.0.0.json`, `SequencerRegistry-latest.json`, and `src/generated/sequencer_registry_artifacts.rs`). Semver string from `version()` remains **1.0.0**; bytecode hash updates reflect the layout/behavior change.
- **Docs**: `docs/spec/system-contracts/sequencer-registry.md` updated for the new error, scheduling bound, packed history row, and current code hash (Prettier applied per project docs rules).

## Testing

- `forge test --match-contract SequencerRegistryTest`
- `cargo build -p mega-system-contracts` (forge validates bytecode against committed artifacts)

## Notes

- **ABI**: `scheduleNextSequencerChange` and `sequencerAt` selectors are unchanged; only revert conditions and gas behavior differ for invalid large `activationBlock` values.
- **Honest paths**: Any activation that can be written through `scheduleNextSequencerChange` already fits `uint96`; the explicit revert mainly documents the limit and rejects absurd inputs early.

---


Troublor (approved):
LGTM, Ty

RealiCZ:
LGTM，ty

PR #270, opened 2026-04-23
fix(sandbox): reject trailing bytes in keyless tx RLP decoding

### LINKED ISSUE #269: Reject trailing bytes in KeylessDeploy transaction decoding
`decode_keyless_tx()` currently calls `TxLegacy::rlp_decode_signed(&mut buf)` but does not verify that the RLP decoder consumed the full input.

  This means `keylessDeploy(bytes,uint256)` may accept a valid pre-EIP-155 CREATE transaction with arbitrary trailing bytes appended.
The trailing bytes are not signed transaction data, but the original raw bytes are still used as `sandbox_tx.enveloped_tx`, which can affect enveloped-tx / L1-data-fee accounting.

TroublorBot:

## Summary

Reject trailing bytes in KeylessDeploy RLP transaction decoding to prevent unsigned bytes from leaking into L1 data-fee accounting. The `decode_keyless_tx()` function now verifies that the RLP decoder consumed the entire input buffer, returning `MalformedEncoding` if trailing bytes are present. Includes updated spec documentation and new unit tests covering both single and multi-byte trailing content.

Fixes #269


RealiCZ:
@TroublorBot @claude
Can you explain why this does not need a Rex5 gate? Since KeylessDeploy has existed since Rex2, rejecting trailing bytes changes validation behavior for Rex2-Rex4 too. Is there any reason this cannot cause historical replay mismatch?

RealiCZ:
@TroublorBot go

RealiCZ:
@TroublorBot  please implement and push the Rex5-gated version.

RealiCZ (approved):

Troublor (approved):

PR #272, opened 2026-04-23
refactor(limit): encapsulate CallFrameInfo parent-flag logic in frame_limit (#271)

### LINKED ISSUE #271: Refactor: encapsulate `CallFrameInfo` parent-flag logic inside `frame_limit.rs`
## Summary

Follow-up to #266. The Rex5 parent-flag dedup logic added `charged_parent_update` to `CallFrameInfo` and spread the check/mutate/unwind pattern across `DataSizeTracker` and `KVUpdateTracker`, which both hold a `FrameLimitTracker<CallFrameInfo>`. The duplication pre-dates #266 but the fix widens it. This issue tracks a cleanup refactor.

### 1. Hide `target_updated` and `charged_parent_update` inside `frame_limit.rs`

Both fields are part of an invariant ("child sets parent's flag on push, unsets it on revert if this child set it") that is conceptually owned by `FrameLimitTracker<CallFrameInfo>`, not by its users. Make them private to `frame_limit.rs` and expose typed methods instead:

- `push_intercept_frame()` — intercept/synthetic frame.
- `push_call_frame(target, has_transfer) -> bool` — returns `parent_needs_update`.
- `push_create_frame() -> bool` — returns `parent_needs_update`.
- `set_created_address(addr)` — called from `after_frame_init_on_frame`.
- `pop_frame_unwind_parent(success)` — pop with revert-unwind of parent flag.

After the refactor, the trackers only compose these calls with their own charge amounts (40 bytes for data size, 1 for KV). A grep for `target_updated` / `charged_parent_update` should return only `frame_limit.rs`.

### 2. Track spec internally on `FrameLimitTracker` to avoid `rex5_enabled` leakage

Currently each tracker caches `rex5_enabled: bool` and threads it into every push call. Move this into `FrameLimitTracker<CallFrameInfo>` so callers don't plumb spec flags across the API boundary. Exact shape TBD — either store a `MegaSpecId` / bool on the tracker at `new()` time, or keep the data structure spec-agnostic and introduce a thin wrapper that owns the spec. The goal is that the spec-gating of the parent-flag mutation is invisible to `DataSizeTracker` / `KVUpdateTracker`.

## Out of Scope

- Behavior changes. This is a pure refactor; Rex4 and Rex5 semantics must remain identical.
- Merging the two trackers. `DataSizeTracker` and `KVUpdateTracker` stay as distinct types; only the shared frame-info mechanics move.

## Context

- PR #266 introduced the new field and the revert-unwind logic.
- `CallFrameInfo`'s doc already says it's "for trackers that need account update deduplication", so centralizing the behavior next to the data is a natural next step.

TroublorBot:

## Summary

Hides `target_updated` and `charged_parent_update` behind typed methods on `FrameLimitTracker<CallFrameInfo>` (`push_call_frame`, `push_create_frame`, `pop_frame_unwind_parent`) so `DataSizeTracker` and `KVUpdateTracker` no longer duplicate the check/mutate/unwind pattern. The spec is now stored on the tracker itself, eliminating `rex5_enabled` leakage to callers. Pure refactor; Rex4/Rex5 semantics unchanged.

Fixes #271

### DIFF
--- a/crates/mega-evm/src/limit/frame_limit.rs
+++ b/crates/mega-evm/src/limit/frame_limit.rs
@@ -394,21 +394,10 @@ mod tests {
 
     const ADDR: Address = address!("0000000000000000000000000000000000001234");
 
-    fn tracker(rex4: bool, rex5: bool) -> FrameLimitTracker<CallFrameInfo> {
-        let spec = if rex5 {
-            MegaSpecId::REX5
-        } else if rex4 {
-            MegaSpecId::REX4
-        } else {
-            MegaSpecId::EQUIVALENCE
-        };
-        FrameLimitTracker::new(spec, u64::MAX)
-    }
-
     /// `set_created_address` on an empty frame stack must be a no-op.
     #[test]
     fn test_set_created_address_empty_stack_is_noop() {
-        let mut t = tracker(false, false);
+        let mut t = FrameLimitTracker::<CallFrameInfo>::new(MegaSpecId::EQUIVALENCE, u64::MAX);
         // No frame on the stack — should not panic, just do nothing.
         t.set_created_address(ADDR);
     }
@@ -418,7 +407,7 @@ mod tests {
     #[test]
     #[should_panic(expected = "created account already recorded")]
     fn test_set_created_address_duplicate_panics() {
-        let mut t = tracker(false, false);
+        let mut t = FrameLimitTracker::<CallFrameInfo>::new(MegaSpecId::EQUIVALENCE, u64::MAX);
         t.push_create_frame();
         t.set_created_address(ADDR);
         t.set_created_address(ADDR); // second call must panic
--- a/crates/mega-evm/src/limit/frame_limit.rs
+++ b/crates/mega-evm/src/limit/frame_limit.rs
@@ -414,7 +414,7 @@ mod tests {
     }
 
     /// `set_created_address` panics when called twice on the same CREATE frame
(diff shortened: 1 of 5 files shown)


crates/mega-evm/src/limit/frame_limit.rs
```diff
@@ -206,6 +226,122 @@ impl<I> FrameLimitTracker<I> {
...
+        }
+    }
+
+    /// Pushes a synthetic frame for inspector / access-control interception paths.
+    ///
+    /// No parent-flag mutation: an intercepted call is short-circuited before any
+    /// state-modifying account update is recorded, so the parent does not need to be marked.
+    /// The frame exists only to keep the per-tracker frame stack aligned with the EVM's call
+    /// stack so that the matching `pop_frame_unwind_parent` finds an entry to pop.
+    pub(crate) fn push_intercept_frame(&mut self) {
```
Troublor: Consider renaming this to `push_dummy_frame` — the method doesn't actually do anything interception-specific; it just pushes a no-op frame to keep the stack aligned. `dummy` reads more accurately at the call sites (`data_size.rs:183`, `kv_update.rs:137`) where the context is the `push_empty_frame` hook.

```suggestion
    pub(crate) fn push_dummy_frame(&mut self) {
```

(Callers in `data_size.rs` and `kv_update.rs` would need the matching rename.)

Troublor (approved):

RealiCZ (approved):
LGTM

PR #277, opened 2026-04-27
fix(system): make system contract validation and interceptor dispatch invariants explicit

### LINKED ISSUE #274: fix: make system contract validation and interceptor dispatch invariants explicit
## Summary

A few system contract and interceptor paths rely on implicit invariants that are currently safe, but brittle. Hardening these would make misconfiguration fail earlier and make dispatch behavior more robust against future changes.

## Affected areas

- `crates/mega-evm/src/system/sequencer_registry.rs`
- `crates/mega-evm/src/system/intercept.rs`
- `crates/mega-evm/src/evm/execution.rs`

### Sequencer registry deploy accepts zero-address config fields

`transact_deploy_sequencer_registry` seeds `SequencerRegistry` storage directly from `SequencerRegistryConfig`.

The config should reject zero values for:

- `initial_system_address`
- `initial_sequencer`
- `initial_admin`

Potential failure modes:

- `initial_admin = address(0)` can permanently lock all admin-only registry operations, because no caller can satisfy the admin check.
- `initial_system_address = address(0)` can cause later system-address resolution to fail after deployment has already succeeded.
- `initial_sequencer = address(0)` creates an invalid initial sequencer state.

Suggested fix: validate these fields at deploy time and return a block validation/execution error if any are zero.

### Oracle hint interceptor return value is discarded

`dispatch_system_contract_interceptors` calls `OracleHintInterceptor::intercept` for side effects and discards the returned `Option<FrameResult>`.

That is correct today if the oracle hint interceptor is guaranteed to never short-circuit. However, the dispatch shape differs from the other interceptors, so a future change that returns `Some(FrameResult)` could be silently ignored.

Suggested fix: make the invariant explicit, for example with a `debug_assert!(...is_none())` and a short comment, or by changing the interceptor API if it is intentionally side-effect-only.

### CALL/STATICCALL policy for system interceptors is implicit

The system contract spec expects interception for `CALL` and `STATICCALL`, but not `DELEGATECALL` or `CALLCODE`.

Current behavior appears to rely on current call input address semantics rather than an explicit call-scheme guard. If upstream call-frame semantics change, interceptors could accidentally start applying to schemes that should be excluded.

Suggested fix: add an explicit scheme check at the dispatch site so only `CALL` and `STATICCALL` enter system contract interceptor dispatch.

## Impact

Low severity overall. These are mostly defense-in-depth and maintainability improvements:

- zero-address validation catches bad genesis/configuration at the earliest point;
- oracle hint dispatch becomes future-proof against accidental short-circuit behavior;
- call-scheme filtering makes the system contract policy explicit and less dependent on implementation details.

## Suggested remediation

- Reject zero-address fields in `SequencerRegistryConfig` during registry deployment.
- Document or assert that `OracleHintInterceptor` does not short-circuit.
- Gate system interceptor dispatch on explicit call schemes: `CALL` and `STATICCALL` only.
- Add focused regression tests for invalid registry config and excluded call schemes.

TroublorBot:

## Summary

Hardens system contract initialization and interception dispatch with three defense-in-depth improvements: (1) reject zero-address config fields in SequencerRegistry deployment to catch misconfiguration early, (2) assert that OracleHintInterceptor never short-circuits to prevent silent breakage if the invariant is violated, and (3) add an explicit call-scheme guard at the interceptor dispatch site so policy no longer depends on implicit target_address semantics. All changes include regression tests and pass the full test suite.

Fixes #274

### DIFF
--- a/crates/mega-evm/src/evm/execution.rs
+++ b/crates/mega-evm/src/evm/execution.rs
@@ -637,17 +637,25 @@ where
         // Short-circuiting paths return Some(FrameResult).
         // These synthetic results skip `AdditionalLimit::before_frame_init`; we only push an
         // empty tracking frame to keep the additional-limit stacks aligned.
+        //
+        // Only `CALL` and `STATICCALL` enter interceptor dispatch. `CALLCODE` and
+        // `DELEGATECALL` execute in the caller's state context and use the caller's
+        // address as `target_address`, so they would never match a system contract
+        // address today — the explicit scheme guard makes this policy independent of
+        // upstream call-frame semantics.
         if let FrameInput::Call(call_inputs) = &frame_init.frame_input {
-            if let Some(result) =
-                dispatch_system_contract_interceptors(self.ctx(), call_inputs, frame_init.depth)
-            {
-                // Push an empty frame to keep the limit tracker stack balanced:
-                // `frame_return_result` / `last_frame_result` will pop a frame, but
-                // `after_frame_init` (which normally pushes) was skipped.
-                if is_mini_rex_enabled {
-                    additional_limit.borrow_mut().push_empty_frame();
+            if matches!(call_inputs.scheme, CallScheme::Call | CallScheme::StaticCall) {
+                if let Some(result) =
+                    dispatch_system_contract_interceptors(self.ctx(), call_inputs, frame_init.depth)
+                {
+                    // Push an empty frame to keep the limit tracker stack balanced:
+                    // `frame_return_result` / `last_frame_result` will pop a frame, but
+                    // `after_frame_init` (which normally pushes) was skipped.
+                    if is_mini_rex_enabled {
+                        additional_limit.borrow_mut().push_empty_frame();
+                    }
+                    return Ok(FrameInitResult::Result(result));
                 }
-                return Ok(FrameInitResult::Result(result));
             }
         }
 
--- a/crates/mega-evm/src/evm/execution.rs
+++ b/crates/mega-evm/src/evm/execution.rs
@@ -639,10 +639,10 @@ where
         // empty tracking frame to keep the additional-limit stacks aligned.
         //
         // Only `CALL` and `STATICCALL` enter interceptor dispatch. `CALLCODE` and
-        // `DELEGATECALL` execute in the caller's state context and use the caller's
-        // address as `target_address`, so they would never match a system contract
-        // address today — the explicit scheme guard makes this policy independent of
-        // upstream call-frame semantics.
+        // `DELEGATECALL` execute in the caller's state context, so intercepting them
+        // would apply system-contract logic in the wrong state context — the scheme
+        // guard enforces this policy explicitly rather than relying on upstream
+        // call-frame semantics to keep `target_address` distinct.
         if let FrameInput::Call(call_inputs) = &frame_init.frame_input {
             if matches!(call_inputs.scheme, CallScheme::Call | CallScheme::StaticCall) {
                 if let Some(result) =
--- a/crates/mega-evm/src/system/sequencer_registry.rs
+++ b/crates/mega-evm/src/system/sequencer_registry.rs
@@ -118,6 +118,29 @@ pub fn transact_deploy_sequencer_registry<DB: Database>(
         return Ok(None);
     }
 
+    // Reject zero-address config fields up front. A zero `initial_admin` would permanently
+    // lock all admin-only registry operations, a zero `initial_system_address` would break
+    // later system-address resolution, and a zero `initial_sequencer` produces an invalid
+    // initial sequencer state.
+    if config.initial_system_address.is_zero() {
+        return Err(BlockValidationError::BlockHashContractCall {
+            message: "SequencerRegistryConfig.initial_system_address must not be zero".into(),
+        }
+        .into());
+    }
+    if config.initial_sequencer.is_zero() {
+        return Err(BlockValidationError::BlockHashContractCall {
+            message: "SequencerRegistryConfig.initial_sequencer must not be zero".into(),
+        }
+        .into());
+    }
+    if config.initial_admin.is_zero() {
+        return Err(BlockValidationError::BlockHashContractCall {
+            message: "SequencerRegistryConfig.initial_admin must not be zero".into(),
+        }
+        .into());
+    }
+
     let acc =
         db.load_cache_account(SEQUENCER_REGISTRY_ADDRESS).map_err(BlockExecutionError::other)?;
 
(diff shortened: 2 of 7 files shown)


crates/mega-evm/src/system/sequencer_registry.rs
```diff
@@ -118,6 +118,29 @@ pub fn transact_deploy_sequencer_registry<DB: Database>(
...
+            message: "SequencerRegistryConfig.initial_sequencer must not be zero".into(),
+        }
+        .into());
+    }
+    if config.initial_admin.is_zero() {
+        return Err(BlockValidationError::BlockHashContractCall {
+            message: "SequencerRegistryConfig.initial_admin must not be zero".into(),
+        }
+        .into());
+    }
```
Troublor: Suggest moving this validation up to the `HardforkParams` trait so it becomes a standard contract for all per-fork param types, not a per-call-site discipline.

Concretely:

- Add `fn validate(&self) -> Result<(), HardforkParamsError>` to `HardforkParams` (`crates/mega-evm/src/block/hardfork.rs:64`) with a default `Ok(())` impl. Existing/future impls opt in by overriding.
- Call it from `MegaHardforkConfig::with_params` (`crates/mega-evm/src/block/hardfork.rs:278`), panicking on `Err` to match the existing missing-fork panic style on line 282.
- `SequencerRegistryConfig::validate` performs the three `is_zero` checks here.
- The deploy-time block becomes either removable or a one-line belt-and-braces guard pointing at `validate`.

Benefits: misconfiguration fails at chain-config load time (genesis) rather than on the block where Rex5 first activates — possibly hours into a live run; future hardfork param types automatically opt into the validation contract.

Scope note: this restricts `validate` to **construction-time invariants** (no cross-field/cross-fork context). If a future params type needs context-dependent validation, the trait can be extended then.

crates/mega-evm/tests/rex4/limit_control.rs
```diff
@@ -668,6 +668,73 @@ fn test_inspector_sees_remaining_compute_gas_system_call() {
...
+    assert!(result.result.is_success(), "outer tx should succeed");
+
+    let output = result.result.output().expect("should have output");
+    let success_flag = U256::from_be_slice(output);
+    assert_eq!(
+        success_flag,
+        U256::from(0),
+        "DELEGATECALL to system contract must not be intercepted — scheme guard must reject it"
+    );
+}
```
Troublor: These two new tests duplicate the pre-existing `test_callcode_not_intercepted` (line 770) and `test_delegatecall_not_intercepted` (line 740) immediately below. Same bytecode (`mstore` + `append_callcode`/`append_delegatecall` + `MSTORE` + `RETURN`), same DB setup, same `success_flag == 0` assertion — only the test name and the assertion message string differ.

The intent is good (document the scheme-guard's purpose by name), but the new tests **can't actually isolate the guard**: even without the new `matches!(scheme, Call | StaticCall)` check in `frame_init`, both `CALLCODE` and `DELEGATECALL` would still fall through to bytecode because their `target_address` resolves to the caller, not `LIMIT_CONTROL_ADDRESS` — so the existing tests already cover this observable behavior. A genuine regression test for the new guard would need to either:

1. Unit-test `dispatch_system_contract_interceptors` directly with crafted `CallInputs` where `scheme = CallCode` and `target_address = LIMIT_CONTROL_ADDRESS` (impossible to construct via real EVM execution, but possible in a unit test), or
2. Assert on a side-effect that only the guard would prevent (e.g., a counter inside an interceptor stub).

Suggested action: delete these two tests as redundant. If you want the guard's intent named in source, add a one-line comment to `test_callcode_not_intercepted` / `test_delegatecall_not_intercepted` like `// Also covers the explicit CallScheme guard at execution.rs:647.`

crates/mega-evm/src/evm/execution.rs
```diff
@@ -637,17 +637,25 @@ where
         // Short-circuiting paths return Some(FrameResult).
         // These synthetic results skip `AdditionalLimit::before_frame_init`; we only push an
         // empty tracking frame to keep the additional-limit stacks aligned.
+        //
+        // Only `CALL` and `STATICCALL` enter interceptor dispatch. `CALLCODE` and
+        // `DELEGATECALL` execute in the caller's state context and use the caller's
+        // address as `target_address`, so they would never match a system contract
+        // address today — the explicit scheme guard makes this policy independent of
+        // upstream call-frame semantics.
```
Troublor: Minor: the "would never match a system contract address today" claim is contingent rather than structural. It's true that in normal execution `target_address = caller` for `CALLCODE`/`DELEGATECALL`, but the absence of a match also depends on no system contract ever issuing a `CALLCODE`/`DELEGATECALL` to itself — a property that holds today only because none of the current system contract bytecodes do so. Future system contracts could change that, at which point the comment would be wrong but the guard would still be correct.

Suggest dropping the "would never match anyway" half-sentence, since it implies the guard is incidental belt-and-braces. The policy reason stands on its own:

```suggestion
        // Only `CALL` and `STATICCALL` enter interceptor dispatch. `CALLCODE` and
        // `DELEGATECALL` execute in the caller's state context, so intercepting
        // them would apply system-contract logic in the wrong state context — the
        // scheme guard enforces this policy explicitly rather than relying on
        // upstream call-frame semantics to keep `target_address` distinct.
```

Troublor (changes requested):
Three follow-up suggestions, all maintainability-focused (no correctness blockers):

1. **Hoist zero-address validation to a `HardforkParams::validate` trait method** so misconfiguration fails at chain-config load time rather than on the Rex5 activation block.
2. **Drop the two new scheme-guard tests** — they duplicate the pre-existing `test_callcode_not_intercepted` / `test_delegatecall_not_intercepted` and can't actually isolate the new guard from the existing `target_address` mismatch.
3. **Tighten the scheme-guard rationale comment** — the "would never match a system contract address today" claim is contingent rather than structural and weakens the policy framing.

The hardening itself is clean: zero-address validation is well-placed, the `OracleHintInterceptor` `debug_assert!` makes the side-effect-only invariant explicit, and the explicit `CallScheme` guard correctly decouples policy from upstream call-frame semantics. Spec doc paragraph is accurate and matches the in-code comment.

Troublor:
@TroublorBot please address the latest review feedback

Troublor (approved):
LGTM — thanks for the thorough turnaround.

All three prior comments substantively addressed:

1. **Validation hoisted** — `HardforkParams::validate` is the right shape: default `Ok(())`, `with_params` panics on error, deploy fn keeps a `debug_assert!` belt-and-braces. Trait-level tests cover the new machinery.
2. **Test duplication removed** — the new unit test in `src/system/intercept.rs` and the doc comments on the existing CALLCODE/DELEGATECALL tests are a clear improvement. Note: the unit test exercises the same observable behavior as the integration tests (real CALLCODE bytecode → `target_address = caller`, not `LIMIT_CONTROL_ADDRESS`), so it doesn't structurally isolate the scheme guard from the existing address check. Accepting as-is — true isolation would require crafting `CallInputs` directly, which is more invasive than this PR warrants.
3. **Comment rewrite** — applied verbatim.

Bonus: the `AGENTS.md` additions ("Override `HardforkParams::validate()` for every new params type" + "Only `CALL` and `STATICCALL` reach interceptor dispatch") capture both new conventions for future contributors.

RealiCZ (approved):
LGTM

PR #278, opened 2026-04-27
fix(system): preserve Oracle storage during bytecode upgrades

### LINKED ISSUE #276: fix: preserve Oracle storage during system contract bytecode upgrades
## Summary

  `transact_deploy_oracle_contract()` currently marks the Oracle account as `created` whenever it updates the Oracle bytecode.

  That is correct for first deployment, but not for in-place upgrades of an existing Oracle account.

  ## Affected area

  - `crates/mega-evm/src/system/oracle.rs`

  ## Details

  During block pre-execution, MegaEVM deploys or upgrades the Oracle system contract at `ORACLE_CONTRACT_ADDRESS`.

  If the account already exists but its code hash differs from the target bytecode, the helper updates the account code and then calls:

  ```rust
  revm_acc.mark_touch();
  revm_acc.mark_created();
```

  The created marker should only be used for a newly created account. For an existing account upgrade, marking it as created can cause commit logic to
  treat the account as newly created and clear existing storage.

  ## Impact

  Medium severity.

  Oracle database-backed storage may be cleared during Oracle bytecode upgrades, for example pre-Rex2 -> Rex2 or Rex2 -> Rex5.

  This matters because Oracle reads can fall back to database storage when the external Oracle environment does not provide a value.

  ## Suggested remediation

  Only call mark_created() when the Oracle account did not previously exist.

  If the Oracle account already exists but needs a bytecode upgrade, update the code and mark it touched, but do not mark it created.

  Add regression tests proving Oracle storage survives:

  1. pre-Rex2 Oracle bytecode -> Rex2 Oracle bytecode upgrade.
  2. Rex2 Oracle bytecode -> Rex5 Oracle bytecode upgrade.

  No spec gate is needed. This is a state-marking bugfix in the pre-block deploy helper, not a protocol behavior change.

### PR DESCRIPTION
**Pull request by TroublorBot**

## Summary

Fixes a bug where in-place bytecode upgrades of the Oracle system contract would silently clear existing storage by incorrectly marking the account as newly created. The fix gates the `mark_created()` call on whether the account previously existed, ensuring upgrades preserve storage. Includes regression tests for pre-Rex2 → Rex2 and Rex2 → Rex5 Oracle bytecode upgrades.

Fixes #276

### DIFF
--- a/crates/mega-evm/src/system/oracle.rs
+++ b/crates/mega-evm/src/system/oracle.rs
@@ -585,4 +585,40 @@ mod tests {
             .expect("Storage read should succeed");
         assert_eq!(read_back, stored_value, "Oracle storage must survive bytecode upgrade");
     }
+
+    /// Covers the `account_existed = true` branch in
+    /// `transact_deploy_high_precision_timestamp_oracle`: when the account already exists with
+    /// wrong code, the deploy updates the code without marking the account as created.
+    #[test]
+    fn test_high_precision_timestamp_oracle_existing_account_not_marked_created() {
+        let wrong_code = alloy_primitives::bytes!("0x6000");
+        let wrong_code_hash = alloy_primitives::keccak256(&wrong_code);
+
+        let mut db = InMemoryDB::default();
+        db.insert_account_info(
+            HIGH_PRECISION_TIMESTAMP_ORACLE_ADDRESS,
+            AccountInfo {
+                balance: revm::primitives::U256::ZERO,
+                nonce: 0,
+                code_hash: wrong_code_hash,
+                code: Some(Bytecode::new_raw(wrong_code)),
+            },
+        );
+
+        let mut state = State::builder().with_database(&mut db).build();
+        let hardforks = MegaHardforkConfig::default().with_all_activated();
+
+        let result = transact_deploy_high_precision_timestamp_oracle(&hardforks, 0, &mut state)
+            .expect("Deployment should succeed")
+            .expect("Should return state");
+
+        let account =
+            result.get(&HIGH_PRECISION_TIMESTAMP_ORACLE_ADDRESS).expect("Account should exist");
+        assert_eq!(account.info.code_hash, HIGH_PRECISION_TIMESTAMP_ORACLE_CODE_HASH);
+        assert!(account.is_touched());
+        assert!(
+            !account.is_created(),
+            "existing account must not be marked as created on code update"
+        );
+    }
 }
--- a/crates/mega-evm/src/system/oracle.rs
+++ b/crates/mega-evm/src/system/oracle.rs
@@ -69,25 +69,31 @@ pub fn transact_deploy_oracle_contract<DB: Database>(
     let acc = db.load_cache_account(ORACLE_CONTRACT_ADDRESS)?;
 
     // If the contract is already deployed with the correct code, return early
-    if let Some(account_info) = acc.account_info() {
+    let existing_info = acc.account_info();
+    if let Some(account_info) = &existing_info {
         if account_info.code_hash == target_code_hash {
             // Although we do not need to update the account, we need to mark it as read
             return Ok(Some(EvmState::from_iter([(
                 ORACLE_CONTRACT_ADDRESS,
-                Account { info: account_info, ..Default::default() },
+                Account { info: account_info.clone(), ..Default::default() },
             )])));
         }
     }
 
     // Update the account info with the contract code
-    let mut acc_info = acc.account_info().unwrap_or_default();
+    let account_existed = existing_info.is_some();
+    let mut acc_info = existing_info.unwrap_or_default();
     acc_info.code_hash = target_code_hash;
     acc_info.code = Some(Bytecode::new_raw(target_code));
 
     // Convert the cache account back into a revm account and mark it as touched.
+    // Only mark it as created when the account did not previously exist; an in-place
+    // bytecode upgrade of an existing account must not clear its storage.
     let mut revm_acc: revm::state::Account = acc_info.into();
     revm_acc.mark_touch();
-    revm_acc.mark_created();
+    if !account_existed {
+        revm_acc.mark_created();
+    }
 
     Ok(Some(EvmState::from_iter([(ORACLE_CONTRACT_ADDRESS, revm_acc)])))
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: crates/mega-evm/src/system/oracle.rs [hunk 1]
```diff
@@ -69,25 +69,31 @@ pub fn transact_deploy_oracle_contract<DB: Database>(
...
 
     // Convert the cache account back into a revm account and mark it as touched.
+    // Only mark it as created when the account did not previously exist; an in-place
+    // bytecode upgrade of an existing account must not clear its storage.
     let mut revm_acc: revm::state::Account = acc_info.into();
     revm_acc.mark_touch();
-    revm_acc.mark_created();
+    if !account_existed {
+        revm_acc.mark_created();
+    }
```
Troublor, line 98: This change is intentionally ungated, but it does change consensus state at the Oracle-upgrade boundary: under the old code, the upgrade marked the account as created and revm's commit wiped its storage; under this fix, the storage is preserved. Any chain that has already crossed an Oracle upgrade (pre-Rex2 → Rex2, or Rex2 → Rex5) under the old code with non-zero DB-backed Oracle storage at that boundary would diverge at the state root when replayed against this version.

The argument in #276 — "this is a state-marking bugfix, not an EVM-semantic change, and gating would leave pre-Rex2 → Rex2 still buggy" — is reasonable, but `AGENTS.md` says "Do NOT modify behavior for existing stable specs," and @Troublor's question on the issue thread doesn't have an on-record acceptance.

Could we confirm one of the following before merge:

1. No live network has crossed an Oracle upgrade boundary with non-empty DB-backed Oracle storage (so there is nothing to diverge from), or
2. Retroactive application is explicitly accepted, or
3. The fix is gated behind a new spec for forward-only chains.

Without one of those on record, this looks like a stable-spec behavior change.

crates/mega-evm/src/system/oracle.rs
```diff
@@ -69,25 +69,31 @@ pub fn transact_deploy_oracle_contract<DB: Database>(
...
     acc_info.code_hash = target_code_hash;
     acc_info.code = Some(Bytecode::new_raw(target_code));
 
     // Convert the cache account back into a revm account and mark it as touched.
+    // Only mark it as created when the account did not previously exist; an in-place
+    // bytecode upgrade of an existing account must not clear its storage.
     let mut revm_acc: revm::state::Account = acc_info.into();
     revm_acc.mark_touch();
-    revm_acc.mark_created();
+    if !account_existed {
```
Troublor: Out-of-scope-but-adjacent: the same `mark_created()`-on-update pattern exists unconditionally in three sibling deploy helpers — `crates/mega-evm/src/system/control.rs:73-81`, `keyless_deploy.rs:72-82`, and `limit_control.rs:52-60`. They each have only one bytecode version today so the bug is latent, but the next time any of them ships a v2 the next author has to remember to gate. Worth applying the same fix in this PR for consistency and to defuse the footgun. (`sequencer_registry.rs` is already safe — it errors out on a different code hash before reaching `mark_created()`.)

Suggested shape (replicated verbatim from this PR's Oracle fix; example for `control.rs`):

```rust
    let account_existed = acc.account_info().is_some();
    let mut acc_info = acc.account_info().unwrap_or_default();
    acc_info.code_hash = ACCESS_CONTROL_CODE_HASH;
    acc_info.code = Some(Bytecode::new_raw(ACCESS_CONTROL_CODE));

    let mut revm_acc: revm::state::Account = acc_info.into();
    revm_acc.mark_touch();
    if !account_existed {
        revm_acc.mark_created();
    }
```

Troublor (changes requested):
Two structural concerns to address before merge:

1. The fix is intentionally ungated but does change consensus state at Oracle-upgrade boundaries — needs an explicit policy decision against `AGENTS.md`'s stable-spec rule (or gating behind a new spec).
2. The same `mark_created()`-on-update pattern is latent in three sibling system-contract deploy helpers (`control.rs`, `keyless_deploy.rs`, `limit_control.rs`). Applying the gating consistently in this PR would defuse the footgun for the next bytecode bump.

The fix itself and the regression tests are clean — these are scope/policy questions, not problems with the implementation.

Troublor:
@TroublorBot Confirmed: mainnet has activated Rex2 through Rex4 and had non-zero DB-backed Oracle storage at the Rex2 activation boundary. The ungated fix would diverge from canonical mainnet state on replay, so the fix must be gated.

Please go with **option 3**. `MegaSpecId::REX5` is the natural gate for the Oracle helper:

- The Rex2 → Rex5 Oracle upgrade has not been crossed on any live chain yet, so gating on REX5 makes the fix forward-only without disturbing the canonical mainnet state at the Rex2 boundary.
- REX5 is the current unstable spec, so adding new behavior there is allowed by `AGENTS.md`.

The four sibling helpers (`high_precision_timestamp_oracle.rs`, `control.rs`, `keyless_deploy.rs`, `limit_control.rs`) each have only one bytecode version, so the `!account_existed` guard never fires today and no spec gate is needed for them — the defensive code can stay as-is.

Please add the gate, regression-test the gated and ungated paths against the Rex2 boundary, and update the PR.

Troublor (commented):
@TroublorBot clarification provided above. Please proceed to gate on REX5.

Troublor:
@TroublorBot clarification provided above. Please proceed to gate on REX5.

Troublor:
@TroublorBot please proceed with the REX5 gate change discussed above.

Troublor (approved):

RealiCZ (approved):
Replayed the Rex2 activation boundary on mainnet to verify the gating decision.

  - Block 7448988 (last pre-Rex2): Oracle slot 0 = 0x00064a0783ee75b6 (non-zero sequencer timestamp).
  - Block 7448989 (Rex2 activation): debug_traceBlockByNumber with prestateTracer shows tx#1's pre-state already has the new v1.1.0 bytecode but slot 0 is absent (= 0). Tx#1 then writes a fresh value.

  So the pre-execution phase of block 7448989 did wipe slot 0 from non-zero to 0 — the buggy mark_created() actually triggered on mainnet, and the cleared-storage state is canonical. An ungated fix would diverge at that block. Rex5 gate is necessary.

  LGTM.

PR #279, opened 2026-04-27
chore(limit): rename BlockLimiter post_execution_check to post_execution_update

### LINKED ISSUE #275: chore: rename BlockLimiter post_execution_check to post_execution_update
## Summary

`BlockLimiter::post_execution_check` does not currently perform fallible post-execution validation. It only updates accumulated block counters and always returns `Ok(())`.

This is not a runtime bug, but the method name makes the call site harder to reason about because it suggests a validation step.

## Affected area

- `crates/mega-evm/src/block/limit.rs`

## Details

`BlockLimiter::post_execution_check` has a fallible return type:

```rust
Result<(), BlockExecutionError>
```

However, the method unconditionally succeeds. Its effective behavior is to update internal counters after transaction execution.

The actual post-execution resource limit enforcement is handled by the pre-execution path for the next transaction. In other words, after a transaction pushes usage over a limit, the limiter blocks subsequent transactions rather than rejecting the already-executed one.

That behavior may be intentional, but the current method name makes it look like this function validates post-execution limits.

## Impact

Info severity. No known runtime correctness issue.

The risk is maintainability: future contributors may assume this method validates post-execution limits, or may skip calling it because they think the checks happen elsewhere. Either mistake could make block-level accounting harder to reason about.

## Suggested remediation

Rename `post_execution_check` to `post_execution_update`.

If there is no intended future error path, also consider changing the return type to `()` so the API reflects that this is an accounting update, not a fallible validation step.

A short doc comment at the method or call site should explain that post-execution usage is accumulated here, while over-limit enforcement happens before admitting the next transaction.

TroublorBot:

## Summary

Rename `BlockLimiter::post_execution_check` to `post_execution_update` to clarify that the method only updates block-level usage counters and does not perform post-execution validation. The `Result` return type is preserved for future extensibility. This resolves issue #275.

Fixes #275

### DIFF
--- a/crates/mega-evm/src/block/limit.rs
+++ b/crates/mega-evm/src/block/limit.rs
@@ -124,8 +124,10 @@
 //!    - During execution, tx-level limits (4-6) are enforced in EVM
 //!    - If exceeded → Transaction fails but continues to step 3
 //!
-//! 3. **Post-execution check** - [`BlockLimiter::post_execution_check`]
-//!    - If block-level violation → Discard outcome, skip, try next transaction
+//! 3. **Post-execution update** - [`BlockLimiter::post_execution_update`]
+//!    - Accumulates resource usage from the executed transaction into block-level counters
+//!    - Does not validate post-execution limits; over-limit enforcement happens before admitting
+//!      the next transaction in [`BlockLimiter::pre_execution_check`]
 //!
 //! 4. **Commit transaction** - [`crate::MegaBlockExecutor::commit_execution_outcome`]
 //!    - Include in block (with success or failed receipt)
@@ -659,8 +661,8 @@ impl BlockLimits {
 ///
 ///     let result = execute_transaction(tx);
 ///
-///     // Post-execution check and update
-///     limiter.post_execution_check(
+///     // Post-execution update
+///     limiter.post_execution_update(
 ///         tx.hash(),
 ///         result.gas_used,
 ///         tx.size(),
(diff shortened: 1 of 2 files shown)


crates/mega-evm/src/block/limit.rs
```diff
@@ -659,8 +661,8 @@ impl BlockLimits {
 ///
 ///     let result = execute_transaction(tx);
 ///
-///     // Post-execution check and update
-///     limiter.post_execution_check(
+///     // Post-execution update
+///     limiter.post_execution_update(
```
Troublor: Pre-existing — this doc example doesn't match the real signature: `post_execution_update` takes a single `&BlockMegaTransactionOutcome<...>` (see L949), not these six raw values (and there's no `tx_hash` parameter on either the wrapper or `post_execution_update_raw`). The `pre_execution_check` call above (L655) is similarly stale — it's missing the `is_deposit: bool` argument.

Since this rename is the natural moment we're touching this block, optional cleanup — replace L665-L672 with:

```rust,ignore
///     limiter.post_execution_update(&outcome)?;
```

Feel free to defer — strictly out of scope of the rename.

Troublor (changes requested):
Rename itself looks good — the new name matches the method's actual accounting-only behavior and aligns with the existing `post_execution_update_raw` helper. One small cleanup inline on the stale doc example.

Troublor:
@TroublorBot please address the latest CI/review state

Troublor (approved):
Feedback addressed — doc example now matches the real signatures, and `is_deposit` is documented on `pre_execution_check`. LGTM.

RealiCZ (approved):
LGTM

PR #280, opened 2026-04-27
fix(limit,gas): harden block and dynamic gas arithmetic against u64 overflow

### LINKED ISSUE #273: fix: harden block and dynamic gas arithmetic against u64 overflow
## Summary

Several resource accounting paths use unchecked `u64` arithmetic. Under current production limits these overflows are not practically reachable, but hardening them would prevent future configuration changes or test/no-limit paths from accidentally turning counter overflow into under-enforcement.

## Affected areas

- `crates/mega-evm/src/block/limit.rs`
- `crates/mega-evm/src/external/gas.rs`

### Block limiter addition checks

`BlockLimiter::pre_execution_check` compares accumulated block usage plus the next transaction's usage against configured limits using unchecked addition.

Examples:

```rust
self.block_gas_used + gas_limit
tx_size + self.block_tx_size_used
da_size + self.block_da_size_used
```

In release builds, overflowing `u64` addition wraps. If a wrapped value becomes small enough, the limit check can incorrectly pass.

Suggested fix: use `saturating_add` for these comparisons. These values are monotonic usage counters; if an addition would overflow, treating the result as maximum usage preserves the intended "block is full / over limit" behavior.

### Block limiter counter updates

`BlockLimiter::post_execution_update_raw` increments several accumulated counters with unchecked `+=`.

Examples include:

```rust
self.block_gas_used += gas_used;
self.block_tx_size_used += tx_size;
self.block_da_size_used += da_size;
self.block_data_used += tx_data;
self.block_kv_updates_used += kv_updates;
self.block_compute_gas_used += compute_gas_used;
self.block_state_growth_used += state_growth_used;
```

If any counter wraps, later checks may see artificially low usage.

Suggested fix: use `saturating_add` for all accumulated block counters. Saturating at `u64::MAX` is appropriate because the pre-execution checks already reject once usage is at or above the configured limit.

### Dynamic gas multiplication

`DynamicGasCost` computes some storage/account gas costs by multiplying base constants by a SALT-derived bucket multiplier. If the multiplier is unexpectedly large, unchecked multiplication can wrap and undercharge the operation.

Suggested fix: use `checked_mul` if the surrounding API can return a meaningful execution/configuration error. Overflow here means the derived gas cost cannot be represented correctly, so aborting is the clearest fail-closed behavior. If returning an error is awkward in this path, `saturating_mul` is still safer than wrapping because it prevents undercharging.

## Impact

Low severity. Current realistic configurations appear far below overflow thresholds, so this is primarily defense-in-depth. The risk is that future limit configurations, `no_limits()` usage, or external environment changes could make overflow reachable and cause resource limits or dynamic gas costs to be under-enforced.

## Suggested remediation

- Replace unchecked additions in `BlockLimiter::pre_execution_check` with `saturating_add`.
- Replace unchecked accumulated counter updates in `post_execution_update_raw` with `saturating_add`.
- Harden `DynamicGasCost` multiplication with `checked_mul` where practical, or `saturating_mul` as a fail-closed fallback.
- Add regression tests for near-`u64::MAX` values where practical.

TroublorBot:

## Summary

Replace unchecked u64 arithmetic on block-level usage counters and dynamic gas multiplications with saturating equivalents to prevent overflow-induced under-enforcement in extreme configurations or test paths. BlockLimiter accumulation and comparison operations now use saturating_add; DynamicGasCost multiplications use saturating_mul and saturating_sub. Includes 9 regression tests covering near-u64::MAX values and zero-multiplier edge cases.

Fixes #273

### DIFF
--- a/crates/mega-evm/src/external/gas.rs
+++ b/crates/mega-evm/src/external/gas.rs
@@ -53,7 +53,7 @@ impl<SaltEnvImpl: SaltEnv> DynamicGasCost<SaltEnvImpl> {
         let multiplier = self.load_bucket_cost_multiplier(bucket_id)?;
 
         let gas = if self.spec.is_enabled(MegaSpecId::REX) {
-            constants::rex::SSTORE_SET_STORAGE_GAS_BASE.saturating_mul(multiplier.saturating_sub(1))
+            constants::rex::SSTORE_SET_STORAGE_GAS_BASE.saturating_mul(multiplier - 1)
         } else {
             constants::mini_rex::SSTORE_SET_STORAGE_GAS.saturating_mul(multiplier)
         };
@@ -69,8 +69,7 @@ impl<SaltEnvImpl: SaltEnv> DynamicGasCost<SaltEnvImpl> {
         let multiplier = self.load_bucket_cost_multiplier(bucket_id)?;
 
         let gas = if self.spec.is_enabled(MegaSpecId::REX) {
-            constants::rex::NEW_ACCOUNT_STORAGE_GAS_BASE
-                .saturating_mul(multiplier.saturating_sub(1))
+            constants::rex::NEW_ACCOUNT_STORAGE_GAS_BASE.saturating_mul(multiplier - 1)
         } else {
             constants::mini_rex::NEW_ACCOUNT_STORAGE_GAS.saturating_mul(multiplier)
         };
@@ -86,8 +85,7 @@ impl<SaltEnvImpl: SaltEnv> DynamicGasCost<SaltEnvImpl> {
         let multiplier = self.load_bucket_cost_multiplier(bucket_id)?;
 
         let gas = if self.spec.is_enabled(MegaSpecId::REX) {
-            constants::rex::CONTRACT_CREATION_STORAGE_GAS_BASE
-                .saturating_mul(multiplier.saturating_sub(1))
+            constants::rex::CONTRACT_CREATION_STORAGE_GAS_BASE.saturating_mul(multiplier - 1)
         } else {
             constants::mini_rex::NEW_ACCOUNT_STORAGE_GAS.saturating_mul(multiplier)
         };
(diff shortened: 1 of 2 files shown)


crates/mega-evm/src/external/gas.rs
```diff
@@ -53,9 +53,9 @@ impl<SaltEnvImpl: SaltEnv> DynamicGasCost<SaltEnvImpl> {
         let multiplier = self.load_bucket_cost_multiplier(bucket_id)?;
 
         let gas = if self.spec.is_enabled(MegaSpecId::REX) {
-            constants::rex::SSTORE_SET_STORAGE_GAS_BASE * (multiplier - 1)
+            constants::rex::SSTORE_SET_STORAGE_GAS_BASE.saturating_mul(multiplier.saturating_sub(1))
```
Troublor: The `saturating_sub(1)` here (and on lines 73, 90) silently turns `multiplier == 0` into a free storage write — fail-open, the opposite direction from the surrounding `saturating_mul` hardening. The MINI_REX paths (lines 58, 75, 92) have the same hazard implicitly: `BASE.saturating_mul(0) == 0`.

`multiplier == 0` requires `bucket_capacity < MIN_BUCKET_SIZE` (256), which `SaltEnv` producers should never return. Let's make that invariant explicit with an `assert!` in `load_bucket_cost_multiplier` so a buggy `SaltEnv` impl panics loudly instead of silently undercharging in production:

```rust
// in load_bucket_cost_multiplier (line ~106), Vacant arm:
let capacity = self.salt_env.get_bucket_capacity(bucket_id)?;
assert!(
    capacity >= MIN_BUCKET_SIZE as u64,
    "SaltEnv returned bucket_capacity={capacity} below MIN_BUCKET_SIZE ({})",
    MIN_BUCKET_SIZE,
);
let multiplier = capacity / MIN_BUCKET_SIZE as u64;
```

If we adopt this, `test_rex_paths_handle_zero_multiplier` should be removed or repurposed since it deliberately violates the invariant.

Troublor (changes requested):
One blocking concern: the REX-path `saturating_sub(1)` makes `multiplier == 0` silently produce free storage writes (fail-open), which is the opposite direction from the rest of the hardening in this PR. Suggestion in the inline comment is to add an `assert!` in `load_bucket_cost_multiplier` so a buggy `SaltEnv` impl panics instead of silently undercharging. Otherwise the change is clean and well-tested.

Troublor:
@TroublorBot please fix the review comments

Troublor (commented):
@TroublorBot labels updated. Please continue.

Troublor (approved):
LGTM

RealiCZ (approved):
LGTM

PR #282, opened 2026-04-28
fix: use live block gas budget for applyPendingChanges() call

RealiCZ:
## Summary

  Fix REX5 pre-block `SequencerRegistry.applyPendingChanges()` running out of gas on role-activation blocks.

  - The pre-block system call inherited revm's upstream-fixed **30M `gas_limit`**, but `applyPendingChanges()` rotates roles via `SSTORE`s whose actual cost scales with **REX dynamic storage gas (SALT bucket capacity)** — 30M is no longer guaranteed to be enough.
  - New `MegaEvm::transact_system_call_with_gas_limit(...)` method lets pre-block helpers pick the gas budget explicitly. The default `transact_system_call` entry point keeps the upstream-fixed 30M, so **EIP-2935 / EIP-4788 and every other system call retain byte-level identical behavior across all stable specs**.
  - Only `transact_apply_pending_changes` opts into the new method, calling with `max(block.gas_limit, SYSTEM_CALL_GAS_LIMIT_FLOOR)`. The 30M floor matches revm's hardcoded `SystemCallTx::new_system_tx_with_caller` literal (revm doesn't export it as a `pub const`, so we mirror it).

  **Note:** Related fixes in `mega-reth` (oracle update tx gas limit chosen too early in `close_window()`) and `test-client` (block-driver progress, oracle tx shape assertions) are tracked outside this repo.

  ## Test plan

  - `cargo fmt --all --check` passed
  - `cargo clippy --workspace --lib --tests --benches --all-features --locked` passed
  - `cargo test -p mega-evm` — all passed 
  - `npx prettier --check 'docs/**/*.md'` passed

### DIFF
--- a/crates/mega-evm/src/evm/mod.rs
+++ b/crates/mega-evm/src/evm/mod.rs
@@ -537,6 +537,43 @@ mod tests {
     }
 
     #[test]
+    fn test_transact_system_call_with_gas_limit_uses_passed_value() {
+        let mut db = MemoryDatabase::default()
+            .account_balance(CALLER, U256::from(1_000_000))
+            .account_code(CALLEE, Bytes::new());
+        let mut evm = MegaEvm::new(configure_context(&mut db));
+
+        let result = evm
+            .transact_system_call_with_gas_limit(CALLER, CALLEE, Bytes::new(), 123_456_789)
+            .unwrap();
+        assert!(result.result.is_success());
+        // The custom gas limit must be applied to the underlying tx.
+        assert_eq!(evm.inner.ctx.tx.base.gas_limit, 123_456_789);
+    }
+
+    #[test]
+    fn test_default_system_call_keeps_upstream_30m_gas_limit() {
+        let mut db = MemoryDatabase::default()
+            .account_balance(CALLER, U256::from(1_000_000))
+            .account_code(CALLEE, Bytes::new());
+        let mut context = MegaContext::new(&mut db, MegaSpecId::REX5);
+        context.modify_chain(|chain| {
+            chain.operator_fee_scalar = Some(U256::ZERO);
+            chain.operator_fee_constant = Some(U256::ZERO);
+        });
+        context.block.gas_limit = 100_000_000;
+        let mut evm = MegaEvm::new(context);
+
+        // The default system-call entry point must NOT be widened by REX5 — only the
+        // explicit `transact_system_call_with_gas_limit` path should pick up the live
+        // block budget. This preserves byte-level behavior of EIP-2935 / EIP-4788
+        // pre-block calls across all specs.
+        SystemCallEvm::transact_system_call_with_caller(&mut evm, CALLER, CALLEE, Bytes::new())
+            .unwrap();
+        assert_eq!(evm.inner.ctx.tx.base.gas_limit, 30_000_000);
+    }
+
+    #[test]
     fn test_mega_evm_exposes_state_wrapper_block_hashes() {
         let mut db = MemoryDatabase::default();
         let mut state = State::builder().with_database(&mut db).build();
--- a/crates/mega-evm/src/evm/mod.rs
+++ b/crates/mega-evm/src/evm/mod.rs
@@ -570,6 +570,9 @@ mod tests {
         // pre-block calls across all specs.
         SystemCallEvm::transact_system_call_with_caller(&mut evm, CALLER, CALLEE, Bytes::new())
             .unwrap();
+        // Literal, not `SYSTEM_CALL_GAS_LIMIT_FLOOR`: this assertion verifies revm's
+        // upstream hardcoded default. If upstream ever drifts from our floor, this
+        // test should fail loudly rather than be auto-aligned by our constant.
         assert_eq!(evm.inner.ctx.tx.base.gas_limit, 30_000_000);
     }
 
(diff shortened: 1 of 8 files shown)


crates/mega-evm/src/system/sequencer_registry.rs
```diff
@@ -943,6 +955,64 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_transact_apply_pending_changes_uses_block_gas_limit() {
```
Troublor, line 959: This test only exercises the **gas-limit propagation path** — with no pending role changes scheduled, `applyPendingChanges()` returns immediately after its no-op check, so the inflated gas budget is never actually consumed. It pins the structural invariant correctly, but it doesn't reproduce the scenario the fix is designed for: an activation block where the role-rotation `SSTORE`s would OOG under the upstream 30M cap once REX dynamic storage gas kicks in.

Could we add a system-level regression test under `crates/mega-evm/tests/rex5/` (sibling to `frame_target_updated_dedup.rs`) that:

1. Schedules pending role changes via `SequencerRegistry`,
2. Sets up a SALT environment with bucket capacities heavy enough that the rotation `SSTORE`s would exceed 30M under REX5 dynamic gas, and
3. Runs the activation block end-to-end, asserting the pre-block call succeeds.

That would be the strongest protection against silent re-regression — e.g. someone "simplifying" `transact_apply_pending_changes` back onto `transact_system_call` would currently pass all existing unit tests but break activation blocks in production. SALT-bucket setup is non-trivial, but `tests/block_executor/sequencer_registry.rs` and the existing rex5 test should provide enough scaffolding to build on.

crates/mega-evm/src/evm/mod.rs
```diff
@@ -536,6 +536,46 @@ mod tests {
...
+        let result = evm
+            .transact_system_call_with_gas_limit(CALLER, CALLEE, Bytes::new(), 123_456_789)
+            .unwrap();
+        assert!(result.result.is_success());
+        // The custom gas limit must be applied to the underlying tx.
+        assert_eq!(evm.inner.ctx.tx.base.gas_limit, 123_456_789);
+    }
+
+    #[test]
+    fn test_default_system_call_keeps_upstream_30m_gas_limit() {
```
Troublor, line 555: Nice that this pins the upstream-30M literal explicitly (rather than via `SYSTEM_CALL_GAS_LIMIT_FLOOR`) — that's the right shape for catching upstream drift.

One scope note though: this is a purely **structural** assertion on `tx.base.gas_limit`, not a behavioral one. The PR description claims 'EIP-2935 / EIP-4788 and every other system call retain byte-level identical behavior across all stable specs,' but no test in this PR actually exercises the EIP-2935 / EIP-4788 paths under REX5 to demonstrate that. Someone who later inverted the dispatch (e.g. routed `transact_system_call` through the new gas-limit method by accident) might still pass this test if they hardcoded `30_000_000` as the override.

Not blocking — the structural invariant is the right unit-level check. Just flagging that the 'byte-identical across all stable specs' claim is, in practice, defended by code review rather than tests.

Troublor (changes requested):
Fix is well-scoped and the policy is correctly gated on the explicit opt-in path, so stable specs keep byte-identical behavior. Two comments to address:

1. The `transact_apply_pending_changes` unit test only exercises gas-limit propagation (no pending role changes), not the OOG scenario the fix targets. A system-level regression test under `crates/mega-evm/tests/rex5/` with a heavy SALT-bucket setup would prevent silent re-regression.
2. The "default system call keeps 30M" test is a structural assertion on `tx.base.gas_limit`, not a behavioral check on EIP-2935 / EIP-4788 paths under REX5. The "byte-identical across all stable specs" claim is defended by code review rather than tests.

vincent-k2026 (approved):
LGTM. Diagnosis is clear, fix is local, and the test design is excellent — particularly:

- `test_default_system_call_keeps_upstream_30m_gas_limit` using the literal `30_000_000` instead of `SYSTEM_CALL_GAS_LIMIT_FLOOR` to deliberately catch upstream drift rather than auto-aligning to it.
- `test_rex5_apply_pending_changes_oogs_under_upstream_30m_cap` as a negative baseline that will start failing if SALT bucket pricing ever drops back under 30M, prompting a re-examination of whether the fix is still load-bearing.
- The EIP-2935 / EIP-4788 isolation tests pinning that REX5 only widens `applyPendingChanges`.

Verified separately that `SequencerRegistry.sol` does not call `gasleft()`, so the `gas_limit` change is not bytecode-observable — state/receipts/logs identical for blocks where 30M was already sufficient. REX5 hasn't activated, so no replay impact either.

Two non-blocking nits:

1. Naming: `SYSTEM_CALL_GAS_LIMIT_FLOOR` lives under `constants::rex5` but the value mirrors revm's universal upstream literal, not a REX5 parameter. The REX5-specific decision is which call sites opt into the widened budget, not the floor value itself. Worth hoisting to top-level `constants` (or a `system_call` submodule named after its provenance) so a future REX spec can reuse it without aliasing.

2. PR body mentions related fixes in mega-reth (oracle `close_window` gas limit) and test-client. Could those PRs be linked here? Want to make sure the merge order is coordinated and devnet validation covers both halves before the REX5 activation block. This PR alone is forward-compatible.

Troublor (approved):

PR #284, opened 2026-04-29
fix(evm): meter CALLCODE new-account storage gas against caller context

### LINKED ISSUE #283: fix: meter CALLCODE new-account storage gas against caller context
## Summary

  `CALLCODE` currently charges MegaETH new-account storage gas using the stack `to` address, which is the code-source address.

  For `CALLCODE`, execution happens in the caller's account context, so new-account metering should be based on the caller/storage context rather than the raw code-source address.

  ## Affected area

  - `crates/mega-evm/src/evm/instructions.rs`

  ## Details

  `storage_gas_ext::wrap_call_with_storage_gas!` currently reads the call target from stack position 1 and uses it for both account inspection and storage-gas charging:

  ```rust
  let to = context.interpreter.stack.inspect::<1>().unwrap().into_address();
  let to_account = context.host.inspect_account_delegated(mega_spec, to)?;
  ...
  context.host.new_account_storage_gas(to)
```

  That is correct for CALL, but not for CALLCODE.

  For CALLCODE, the stack to address only selects the code to execute. The storage/account context remains the caller/current frame address. Charging new-account storage gas against the code-source address can charge when the wrong account is empty, or miss charging when the caller/storage account is the relevant account.

  ## Impact

  MegaETH storage-gas accounting for value-transferring CALLCODE can be wrong.

  This is especially relevant around account-creation metering and EIP-7702 delegated-account emptiness checks.

  ## Suggested remediation

  Split the storage-gas wrapper behavior for CALL and CALLCODE.

  For CALL:

  - keep using the stack to address as the account being value-transferred to.

  For CALLCODE:

  - use the current execution/storage account, i.e. context.interpreter.input.target_address(), for emptiness checks and new_account_storage_gas(...).
  - keep using the stack to address only as the code-source address for the actual CALLCODE instruction.

  Add regression tests showing that a value-transferring CALLCODE meters new-account storage gas based on the caller/storage context, not the code-source address.

  Please verify whether this behavior needs to be gated to the current unstable spec according to AGENTS.md.

### PR DESCRIPTION
**Pull request by TroublorBot**

## Summary

Fixes a bug where `CALLCODE` was charging new-account storage gas based on the code-source address instead of the caller's storage context. For `CALLCODE`, execution happens in the caller's account, so the emptiness check must be against the caller, not the code-source. Rex5 gates the fix by using `interpreter.input.target_address()` for new-account metering, while pre-Rex5 specs preserve frozen behavior for backward compatibility. Includes 4 regression tests covering both the fix and frozen behavior for `CALLCODE` and `CALL`.

Fixes #283

### CODE REVIEW

**Comment by RealiCZ** [comment]
@TroublorBot fix ci

File: crates/mega-evm/src/evm/instructions.rs [hunk 1]
```diff
@@ -1212,6 +1211,67 @@ pub mod storage_gas_ext {
     );
     wrap_call_with_storage_gas!(static_call, "STATICCALL", compute_gas_ext::static_call, false);
 
+    /// `CALLCODE` opcode implementation modified from `revm` with compute gas tracking and
+    /// dynamically-scaled storage gas costs.
+    ///
+    /// This is intentionally not generated via [`wrap_call_with_storage_gas!`] because `CALLCODE`
```
Troublor: The hand-written `call_code` is ~90% identical to the macro expansion — the only real differences are:

1. The storage address: `to` (macro) vs. spec-gated `target_address()`/`to` (here).
2. Stack[2] inspect being unconditional (CALLCODE always has a value operand).

Everything else — stack inspect for `to`, `inspect_account_delegated`, `state_clear_aware_is_empty`, the `is_empty && has_transfer` charge, the `new_account_storage_gas` lookup, the halt/abort branches, and `run_inner_instruction_or_abort!` — is line-for-line the same. If the shared logic ever changes (e.g. stack-inspect ordering, halt-branch additions), we'd have to remember to update both sites.

Suggestion: parameterize the macro with an optional storage-address selector that defaults to `to`. Then CALLCODE becomes a normal `wrap_call_with_storage_gas!` invocation:

```rust
macro_rules! wrap_call_with_storage_gas {
    // Default: storage address is the stack `to`.
    ($fn_name:ident, $opcode_name:expr, $wrapped_fn:path, $has_transfer_logic:expr) => {
        wrap_call_with_storage_gas!(
            $fn_name, $opcode_name, $wrapped_fn, $has_transfer_logic, to
        );
    };
    // Custom storage-address selector: `to`, `mega_spec`, `context` are in scope.
    ($fn_name:ident, $opcode_name:expr, $wrapped_fn:path, $has_transfer_logic:expr, $storage_addr:expr) => {
        #[doc = concat!("`", $opcode_name, "` opcode implementation modified from `revm` with compute gas tracking and dynamically-scaled storage gas costs.")]
        pub fn $fn_name<
            WIRE: InterpreterTypes<Stack: StackInspectTr>,
            H: HostExt + ContextTr + JournalInspectTr + ?Sized,
        >(
            context: InstructionContext<'_, H, WIRE>,
        ) {
            let spec = context.interpreter.runtime_flag.spec_id();
            let Some(to) = context.interpreter.stack.inspect::<1>() else {
                context.interpreter.halt(InstructionResult::StackUnderflow);
                return;
            };
            let to = to.into_address();
            let mega_spec = context.host.spec_id();
            let storage_address: Address = $storage_addr;
            let Ok(storage_account) =
                context.host.inspect_account_delegated(mega_spec, storage_address)
            else {
                context.interpreter.halt(InstructionResult::FatalExternalError);
                return;
            };
            let is_empty = storage_account.state_clear_aware_is_empty(spec);
            let has_transfer = if $has_transfer_logic {
                let Some(value) = context.interpreter.stack.inspect::<2>() else {
                    context.interpreter.halt(InstructionResult::StackUnderflow);
                    return;
                };
                !value.is_zero()
            } else {
                false
            };
            if is_empty && has_transfer {
                let Some(new_account_storage_gas) =
                    context.host.new_account_storage_gas(storage_address)
                else {
                    context.interpreter.halt(InstructionResult::FatalExternalError);
                    return;
                };
                gas!(context.interpreter, new_account_storage_gas);
            }
            run_inner_instruction_or_abort!($wrapped_fn, context);
        }
    };
}

wrap_call_with_storage_gas!(call,          "CALL",         compute_gas_ext::call,          true);
wrap_call_with_storage_gas!(delegate_call, "DELEGATECALL", compute_gas_ext::delegate_call, false);
wrap_call_with_storage_gas!(static_call,   "STATICCALL",   compute_gas_ext::static_call,   false);
wrap_call_with_storage_gas!(
    call_code, "CALLCODE", compute_gas_ext::call_code, true,
    if mega_spec.is_enabled(MegaSpecId::REX5) {
        context.interpreter.input.target_address()
    } else {
        to
    }
);
```

Trade-off: the macro grows one knob (and a touch more macro-magic), but ~60 lines of duplicated logic go away and CALL/CALLCODE/DELEGATECALL/STATICCALL stay in lockstep on the shared parts. The CALLCODE-specific rationale (why the storage address differs) can stay as a comment at the invocation site.

Not blocking — happy to leave as-is if you prefer the explicit form.

Troublor (changes requested):
Refactor suggestion inline: the new `call_code` is largely a copy of the `wrap_call_with_storage_gas!` expansion. Consider parameterizing the macro with an optional storage-address selector so CALL/CALLCODE/DELEGATECALL/STATICCALL stay in lockstep on the shared logic. Non-blocking — fix is correct and well-tested.

vincent-k2026 (approved):

Troublor (approved):

flyq (approved):

PR #285, opened 2026-05-01
refactor: harden sequencer registry bootstrap and admin handoff

RealiCZ:
## Summary

  - **Two-step admin transfer**: `SequencerRegistry.transferAdmin` is now a two-step handoff. `transferAdmin` only sets `_pendingAdmin`; the new admin must call `acceptAdmin` to take effect. Prevents permanent admin lockout from a mistyped, phished, or clipboard-substituted address.
  - **`SequencerRegistryConfig`**: drop `initial_system_address` (the contract now hardcodes `MEGA_SYSTEM_ADDRESS` at genesis); rename remaining fields to `rex5_initial_sequencer` / `rex5_initial_admin`.
  - **Storage layout**: `_pendingAdmin` placed at slot 3, immediately after `_admin`.

### Why slot 3

  Pairing `_pendingAdmin` with `_admin` matches OpenZeppelin `Ownable2Step`, where `_pendingOwner` sits next to `_owner`. 
  
### Why no `__gap`

  OpenZeppelin's `uint256[N] __gap` exists to let a parent contract grow its storage without colliding with a child contract in an upgradeable proxy / multi-layer inheritance setup. `SequencerRegistry` is deployed via raw state patch, has no contract inheritance with state, and is upgraded by spec-gated bytecode replacement that only ever appends new slots — so `__gap` solves an insertion problem we don't have. The real invariant is now stated in the spec doc: future versions may only **append**; never reorder or insert.

## Test plan

  - [x] `cargo fmt --all --check` — clean.
  - [x] `cargo clippy --workspace --lib --examples --tests --benches --all-features --locked` — clean .
  - [x] `prettier --check 'docs/**/*.md'` — clean.
  - [x] `cargo test --workspace` — clean


Troublor (approved):

vincent-k2026 (approved):
LGTM. Two related hardenings landed cleanly together:

- Two-step admin transfer (`transferAdmin` → `acceptAdmin`), matching OZ `Ownable2Step` semantics with the slot placement (slot 3 next to `_admin`, no `__gap`) reasoned through in the PR body.
- `_initialSystemAddress` pinned to `MEGA_SYSTEM_ADDRESS` at genesis, removing a configurability that would silently diverge non-EVM components (payload executor, txpool, replay) from the on-chain state.

Cross-checked layout consistency end-to-end:

- Solidity field declaration order in `SequencerRegistry.sol:14-51` matches the doc table, which matches the Rust slot constants in `system-contracts/src/lib.rs`, which matches the slot-number assertions in both forge and Rust tests.
- Dynamic-array element keys in the new deployed bytecode include `0x0175b7a6…` (= `keccak256(uint256(11))` for `_systemAddressHistory`) and the new `_sequencerHistory[0]` hash — the slot reorder propagates through to the bytecode literals, not just the source.
- `acceptAdmin` access control is sound: `pending == 0` after a cancel cannot be exploited because `msg.sender` is always non-zero in EVM, and `test_acceptAdmin_revertsWhenNoPending` pins this.
- The two `test_setUp_*` invariant tests (one per forge file) are nice fixture-drift guards — anyone reordering the layout in the future without keeping fixtures in sync gets a loud failure rather than silent reinterpretation.

This PR pairs with #282 (gas budget for `applyPendingChanges`); both should land together or close to each other to avoid devnet rebase churn.

Two non-blocking nits:

1. `transferAdmin(address(0))` semantics inverted vs. the previous version (used to revert `ZeroAddress`, now silently cancels pending). This is by design and tested, but the inversion is easy to miss for operators / frontend integrators. Could the NatSpec on `transferAdmin` get a one-line "passes zero == cancel pending; does NOT renounce admin" callout? Pure docs.

2. The "future versions may only **append** new slots" rule is doc-only. A follow-up forge invariant test (or a `forge inspect storage-layout` snapshot in CI) that asserts each named slot's index would catch a future accidental reorder before bytecode lands.

PR #292, opened 2026-05-11
fix: expand memory before slicing initcode in CREATE2 wrapper

RealiCZ:
Fix an out-of-bounds memory access in the `storage_gas_ext::create<IS_CREATE2>` wrapper.

  The wrapper sliced initcode out of interpreter memory via `memory.slice_len(offset, len)` before memory was expanded. The canonical revm CREATE2 path called later does call `resize_memory!`, but by then the slice has already been taken — so a CREATE2 whose `(offset, len)` extended past the current memory size could read past the allocated region in our wrapper.

 ## Test plan

  - New `crates/mega-evm/tests/rex4/create_safety.rs` covering CREATE2 with (offset, len) extending past current memory, including the clean-OOG and successful-expansion paths.
  - `cargo test -p mega-evm` (run before pushing).
  - `cargo clippy --workspace --lib --examples --tests --benches --all-features --locked` and `cargo fmt --all --check`.

### DIFF
--- a/crates/mega-evm/src/evm/instructions.rs
+++ b/crates/mega-evm/src/evm/instructions.rs
@@ -14,8 +14,8 @@ use revm::{
         instructions::{self, control, utility::IntoAddress},
         interpreter::EthInterpreter,
         interpreter_types::{InputsTr, LoopControl, MemoryTr, RuntimeFlag},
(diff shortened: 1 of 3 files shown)


RealiCZ:
Closing this PR — it was opened against the wrong target branch (`main`).

 Reopened here: #293

PR #298, opened 2026-05-20
perf(limit): cache net_usage in FrameLimitTracker

WonderLawrence:
## Summary

- `AdditionalLimit::check_limit` runs on every opcode and fans out to four sub-trackers, each previously walking the full frame stack inside `FrameLimitTracker::net_usage` — making the hot path O(depth) per tracker per opcode even when only `compute_gas` actually mutated state.
- Maintain `Σ(persistent + discardable)` and `Σ refund` incrementally on `FrameLimitTracker` via `cached_total_used` / `cached_total_refund`, so `net_usage()` becomes a single `saturating_sub`.
- All mutations now go through cache-aware helpers (`add_tx_persistent`, `add_frame_persistent`, `add_frame_discardable`, `add_frame_refund`) and the existing `pop_frame`, which on revert subtracts the child's `discardable_usage` and `refund` from the cache (those values vanish rather than being merged into the parent). The four sub-trackers (`compute_gas`, `data_size`, `kv_update`, `state_growth`) are updated to call the helpers instead of touching `FrameLimitEntry` fields directly. The unused `tx_mut` accessor is removed.

## Behavior

- No semantic change. `net_usage()` still uses `saturating_sub`, so the transient `refund > used` state observed in `state_growth` continues to clamp to zero externally.
- All four resource dimensions (compute gas / data size / KV updates / state growth) follow the same cache invariant; revert paths are unwound through `pop_frame` as before.

## Test plan

- [x] `cargo fmt --all --check`
- [x] `cargo clippy --workspace --lib --examples --tests --benches --all-features --locked`
- [ ] Existing `crates/mega-evm/tests/` suites covering frame revert / refund accounting (relied on in CI)
- [x] Benchmark sanity check on `transact` to confirm the hot-path improvement is realized

### DIFF
--- a/crates/mega-evm/src/limit/compute_gas.rs
+++ b/crates/mega-evm/src/limit/compute_gas.rs
@@ -134,10 +134,8 @@ impl ComputeGasTracker {
     ///
     /// Compute gas is always persistent because CPU cycles cannot be undone.
     pub(crate) fn record_gas_used(&mut self, gas: u64) {
-        if let Some(entry) = self.frame_tracker.frame_mut() {
-            entry.persistent_usage += gas;
-        } else {
-            self.frame_tracker.tx_mut().persistent_usage += gas;
+        if !self.frame_tracker.add_frame_persistent(gas) {
+            self.frame_tracker.add_tx_persistent(gas);
         }
     }
 }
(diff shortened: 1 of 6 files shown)


Troublor (approved):
LGTM

RealiCZ (approved):

PR #299, opened 2026-05-22
feat(bench): add attack_replay benchmark

vincent-k2026:
## Summary

Adds `attack_replay`, a hermetic regression benchmark that replays a real MegaETH mainnet attack contract deployment through `MegaEvm`.

The fixture is a self-contained ~64 KB JSON snapshot captured via `debug_traceCall` + `prestateTracer` (`diffMode=false`) at a fixed block:
- **tx**: caller / nonce / gas / value / 17 KB initcode / chain_id
- **prestate** (11 accounts): 4 proxy contracts at `0x4200...`, 1 ERC-20 with code + 10 storage slots, the caller, and 5 supporting storage contracts
- **block env**: number / timestamp / basefee / gas_limit / beneficiary / mix_hash

## Bench arms

Three arms on the same in-memory state:

| arm | typical wall time | engine |
|---|---:|---|
| `attack_replay/equivalence` | ~1.15 ms | `MegaSpecId::EQUIVALENCE` |
| `attack_replay/mini_rex`    | ~35.9 ms | `MegaSpecId::MINI_REX` |
| `attack_replay/pure_revm`   | ~1.10 ms | vanilla `revm` (`Context::mainnet`) baseline |

Both mega-evm specs execute the exact same **205,951 opcodes** and deploy the same 582-byte runtime. The ~30x gap between `EQUIVALENCE` and `MINI_REX` isolates the cost of the multi-dimensional `AdditionalLimit` accounting (quadratic LOG / compute / storage / data / KV buckets) that `MINI_REX` enables.

The `pure_revm` arm is a self-check: it should land near the `EQUIVALENCE` arm, confirming the bench is honest and not short-circuiting.

## Why this bench

Numbers correlate directly with production: the `mini_rex` arm matches the sequencer-monitor's observed ~33 ms inside `api.inspect(...)` for this exact transaction, making the bench a stable, reproducible target for any limit-tracker / hot-path optimization (e.g. caching `net_usage` in `FrameLimitTracker`).

## Sanity checks

Run before criterion warm-up:
- Asserts `ExecutionResult::Success` variant + reports deployed code length + addr + accounts/slots touched.
- Counts opcode steps via a minimal `OpcodeCounter` inspector and asserts `steps >= MIN_EXPECTED_OPCODE_STEPS` (100,000). Any future setup mistake that silently short-circuits tx validation will fail the bench loudly instead of producing artificially fast numbers.

## Test plan
- [x] `cargo bench --bench attack_replay -p mega-evm --no-run` builds clean
- [x] `cargo +nightly fmt --check -p mega-evm` clean
- [x] `cargo clippy --bench attack_replay -p mega-evm` 0 warnings
- [x] `cargo bench --bench attack_replay -p mega-evm -- --quick` runs, all sanity assertions pass, numbers as expected


flyq (approved):

Troublor:
@RealiCZ — non-blocking follow-up suggestion; happy to see this merge as-is.

The hand-rolled fixture parser is the right local call for one bench, but if more replay-style benches land we'll want a unified path. Sketching what the follow-up could look like:

1. **Adopt the EEST state test schema as the canonical fixture format.** [`state-test`](crates/state-test/src/types/test_unit.rs) already has it — `TestUnit { env, pre, transaction, post, out }`, with `AccountInfo` deriving both `Serialize` and `Deserialize` (camelCase + `alloy_serde::quantity`), so the format is already round-trippable.

2. **Extend `mega-evme replay` with `--dump-fixture <FILE>`.** Record every `Database::basic` / `storage` / `code_by_hash` access during the replay to populate `pre`; copy block env into `env`, the on-chain tx into `transaction`, and the actual `ResultAndState` into `post` so the fixture self-checks on read. Workflow becomes:
   ```
   mega-evme replay <tx_hash> --rpc <url> --dump-fixture foo.json
   ```
   One command, no external `cast rpc debug_traceCall + prestateTracer` needed.

3. **Reuse `state-test` types from this bench.** Add `state-test` as a `dev-dependency` on `mega-evm` — Cargo permits the dev-dep cycle (`mega-evm --dev→ state-test --regular→ mega-evm`). The bench drops `parse_u256` / `parse_bytes` / `parse_address` + the bespoke `TxFixture` / `AccountFixture` / `BlockFixture` structs in favor of `use state_test::types::TestUnit`. `mega-evme`'s private `AccountState` (in `bin/mega-evme/src/common/state.rs`) also folds into the same type.

Implementation notes for whoever picks this up:
- `state-test`'s lib target is implicit but real (`src/lib.rs` alongside `[[bin]]`) — no Cargo.toml change on the `state-test` side.
- EEST `TransactionParts` is multi-variant (matrix of `data × gas × value`); for a single-shot bench fixture, use a single-element variant via `TxPartIndices(0, 0, 0)`.
- Worth carving `default-features = false` + a slim `types` feature on `state-test` so consumers don't drag in `walkdir` / `indicatif` / `triehash` / `k256` / `plain_hasher`.
- Keep this *separate* from the existing `--rpc.capture-file` / `ReplayTransport` machinery in `bin/mega-evme/src/common/provider/` — that's a transport-level artifact feeding `AlloyDB`, solving a different problem.

Troublor (approved):

Troublor (approved):

RealiCZ (approved):

PR #300, opened 2026-05-26
ci(benchmark): add baseline-gap table and raise timeout

RealiCZ:
## Summary

  Two changes to the Criterion benchmark workflow (`.github/workflows/benchmark.yml`):

  1. **Baseline-gap table in the `/benchmark` comment.** For every group that contains a `revm_pinned` row, the comment now renders a "× vs `revm_pinned`" table showing how each later row — `revm_latest`, `op_revm_pinned`, `op_revm_latest`, and the mega specs (`equivalence` … `rex5`) — scales relative to vanilla revm at the current HEAD. This answers "how far is mega from vanilla revm on this workload, right now," which is orthogonal to the  existing PR-base → PR-head regression table. Groups without a `revm_pinned` row are skipped, so the section stays empty until the baseline rows exist.

  2. **Job timeout raised 60 → 120 min.** The follow-up benchmark work adds `revm 38` / `op-revm 20` baseline dev-dependencies; the cold compile of both revm trees plus running the bench targets twice (feature + baseline checkout) exceeded 60 min on a fresh branch.

  ## Why this lands first

  `benchmark.yml` is triggered by `issue_comment` (`/benchmark`), `schedule`, and `workflow_dispatch` — none of which run the workflow definition from the PR branch. For these events GitHub Actions always loads the workflow from the default branch. So both the higher timeout and the new table only take effect once this is on `main`. Merging it first lets the follow-up bench-code PR's `/benchmark` run pick up the raised timeout and render the baseline-gap table.

  ## Notes

  - No behavior change to the existing regression check; the new section is inserted above it.
  - No new dependencies and no source/EVM changes — workflow-only.


Troublor (approved):

.github/workflows/benchmark.yml
```diff
@@ -156,11 +160,84 @@ jobs:
...
+            // ----------------------------------------------------------------
+            // Baseline gap section (within HEAD commit):
+            // for every group containing `revm_pinned`, show how each later
+            // row scales relative to it. Answers "how far is mega from
+            // vanilla revm on this workload, right now" — orthogonal to the
+            // PR-base vs PR-head regression table below.
+            // ----------------------------------------------------------------
+            function formatTime(ns) {
+              if (ns >= 1_000_000) return (ns / 1_000_000).toFixed(2) + ' ms';
+              if (ns >= 1_000) return (ns / 1_000).toFixed(2) + ' µs';
```
flyq, line 172: 这里是 µs

.github/workflows/benchmark.yml
flyq: us µs 混用，

flyq (approved):

.github/workflows/benchmark.yml
RealiCZ: good catch. fixed

PR #301, opened 2026-05-28
bench: add vanilla revm/op-revm baselines to bench suite

RealiCZ:
## Summary

Add vanilla `revm` and `op-revm` baseline rows to the criterion bench suite at both the currently-pinned versions and the latest crates.io releases, so each group shows how far the mega specs sit from upstream — at versions we depend on today and at versions we'll likely move to next.

### What's new

- **Four baseline rows per group**: `revm_pinned`, `revm_latest`, `op_revm_pinned`, `op_revm_latest`. Joined by the three mega rows (`equivalence`, `mini_rex`, `rex4`) so each criterion group emits seven comparable rows.
- **Shared adapters** in `benches/common/mod.rs`, pulled into `mega_bench`, `revm_bench`, and `transact` via a plain `mod common;`. Also exposes `SPEC_IDS`, `make_mega_evm`, and `build_mega_tx` so the operator-fee zero-out and the `MegaTransaction` envelope-wrap exist in one place instead of seven.
- **`revm 38` + `op-revm 20`** added as renamed dev-dependencies (`revm_latest` / `op_revm_latest`). `op-revm 20` transitively pulls `revm 38`, so both `_latest` baselines share one revm tree.
- **Criterion HTML reports** opted in (`html_reports`, `plotters`, `cargo_bench_support`) for local violin plots under `target/criterion/report/`.

### Notes

- **All four vanilla baselines are pinned to Cancun** so each row differs only by crate and version, not by hardfork. `revm_pinned` / `revm_latest` pin `SpecId::CANCUN` directly; the op rows pin `OpSpecId::HOLOCENE` (eth Cancun) because `DefaultOp::op()` hard-codes `OpSpecId::BEDROCK` (eth Merge) regardless of the enum default — without the pin the op rows would run at Merge and the op-vs-revm gap would reflect a hardfork difference rather than a version one. Cancun also sidesteps EIP-7825's `tx_gas_limit_cap` against the multi-gigagas `gas_limit` workloads.
- **Not strictly dev-deps-only at the lockfile level**: pulling `revm 38` / `op-revm 20` unifies a few shared leaf crates that production builds against too (`alloy-primitives 1.4.1→1.6.0`, `alloy-eips 1.1→1.8.3`, `sha3 0.10→0.11`, `c-kzg`, `blst`). There's no feature leak and the two evm stacks stay duplicated; a full non-bench test run passes (see Test plan).
- `bench_transfer_multi` reuses one EVM across 1000 txns, so it inlines the baseline setup instead of using the single-shot `transact_call_*` adapters.
- No production source touched — dev-dependencies and benches only.

## Test plan

- [x] `cargo check -p mega-evm --benches`
- [x] `cargo fmt --all --check`
- [x] `cargo clippy -p mega-evm --benches --all-features --locked`
- [x] `cargo test -p mega-evm` — 726 passed, 0 failed (full non-bench suite); rest of the workspace tests pass too
- [x] `cargo bench -p mega-evm --bench transact -- --quick` — 7/7 rows pass per group
- [x] `cargo bench -p mega-evm --bench mega_bench -- --quick`
- [x] `cargo bench -p mega-evm --bench revm_bench -- --quick`


crates/mega-evm/benches/transact.rs
```diff
@@ -1,21 +1,26 @@
...
+use mega_evm::{test_utils::MemoryDatabase, MegaSpecId};
 use revm::{
     context::{result::ResultAndState, tx::TxEnvBuilder},
-    database::{CacheDB, EmptyDB},
     primitives::{keccak256, B256},
     ExecuteEvm,
 };
 
+#[path = "common/baseline_adapters.rs"]
+mod common;
```
Troublor: nit: since `common/baseline_adapters.rs` is the only file under `common/`, renaming it to `common/mod.rs` would let all three bench files drop the `#[path]` attribute — `mod common;` would resolve via the standard sibling-folder lookup. Same change in `mega_bench.rs:34-35` and `revm_bench.rs:59-60`.

Trade-off: requires renaming the file, so probably only worth doing if you're already touching this area. If a second shared file ever joins, you'd switch to a `mod.rs` that re-exports submodules anyway.

```suggestion
mod common;
```

crates/mega-evm/benches/common/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+    group.bench_function(with("revm_latest"), |b| {
+        b.iter(|| transact_call_revm_latest(make_latest_db(), params))
+    });
+    group.bench_function(with("op_revm_pinned"), |b| {
+        b.iter(|| transact_call_op_revm_pinned(make_pinned_db(), params))
+    });
+    group.bench_function(with("op_revm_latest"), |b| {
+        b.iter(|| transact_call_op_revm_latest(make_latest_db(), params))
+    });
+}
```
Troublor, line 293: All three bench files repeat the same pipeline per subject — build the DB, build a tx, transact, assert — written seven times per group because each subject needs its own DB type, EVM constructor, and tx envelope. The current `add_baseline_rows_*` helper plus the parallel mega-row loop in each caller is the closest we can get to deduping under this shape.

Consider a follow-up refactor that hoists this into a portable `Workload` + `Subject` trait:

```rust
pub struct Workload {
    pub accounts: Vec<Account>,   // address → { balance, code, storage }
    pub txs:      Vec<TxSpec>,    // single-tx benches pass vec![tx]; transfer_multi
                                  // passes 1000 and a Subject loops internally
}

pub trait Subject {
    fn name(&self) -> &str;
    fn run(&self, workload: &Workload);  // owns DB build, EVM build, tx loop, assert
}
```

The bench file shrinks to declaring a `Workload` and calling one of `register_all` / `register_baselines` / `register_mega`. Each `Subject` impl absorbs its own quirks — Cancun pin, operator-fee zero-out, `MegaTransaction` envelope wrap, `is_success()` assert.

Two design points worth flagging:

- **Keep `TxSpec` minimal** (today's `CallParams`, ~5 fields). Reusing `revm::TxEnv` directly would tie the abstraction to one revm version — both `_latest` subjects would have to translate it into their own `TxEnv` anyway, since revm 27 and revm 38 are separate crates.
- `Vec<TxSpec>` absorbs `bench_transfer_multi` without a parallel "sequence workload" type — the subject decides build-once-loop.

Not for this PR — #301 is a clean introduction of the baselines and is better merged as-is, with the refactor landing on top.

crates/mega-evm/benches/revm_bench.rs
```diff
@@ -292,34 +405,138 @@ fn bench_transfer_multi(c: &mut Criterion) {
...
                         .caller(CALLER)
                         .call(*target)
                         .value(U256::from(1))
                         .gas_limit(100_000)
                         .build_fill();
-                    let mut mega_tx = MegaTransaction::new(tx);
-                    mega_tx.enveloped_tx = Some(Bytes::new());
-                    let r = evm.transact(mega_tx).expect("transfer should succeed");
+                    let r = evm.transact(build_mega_tx(tx)).expect("transfer should succeed");
                     black_box(&r);
```
Troublor, line 542: Missing the `is_success()` assertion that all four baseline blocks above have (lines 437-442, 461-467, 488-494, 516-522). `expect()` on `transact()` only catches internal errors — a transaction that halts (out-of-gas, a per-tx resource limit being exceeded under one of the mega specs, etc.) returns `Ok(ResultAndState)` with `result.is_success() == false`, and the bench keeps spinning. The mega rows would then report numbers from "1000 transfers where some fraction actually halted," which is silently incomparable to the baseline rows that always assert success — defeating the cross-row comparison this bench exists to enable.

```suggestion
                    let r = evm.transact(build_mega_tx(tx)).expect("transfer should succeed");
                    assert!(
                        r.result.is_success(),
                        "mega transfer should succeed: {:?}",
                        r.result
                    );
                    black_box(&r);
```

crates/mega-evm/Cargo.toml
```diff
@@ -42,11 +42,14 @@ serde = { workspace = true, features = ["derive"] }
 thiserror.workspace = true
 
 [dev-dependencies]
-criterion.workspace = true
+criterion = { workspace = true, features = ["html_reports", "plotters", "cargo_bench_support"] }
 hex.workspace = true
 mega-evm = { path = ".", features = ["test-utils"] }
+op_revm_latest = { package = "op-revm", version = "20.0.0", default-features = false, features = ["dev", "serde", "std"] }
```
Troublor: The two renamed packages use snake_case (`op_revm_latest`, `revm_latest`) while the rest of `[dev-dependencies]` follows the crate's upstream name — kebab-case for `mega-evm` and `revm-inspectors`, snake_case only for `serde_json` because that's literally the crate's name on crates.io. Since these are renamed packages, the local key is whatever we choose — kebab-case matches house style.

Cargo auto-converts dashes to underscores in Rust identifiers, so `use op_revm_latest::...` and `use revm_latest::...` keep working unchanged across all three bench files. Only `Cargo.toml` changes.

```suggestion
op-revm-latest = { package = "op-revm", version = "20.0.0", default-features = false, features = ["dev", "serde", "std"] }
```

Same change on line 52:

```toml
revm-latest = { package = "revm", version = "38.0.0", default-features = false, features = ["dev", "serde", "std"] }
```

Troublor (changes requested):

Troublor (approved):

vincent-k2026 (commented):
Took another pass. The harness itself looks right — setup is out of the timed loop, black_box is in place, and the baselines assert success. Two things on what the rows actually compare:

Also worth noting this isn't strictly dev-deps-only at the lockfile level: pulling revm 38 / op-revm 20 unifies a few shared leaf crates that production builds against too (alloy-primitives 1.4.1→1.6.0, alloy-eips 1.1→1.8.3, sha3 0.10→0.11, c-kzg, blst). No feature leak and the evm stacks stay duplicated, but it's worth a line in the description plus one full non-bench test run.

crates/mega-evm/benches/common/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+        .data(p.data.clone())
+        .build_fill();
+    let r = evm.transact(tx).expect("revm_latest transact");
+    assert!(r.result.is_success(), "revm_latest should succeed: {:?}", r.result);
+    black_box(r);
+}
+
+/// `op-revm` at the latest crates.io release, operator fee = 0.
+///
+/// No spec pin: op-revm defaults to Isthmus, which does not carry EIP-7825's
```
vincent-k2026: `::op()` hard-codes `OpSpecId::BEDROCK` (→ eth `MERGE`) and never looks at the enum default, so both op rows actually run at Merge, not Isthmus. Each group then compares Merge / Cancun / Merge / mega-specs, so the op-vs-revm gap is mostly a hardfork difference rather than a version one. The comment is off twice: op-revm 20's default is Jovian, and `::op()` ignores it either way. Pinning the op rows to a matching fork (e.g. Holocene→Cancun) and dropping the Isthmus note would make the columns comparable.

vincent-k2026 (approved):

PR #304, opened 2026-06-05
fix: land REX5 resource-accounting and metering fixes

RealiCZ:
## Summary

  Lands the **REX5** hardfork (frozen, live on mainnet), bringing the repo up to the deployed state. REX5 closes a set of resource-accounting and
  metering gaps across the EVM, system contracts, and transaction validation. All consensus-visible behavior is gated on the REX5 spec — pre-REX5
  specs execute with identical semantics before and after this change, so replay of historical blocks produces identical state roots and receipts.

  Full specification: [`docs/spec/upgrades/rex5.md`](https://github.com/megaeth-labs/mega-evm/blob/cz/sync/rex5-from-internal/docs/spec/upgrades/rex5.md)

  ## What's included

  **KeylessDeploy**
  - Sandbox execution now propagates its resource consumption (compute gas, data size, KV updates, state growth) back to the parent transaction,
  preventing low-cost state bloat via unmetered sandbox work; the override gas is capped to the parent's remaining budget.
  - Selector-only error ABI (decoupled from upstream revm/op-revm `Display`); empty-runtime-code deployments forward the constructor's logs before
  returning.

  **Precompiles & gas**
  - Compute gas for failed precompile calls is now recorded correctly, and precompile calls are bounded by the remaining compute-gas budget (fail-fast
  instead of overshooting).
  - Fixed KZG cost is recorded on REX5 precompile verification failure.

  **EVM metering & state growth**
  - EIP-7702 metering fixes and state-growth tracking for authority accounts created during pre-execution.
  - SELFDESTRUCT charges new-account cost (data size, KV, state growth) when it creates a beneficiary account.
  - Deposit transactions that materialize an empty caller pay the new-account storage gas and contribute `+1` state growth.
  - Caller-account update deduplication: repeated value-transferring sub-calls / creates from the same frame charge the caller update once.

  **CALL / CREATE**
  - CALLCODE new-account storage gas is charged against the caller's storage context, not the code-source address.
  - Top-level CREATE uses the correct storage-gas address source.
  - CREATE code-deposit compute gas is charged atomically with the deployment commit (no double-counting).
  - CREATE2 with zero-length init code short-circuits after the salt check using the empty-code hash.

  **Block & system boundaries**
  - EIP-2935 / EIP-4788 pre-block system calls use a `max(block_gas_limit, 30M)` gas floor and are fail-closed: a non-successful call rejects the
  block.
  - System-contract interceptor dispatch respects `CALL_STACK_LIMIT`.
  - Oracle hint forwarding requires positive `gas_limit` and meters the hint payload against the data-size budget.

  **Stipend & validation**
  - `STORAGE_CALL_STIPEND` reworked into a separated per-frame allowance that no longer inflates the callee's gas limit (drawn only at storage-gas
  surcharge sites; nothing burned, nothing rescued).
  - System-tx chain-id and nonce validation; final Mega-side gas validation deferred to the correct point.

  ## Backward compatibility

  Every behavioral change is gated on `MegaSpecId::REX5`. REX4 and earlier retain their exact prior semantics — the `rex`, `rex2`, `rex3`, and `rex4`
  test suites pass unchanged, confirming pre-REX5 output (state roots, receipts, gas) is byte-for-byte identical.

  ## Verification

  - `cargo test --workspace` — all pass (1000+ tests, 0 failures), including the REX5 suite (204) and the resource-tracker `net_usage` consistency
  assertion.
  - `cargo clippy --workspace --lib --examples --tests --benches --all-features --locked` — clean.
  - `cargo fmt --all --check` and `cargo sort --check` — clean.
  - `cargo check -p mega-evm --target riscv64imac-unknown-none-elf --no-default-features` — `no_std` / cross-arch clean.


Troublor (approved):

crates/mega-evm/src/system/intercept.rs
```diff
@@ -157,6 +186,65 @@ impl<DB: Database, ExtEnvs: ExternalEnvTypes> SystemContractInterceptor<DB, ExtE
...
+            // whether or not `abi_decode` later succeeds.
+            //
+            // The hint is a TX-scoped side-channel into the off-chain oracle backend
+            // (not consensus state), so we record into the TX intrinsic lane — same as
+            // calldata. On overflow: do NOT forward the hint, and do NOT synthesize a
+            // result. Returning `None` lets the next `before_frame_init` step observe
+            // the freshly-flipped `has_exceeded_limit` and produce the canonical
+            // TX-level `OutOfGas` halt via `create_exceeded_limit_result`. The failure
+            // shape (and rescued-gas refund) matches every other data-size overflow.
+            let within = ctx
```
flyq, line 229: **Oracle hint payload double-counted at top level** — `system/intercept.rs:227` calls `record_oracle_hint_bytes(input_bytes.len())` → `add_tx_persistent` (`data_size.rs:96`), the *same* lane as intrinsic calldata (`data_size.rs:200`). `OracleHintInterceptor` has no `depth == 0` guard (unlike `KeylessDeployInterceptor`), so a tx whose **top-level** `to` is the oracle with `sendHint(...)` calldata is charged for those bytes twice against its data-size budget. Verifiers split: one notes `rex5.md:342` / `oracle.md:162` mandate charging "the full raw calldata byte length ... regardless," making this intended; the other confirms the top-level overlap is real. It's REX5-only (no backward-compat impact), only over-charges the sender (conservative), and affects only the unusual direct-to-oracle tx shape — and that path is untested (all oracle-hint tests call from `depth >= 1`). **Action:** confirm the top-level overlap is intended; if so, add a one-line comment; if not, skip the charge when `depth == 0`.

flyq (approved):

PR #306, opened 2026-06-09
feat(replay): self-validating fixtures + throughput benchmark (Rex5 mainnet, deploy registry)

RealiCZ:
## Summary

Adds a self-validating replay-fixture workflow and a replay-throughput benchmark to `mega-evme`, plus the foundation to make them correct: Rex5 mainnet activation, a registry-driven system-contract deploy refactor, and a unified hardfork schedule.

`mega-evme replay --dump-fixture <FILE>` turns one replayed mainnet transaction into a self-contained EEST state-test fixture (pre-state + tx + env + megaEnv + post roots), written only if the local replay reproduces the on-chain receipt's gas and status. A single committed corpus of such fixtures is used two ways — validated for correctness on every PR, and timed for throughput (base-vs-PR) on demand — making it the real-transaction counterpart to the synthetic Criterion suite.

## Foundation (consensus-sensitive — please review closely)

- **Rex5 mainnet activation** — the mainnet hardfork config had Rex5 as `Never`, so replaying post-Rex5 mainnet blocks ran under Rex4 and produced wrong gas. Set Rex5 to `Timestamp(1780632000)` with the mainnet `SequencerRegistryConfig`. This is the only behavior change: it fixes *when* Rex5 activates on mainnet; no spec's semantics change (all specs stay frozen).
- **Unified hardfork schedule** (`block/chain.rs`) — one source for the mainnet (4326) / testnet (6343) activation timestamps and `hardfork_schedule`.
- **Registry-driven system-contract deploy** (`system/deploy.rs`) — a declarative `SystemContractSpec` + a shared `transact_deploy` replace five near-identical deploy functions; each contract exposes one `*_spec` builder as the single source of its gate / bytecode-version / force-create logic. Behavior-preserving: deploy order and per-spec semantics are byte-identical, and SequencerRegistry stays a separate seeded deploy.

## Feature

- **`mega-evme replay --dump-fixture`** — dump a self-validating EEST `TestUnit`. An on-chain fidelity gate (gas + success status) rejects the dump unless the replay reproduces the receipt, so a wrong spec/config can't slip through. Fixtures are byte-reproducible (ordered `pre` / storage / megaEnv) and reject transaction overrides and `--override.spec`.
- **`state-test` gains two modes** — `--bench` times a fixture's isolated EVM `transact` (min/median/mean + Mgas/s, JSON) and is the single throughput-benchmark surface; `--fill` computes and writes a fixture's `post` in place (the offline analog of the dump's post-fill, for fixtures with no on-chain origin).
- **One corpus, two uses** (`bench/replay/fixtures/`) — five self-contained fixtures (system-contract call, plain call, log-heavy, large DeFi, and the 142M-gas attack deploy from #299): validated per-PR by `replay_corpus.rs` and timed by an ABBA base-vs-PR driver (`run.py`) in a new `replay-bench.yml` workflow.
- **Criterion CI** — add the existing `attack_replay` (#299) bench to the compared targets so its limit-tracker regression target is actually observed.
- **state-test schema/runner** — `pre` and account storage are now `BTreeMap` (byte-stable fixtures); a symmetric `to` serializer so CREATE transactions round-trip (`None` ↔ `""`); added `megaEnv` and `megaGasUsed`/`megaStatus` expectations; the production SALT hasher moved into `mega-evm` behind a `test-utils` feature so fixtures replay with mainnet bucket pricing. `mega-t8n` updated for the storage type.

## Testing

- `mega-evm` / `state-test` / `mega-t8n` suites pass; clippy, fmt, `cargo sort`, prettier, and the `no_std`/riscv check are clean.
- New tests: dump round-trip + tamper detection, per-PR corpus self-validation, end-to-end `--dump-fixture` CLI write+validate, byte-reproducibility, the dump override/spec guards, and `state-test --bench` / `--fill` of a fixture.
- Verified end-to-end against mainnet RPC: replay gas/status match on-chain receipts; dumped fixtures self-validate offline; the attack fixture reproduces the source's exact 141,927,106 gas under Rex5.

## Notes / follow-ups

- The replay benchmark complements (does not replace) the Criterion suite; sub-10µs cases are noisy on shared runners, so read the larger cases and the overall trend.
- More corpus cases can be added incrementally — each is one self-contained JSON plus a manifest line (a mined tx via `--dump-fixture`, a non-on-chain case via `--fill`).


RealiCZ:
@codex re: your P2 "Check receipt logs before accepting fixtures" (fixture.rs) — not adding it. The gate stays gas + status; logs are already pinned by logs_root (recorded in finalize_and_write, re-validated by replay_corpus.rs every PR), and a deterministic replay can't diverge in logs. Closing as intentional.

crates/mega-evm/src/block/chain.rs
```diff
@@ -0,0 +1,113 @@
...
+    MegaHardforkConfig::new()
+        .with(MegaHardfork::MiniRex, ForkCondition::Timestamp(0))
+        .with(MegaHardfork::MiniRex1, ForkCondition::Never)
+        .with(MegaHardfork::MiniRex2, ForkCondition::Never)
+        .with(MegaHardfork::Rex, ForkCondition::Timestamp(1764694618))
+        .with(MegaHardfork::Rex1, ForkCondition::Timestamp(1766147599))
+        .with(MegaHardfork::Rex2, ForkCondition::Timestamp(1770116400))
+        .with(MegaHardfork::Rex3, ForkCondition::Timestamp(1771380000))
+        .with(MegaHardfork::Rex4, ForkCondition::Timestamp(1776400000))
+        .with(MegaHardfork::Rex5, ForkCondition::Never)
```
flyq: **Testnet Rex5 = `Never` is wrong — the live testnet has Rex5 active.** `crates/mega-evm/src/block/chain.rs:55`
The spec docs (`docs/spec/upgrades/overview.md`) publish testnet Rex5 activation at `1780459200` (Jun 3, 2026 — past). A verifier confirmed against live RPC: chain 6343 has the Rex5-introduced SequencerRegistry (`0x6342…0006`) deployed with seeded slots (initial sequencer `0xb8db54eb…`, admin `0x1d9bd232…`). So this PR — whose stated purpose includes fixing exactly this class of wrong-spec replay for mainnet — promotes a wrong testnet schedule into the core crate as "the single source of truth," and adds `test_testnet_does_not_activate_rex5` (chain.rs:99–104) codifying it. Every testnet replay after Jun 3 silently runs under REX4. Fix: `Timestamp(1780459200)` + a testnet `SequencerRegistryConfig` (seeds above) + corrected test — or, if `Never` is somehow intentional, fix the spec docs; the two canonical artifacts must not disagree.

bin/mega-evme/Cargo.toml
```diff
@@ -20,6 +20,7 @@ path = "src/main.rs"
 [dependencies]
 # megaeth
 mega-evm = { workspace = true, features = ["default", "test-utils"] }
+state-test.workspace = true
```
flyq: **The next release will fail: `mega-evme` now depends on an unpublished crate.** `bin/mega-evme/Cargo.toml:23`
`mega-evme` 1.5.1 is live on crates.io; this PR adds `state-test` as a regular dependency (workspace pin `path + version "1.5.1"`), but `state-test` does not exist on crates.io (index 404) and `publish.yml` never publishes it. `cargo publish -p mega-evme --locked` will fail dependency resolution at the next release. Options: publish `state-test` (add to publish.yml before mega-evme — though consider whether you want to claim that generic crate name), extract the fixture types into a publishable crate, or restructure the dependency.

bin/mega-evme/src/replay/fixture.rs
```diff
@@ -0,0 +1,342 @@
...
+    // catch this, because the fixture is validated under the same spec it was
+    // dumped with. Refuse to build a fixture that does not match the chain.
+    //
+    // Logs are not compared against the receipt here: for a deterministic replay
+    // with identical pre-state and transaction the emitted logs cannot differ, and
+    // the logs root is independently pinned by `finalize_and_write` (it re-executes
+    // the isolated fixture and records the resulting logs root into `post`, which
+    // the corpus then validates) — that re-execution, not gas, is the real guarantee.
+    let anchor = &inputs.anchor;
+    if actual_gas != anchor.gas_used {
```
flyq, line 165: Should fix (the fidelity gate has a logs-shaped hole)

**Neither fidelity check compares logs**, and the doc's justification ("gas already implies log fidelity, since LOG opcodes are metered") is incorrect — LOG gas depends on topic count and data length, never content. Two layers:
- On-chain anchor (`bin/mega-evme/src/replay/fixture.rs:126–147`) checks only `gas_used` + `success`. A preceding-tx divergence that changes a value the target SLOADs and re-emits yields identical gas/status with different log payloads — and the wrong fixture passes. The receipt with on-chain logs is already fetched; comparing is nearly free.
- Isolated-vs-replay cross-check (`fixture.rs:183–199`) compares gas/status/output but not logs, and a concrete divergence channel exists: the full replay deducts the op-stack L1 data fee from the sender's balance, the isolated run doesn't (zeroed `enveloped_tx`/L1Block), so BALANCE/SELFBALANCE-derived log data can differ while gas/status/output match — and the divergent logs root gets baked into `post` and self-validates forever.

Fix both: extend `OnchainAnchor` with receipt logs, compare the replay logs root against `executed.logs_root` in `finalize_and_write`, and correct `docs/mega-evme/commands/replay.md:108`.

.github/workflows/replay-bench.yml
```diff
@@ -0,0 +1,220 @@
...
+        with:
+          name: replay-bench-report
+
+      - name: Comment on PR
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const fs = require('fs');
+            if (!fs.existsSync('replay-bench.md')) return;
+            const table = fs.readFileSync('replay-bench.md', 'utf8');
```
flyq: - **CI**: the write-privileged `compare` job posts the bench-job's markdown verbatim as a PR comment (`replay-bench.yml:207`) — `run.py` and manifest strings are PR-controlled, unlike `benchmark.yml` which re-renders parsed numbers. Maintainer-gated trigger keeps this low, but parse the JSON and build the table in the trusted job. Also add `persist-credentials: false` to the bench job's checkout (`replay-bench.yml:108`).

bin/mega-evme/src/replay/fixture.rs
```diff
@@ -0,0 +1,342 @@
...
+        nonce: U256::from(tx.nonce()),
+        secret_key: B256::ZERO,
+        sender: Some(sender),
+        to: tx.to(),
+        value: vec![tx.value()],
+        max_fee_per_gas,
+        max_priority_fee_per_gas: tx.max_priority_fee_per_gas().map(U256::from),
+        initcodes: None,
+        access_lists: vec![tx.access_list().cloned()],
+        authorization_list: None,
```
flyq, line 437: - **EIP-7702 txs silently drop their authorization list** (`fixture.rs:338`); the dump then fails with a generic error — reject type-4 explicitly like deposits, or serialize the list (the schema already supports it).

crates/mega-evm/src/system/deploy.rs
```diff
@@ -0,0 +1,201 @@
...
+            return Ok(EvmState::from_iter([(
+                spec.address,
+                Account { info: account_info.clone(), ..Default::default() },
+            )]));
+        }
+    }
+
+    let account_existed = existing_info.is_some();
+    let mut acc_info = existing_info.unwrap_or_default();
+    acc_info.code_hash = spec.code_hash;
```
flyq, line 114: - **`transact_deploy` trusts caller-supplied `code_hash`** without checking it matches `code` (`system/deploy.rs:99`) — the old per-contract functions made this misuse impossible; a `debug_assert_eq!(keccak256(&code), code_hash)` restores it.

flyq (commented):
- **Stale docs**: `system/AGENTS.md` still describes per-contract hook calls in the executor (line 49) and omits `deploy.rs`; the BLOCKHASH dump rejection is missing from `replay.md`'s incompatibility list (line 112).

flyq (approved):

bin/mega-evme/src/replay/fixture.rs
```diff
@@ -0,0 +1,441 @@
...
```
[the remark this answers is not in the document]
RealiCZ: Type-4 transactions are explicitly rejected upstream in `build_draft` (fixture.rs:133-139) before this constructor runs, so the authorization list is never silently dropped — the dump fails with a clear error pointing at the unsupported tx type. Supporting EIP-7702 dumps (serializing the list into `TransactionParts.authorization_list`) is a future enhancement, not a correctness gap on this path.

crates/mega-state-test/src/runner.rs
```diff
@@ -0,0 +1,1418 @@
...
+    cache.set_state_clear_flag(cfg.spec.into_eth_spec().is_enabled_in(SpecId::SPURIOUS_DRAGON));
+    let mut state =
+        database::State::builder().with_cached_prestate(cache).with_bundle_update().build();
+    inject_block_hashes(&mut state, unit)?;
+
+    let evm_context = MegaContext::default()
+        .with_db(&mut state)
+        .with_cfg(cfg.clone())
+        .with_block(block)
+        .with_external_envs(external_envs_for(unit)?.into());
```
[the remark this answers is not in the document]
RealiCZ: Real divergence channel, but it cannot produce a wrong fixture: a rotated system address only changes execution via deposit-style promotion (sender == system address) or the interceptor caller checks, both of which shift gas/status/logs — and any such divergence aborts the dump at the replay↔isolated cross-check, so no fixture is written. Worst case is a loud dump failure for system-address-sent targets, which are out of scope for `--dump-fixture` the same way deposit transactions are. An explicit early rejection for sender == resolved system address (clearer error than a gas mismatch) is a possible small follow-up.

crates/mega-state-test/src/runner.rs
```diff
@@ -0,0 +1,1418 @@
...
+    let mut megatx = MegaTransaction::new(tx);
+    megatx.enveloped_tx = Some(Bytes::default());
```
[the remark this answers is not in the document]
RealiCZ: Deliberate, documented trade-off — see the comment at this line and the residual note in `docs/mega-evme/commands/replay.md`: carrying real envelope bytes would couple fixtures to the L1Block predeploy state and op-stack L1-fee logic, breaking the self-contained pure-EVM fixture design. The dual gate (chain↔replay, then replay↔isolated) closes every gas/status/output/logs-visible channel; the only open channel is a contract storing a balance-derived value of the fee-shifted sender, and the fixture still self-validates byte-for-byte.

crates/mega-evm/src/system/deploy.rs
```diff
@@ -0,0 +1,211 @@
...
+    revm_acc.mark_touch();
+    // Mark created on a fresh deploy, or when the contract opts to clear storage
+    // on an in-place upgrade. Not marking created preserves existing storage.
+    if !account_existed || spec.force_create_on_upgrade {
+        revm_acc.mark_created();
+    }
+
+    for (slot, value) in &spec.seed {
+        revm_acc.storage.insert(*slot, EvmStorageSlot::new_changed(U256::ZERO, *value, 0));
+    }
```
Troublor, line 129: The `seed` loop runs unconditionally, decoupled from `mark_created`. Today this is safe only by convention — the sole seeder (`SequencerRegistry`) also sets `force_create_on_upgrade = true`, and every flat contract has an empty `seed`. But the coupling isn't enforced: a future spec with a non-empty `seed` and `force_create_on_upgrade = false` would, on an **in-place upgrade of an already-existing account**, insert genesis slots *on top of* preserved storage — and record each with `original_value = U256::ZERO` even if the slot was already non-zero. That's a silent state-corruption footgun in a consensus path.

Since `seed` is genesis storage, it only makes sense when storage is being (re)created — i.e. exactly when `mark_created` fires. Gating the loop on the same condition makes the invariant structural, with zero behavior change for current contracts (registry: condition always true; flat contracts: empty seed):

```suggestion
    if !account_existed || spec.force_create_on_upgrade {
        revm_acc.mark_created();
        // Seed only on (re)creation. Seeding onto an account whose storage is
        // preserved (an in-place upgrade) would mix genesis slots with live
        // storage and record a wrong `original_value` for an already-set slot.
        for (slot, value) in &spec.seed {
            revm_acc.storage.insert(*slot, EvmStorageSlot::new_changed(U256::ZERO, *value, 0));
        }
    }
```

(Optionally pair with a one-line note on `with_seed` that seed applies only on (force-)creation, and/or a `debug_assert!(spec.seed.is_empty() || !account_existed || spec.force_create_on_upgrade)` to surface the misconfig loudly.)

Troublor (approved):

Troublor (approved):

PR #313, opened 2026-06-12
perf(evm): trim per-opcode hot path and pre-execution scans

RealiCZ:
## Summary

Four behavior-preserving performance optimizations on the per-opcode hot path and pre-execution, plus benchmarks and regression tests. No semantics change for any existing spec — results are bit-identical, so no new spec is introduced.

- **Per-opcode limit check, compute-only.** `record_compute_gas` used to fan out to all four resource dimensions on every opcode, but only compute gas can change there. It now checks that single dimension; the other three are checked and latched at their own mutation sites (`on_sstore`, `on_log`, frame-lifecycle hooks, …) and surfaced by the per-opcode short-circuit, so every exceed still halts at the same opcode. The latch protocol is documented in AGENTS.md and guarded by a debug-only assertion in `record_compute_gas`.
- **Lean opcode wrapper.** `wrap_op_compute_gas!` is split into a default and a `@frame` variant; only the call/create family can set `InterpreterAction::NewFrame`, so the ~95 simple opcodes drop the always-missed child-gas-subtraction match.
- **EIP-7702 authority scan O(N²) → O(N log N).** The transaction-local seen-authorities list becomes a `BTreeMap`, removing a node-CPU amplification reachable with ~1200 unique authorities in one type-4 tx.
- **Static-frame SELFDESTRUCT early-exit.** Inside a STATICCALL frame the inner instruction always halts on revm's static check, so the wrapper skips the beneficiary/caller inspections and SALT pricing that the frame revert would discard anyway.

A fifth change (inlining the SSTORE storage-gas charge to drop a redundant warm-path `inspect_storage`) was reverted after review: it made SSTORE the only opcode charging storage gas outside `storage_gas_ext`, breaking the wrapper-layer invariant for a few in-memory map probes. The SSTORE speedups below come from the compute-only check, not the inlining.

## Measurements

CI Criterion, same-runner back-to-back (baseline `origin/main` → PR head). 267 benchmarks: 0 regressions, 1 warning (snailtracer/equivalence, runner noise on the un-instrumented mainnet path), 55 improvements.

- Hot path (rex4 vs vanilla revm): snailtracer −38%, sstore_100 −20%, sstore_sload_100 −19%, mixed_workload −17%, bytecode analysis −14%; gains span every workload category since the optimization targets the shared metering base.
- `eip7702_authlist` sweep (150/400/800/1200 authorities): per-auth cost was growing 0.35 → 0.74 µs (quadratic), now flat ~0.3 µs; −57% total at N=1200, with the improvement growing with N as expected from removing the O(N²) term.
- `staticcall_selfdestruct` (500 calls): the wasted-work gap vs the STOP-child control shrinks 26.1 µs → 8.6 µs (−67%, ~35 ns/event).

## Tests & benches

- New bench `eip7702_authlist`: sweeps authority count with pre-recovered authorizations, so a quadratic term shows as super-linear growth.
- New bench `staticcall_selfdestruct`: SELFDESTRUCT child vs STOP control, isolating the skipped host work.
- New tests: static SELFDESTRUCT leaks no tracked usage; compute gas is recorded even when another dimension already latched; SELFDESTRUCT beneficiary usage latches only at the trailing all-dimension check; SSTORE SALT-pricing failure halts with `FatalExternalError`.
- The debug-only latch-protocol assertion runs across the whole test suite, so any mutation site that forgets to latch trips at the exact opcode.
- Full suite passes (965 tests); clippy and fmt clean.

### DIFF
--- a/crates/mega-evm/src/evm/instructions.rs
+++ b/crates/mega-evm/src/evm/instructions.rs
@@ -121,7 +121,8 @@ use revm::{
 ///   - Block env opcodes (TIMESTAMP, NUMBER, etc.): `volatile_data_ext::*`
 ///   - BALANCE, EXTCODESIZE, EXTCODECOPY, EXTCODEHASH: `volatile_data_ext::*`
 ///   - SLOAD: `compute_gas_ext::sload`
-///   - SSTORE: `additional_limit_ext` → `storage_gas_ext` → `compute_gas_ext`
+///   - SSTORE: `additional_limit_ext` → `compute_gas_ext` (storage-gas charge inlined in
+///     `additional_limit_ext::sstore`)
 ///   - LOG0–LOG4: `additional_limit_ext` → `storage_gas_ext` → `compute_gas_ext`
 ///   - CALL: `forward_gas_ext` → `storage_gas_ext` → `compute_gas_ext`
 ///   - CREATE, CREATE2: `forward_gas_ext` → `storage_gas_ext` → `compute_gas_ext`
@@ -397,7 +398,8 @@ mod mini_rex {
     /// - Most opcodes: `compute_gas_ext::*` (compute gas tracking only)
     /// - Block env opcodes: `volatile_data_ext::*` (gas detention)
     /// - BALANCE, EXTCODESIZE, EXTCODECOPY, EXTCODEHASH: `volatile_data_ext::*` (conditional)
-    /// - SSTORE: `additional_limit_ext` → `storage_gas_ext` → `compute_gas_ext`
+    /// - SSTORE: `additional_limit_ext` → `compute_gas_ext` (storage-gas charge inlined in
+    ///   `additional_limit_ext::sstore`)
     /// - LOG0–LOG4: `additional_limit_ext` → `storage_gas_ext` → `compute_gas_ext`
     /// - CALL: `forward_gas_ext` → `storage_gas_ext` → `compute_gas_ext`
     /// - CREATE, CREATE2: `forward_gas_ext` → `storage_gas_ext` → `compute_gas_ext`
--- a/crates/mega-evm/src/evm/instructions.rs
+++ b/crates/mega-evm/src/evm/instructions.rs
@@ -1567,6 +1567,8 @@ pub mod storage_gas_ext {
         // installs this wrapper only for REX5+, so pre-REX5 specs never reach here.
         if context.interpreter.runtime_flag.is_static() {
             run_inner_instruction_or_abort!(compute_gas_ext::selfdestruct, context);
+            // Defensive: unreachable in practice — a static SELFDESTRUCT always halts
+            // inside the inner instruction, so the macro returns early above.
             return;
         }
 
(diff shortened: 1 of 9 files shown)


crates/mega-evm/src/evm/instructions.rs
```diff
@@ -1590,6 +1557,19 @@ pub mod storage_gas_ext {
...
+        // Inside a static frame, revm's inner SELFDESTRUCT halts on the
+        // static-context check without changing state. Skip the mega host work below
+        // (two account inspections, SALT account-creation pricing, the storage-gas
+        // stipend draw and tracker write) — the frame reverts and discards all of it
+        // anyway — and let the inner instruction produce the identical halt. This is
+        // behavior-neutral: the static halt is exceptional, so the frame's gas and
+        // tracked usage are the same whether or not the host work ran first. The table
+        // installs this wrapper only for REX5+, so pre-REX5 specs never reach here.
+        if context.interpreter.runtime_flag.is_static() {
+            run_inner_instruction_or_abort!(compute_gas_ext::selfdestruct, context);
```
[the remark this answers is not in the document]
RealiCZ: Accurate observation, and we considered it — accepting the new ordering rather than re-adding a peek:

- Both outcomes are error halts: the child frame consumes the same gas, no state is touched, and the parent observes the same failed-call result either way, so state root, receipts, and gas are byte-identical. The only difference is the child frame's internal `InstructionResult`, which is visible to tracers only.
- The path is reachable only by the contract's own bytecode (a SELFDESTRUCT with an empty stack) inside a static frame its author already controls — there is no cross-contract surface, and the chain is single-implementation, so there is no consensus-split axis.
- Restoring the `StackUnderflow` priority would add a stack peek to every static-frame SELFDESTRUCT solely to relabel that attacker-only trace value.

This follows the same triage we apply to other halt-class-on-malformed-stack corners in wrapper prechecks: when gas/state/receipt are unchanged and the path is self-inflicted, the ordering is accepted as specified behavior rather than reordered.

vincent-k2026 (approved):
LGTM.

crates/mega-evm/src/evm/instructions.rs
```diff
@@ -1108,8 +1113,31 @@ pub mod additional_limit_ext {
...
-        run_inner_instruction_or_abort!(storage_gas_ext::sstore, context);
+        // Charge dynamically-scaled storage gas before the instruction executes (only when first
+        // writing a non-zero value to an originally-zero slot). This was previously done in a
+        // separate `storage_gas_ext::sstore` wrapper layer that re-loaded the same slot via a
+        // second `inspect_storage`; the charge is inlined here to reuse the slot already loaded
+        // above, eliminating the redundant per-SSTORE inspection. Behavior is identical: the
+        // first-time-write decision uses the same pre-execution `original`/`present` values, the
+        // charge happens at the same point (before the inner SSTORE runs), and REX5 still drains
+        // the storage stipend allowance first (pre-REX5 returns 0).
+        if original_value.is_zero() && present_value.is_zero() && !new_value.is_zero() {
```
Troublor: This inlining breaks the single-responsibility layering this file documents in the wrapper table above (`instructions.rs:121-127`). Every other storage-gas-charging opcode — LOG, CALL, CREATE, SELFDESTRUCT — routes its storage-gas charge through `storage_gas_ext`. This change makes SSTORE the sole exception: it charges in `additional_limit_ext` and deletes `storage_gas_ext::sstore`, so `additional_limit_ext::sstore` now conflates limit-tracking with storage-gas pricing, and anyone looking for "where is SSTORE storage gas charged?" has to discover the special case (the convention table even had to be edited to call it out).

The payoff is removing a second `inspect_storage` that, by this point, hits the warm path and short-circuits before any DB access — ~5-6 in-memory HashMap probes per SSTORE, with no I/O saved and no dedicated benchmark isolating the effect. That's not worth giving up a clean cross-cutting invariant for.

Suggest dropping the inlining and keeping `storage_gas_ext::sstore` as its own layer. The other four optimizations in this PR are independent and unaffected. If avoiding the redundant inspection is genuinely worth it, thread the already-loaded slot down into `storage_gas_ext::sstore` rather than dissolving the layer.

crates/mega-evm/src/limit/limit.rs
```diff
@@ -357,9 +357,50 @@ impl AdditionalLimit {
...
     pub(crate) fn record_compute_gas(&mut self, compute_gas_used: u64) -> bool {
+        // Record unconditionally, even when another dimension has already latched an exceed:
+        // the compute work was performed, and the recorded total feeds the transaction outcome
+        // and block-level compute accounting. Skipping the record would under-report compute
+        // usage for transactions halted on a non-compute dimension (e.g. intrinsic data size
+        // latched in `before_tx_start` before `validate` records the initial gas).
         self.compute_gas.record_gas_used(compute_gas_used);
+        // If any dimension was already latched as exceeded, surface it immediately (matches the
+        // leading short-circuit in `check_limit`).
+        if self.has_exceeded_limit.exceeded_limit() {
```
Troublor, line 388: The compute-only fast path here is correct only as long as every non-compute mutation site latches its own exceed into `has_exceeded_limit` (the protocol in AGENTS.md). That invariant is load-bearing but currently enforced only by documentation — a future hot-path opcode that records data-size/KV/state-growth without latching (or wrongly uses this wrapper instead of `record_compute_gas_all_dims`) would have its exceed **silently never enforced**, since this method never looks at those three dimensions.

Since all four sub-tracker `check_limit(&self)` are non-mutating, we can guard the invariant for free in debug/test builds without changing release behavior. This trips at the exact opcode that forgot to latch:

```suggestion
        // Debug-only guard for the latch protocol: the compute-only fast path below is
        // sound only if every non-compute mutation site already latched its own exceed.
        // If a non-compute dimension is over limit but not yet latched, some mutation
        // site is missing its check_limit() — catch it here in tests, not in production.
        // (The one pre-inner recorder, SELFDESTRUCT, routes through
        // record_compute_gas_all_dims, not this method, so it never trips this.)
        debug_assert!(
            self.has_exceeded_limit.exceeded_limit()
                || (!self.data_size.check_limit().exceeded_limit()
                    && !self.kv_update.check_limit().exceeded_limit()
                    && !self.state_growth.check_limit().exceeded_limit()),
            "non-compute limit exceeded without latching: a mutation site is missing check_limit()",
        );
        if self.has_exceeded_limit.exceeded_limit() {
```

Troublor (changes requested):

Troublor (approved):

PR #315, opened 2026-06-16
docs: sync spec and mega-evme docs to Rex5

RealiCZ:
## Summary

Brings the `docs/` set in line with the current code at the Rex5 spec, covering both the MegaETH protocol spec and the `mega-evme` CLI reference, and fixes a stale code comment found along the way.
Every numeric value, behavioral rule, interface, and CLI flag was verified against source across `mega-evm`, `mega-reth`, and `salt`; all `mega-evme` example commands were run locally against a release build; and the `docs/**/*.md` tree passes `prettier --check`.
This addresses the documentation audit in #314 (C-001–C-004, R-001, R-002) and several additional discrepancies surfaced during verification.

## mega-evme CLI reference

- Correct the `--spec` default from `Rex4` to `Rex5` and add the `Rex5` row to the spec / available-specs tables (chain-and-spec, run, tx). (#314 C-001, C-002)
- Remove the documented `--rpc.chain-id` flag, which does not exist in the CLI (confirmed at runtime: "unexpected argument").
- Correct the `--rpc` documentation: it has no default endpoint and the `RPC_URL` environment variable is not consulted; `--fork` and `replay` require an `--rpc*` source. Fix the affected examples in state-management/tx/run accordingly.
- Regenerate the stale "Full Help Output" blocks in run.md and tx.md so the option groups match the real `--help`, including a dedicated RPC Options group listing the actual `--rpc.*` cache/retry flags.
- Fix the transaction-types examples: fund the fee-bearing type-0/type-2 transactions (otherwise `LackOfFundForMaxFee`) and use full 32-byte access-list storage keys (abbreviated `0x0` keys are rejected), with a note documenting both requirements.

## Spec

- Move historical pre-latest behavior out of the gas-forwarding and selfdestruct Specification sections into Spec History, per the spec-versioning convention. (#314 R-001, R-002)
- Note that the block-environment and beneficiary gas-detention caps share a single underlying value. (#314 C-004)
- Correct the High-Precision Timestamp interface to the contract's actual method names (`timestamp`, `update(uint256 ts)`, `ORACLE_CONTRACT_ADDRESS`, `ALLOCATION_START`, `ALLOCATION_SIZE`); the previously documented `oracle()`/`baseSlot()`/`maxSlots()` getters do not exist in the deployed bytecode (their selectors revert). Names sourced from the `OracleService` base in `mega-reth`.

## Other

- AGENTS.md: fix the `AdditionalLimit` source path (`limit/mod.rs` → `limit/limit.rs`). (#314 C-003)
- chore(evm): correct a stale comment in `execution.rs` (MegaETH calldata storage gas is a 10x rate: 40 gas/token, not "400"; floor 100 gas/token).

## Verification

- Built `mega-evme` (release) and ran ~35 documented example commands; all succeed (or fail only for the expected reasons the docs now state).
- PoC-verified the timestamp contract by executing its deployed bytecode against each selector.
- Cross-checked every constant against `mega-evm/src/constants.rs`, `mega-reth` (timestamp invariants), and the `salt` repo (`NUM_META_BUCKETS`, `NUM_KV_BUCKETS`, `MIN_BUCKET_SIZE`, bucket-id formula).
- `prettier --check 'docs/**/*.md'` passes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -121,7 +121,7 @@ A transaction can be halted by exceeding either limit.
 
 #### Multidimensional Resource Limits
 
-Beyond the dual gas model, mega-evm enforces **four independent per-transaction resource limits** via `AdditionalLimit` (`limit/mod.rs`):
+Beyond the dual gas model, mega-evm enforces **four independent per-transaction resource limits** via `AdditionalLimit` (`limit/limit.rs`):
 
 - **Compute gas** — Computational opcode cost
 - **Data size** — Calldata + logs + storage writes + code deploy + account updates
--- a/crates/mega-evm/src/evm/execution.rs
+++ b/crates/mega-evm/src/evm/execution.rs
@@ -499,9 +499,9 @@ where
                 .borrow_mut()
                 .record_compute_gas(initial_and_floor_gas.initial_gas);
 
-            // MegaETH MiniRex modification: calldata storage gas costs
-            // - Standard tokens: 400 gas per token (vs 4)
-            // - EIP-7623 floor: 100x increase for transaction data floor cost
+            // MegaETH MiniRex modification: calldata storage gas costs (10x the standard EVM rates)
+            // - Standard tokens: 40 gas per token (vs 4)
+            // - EIP-7623 floor: 100 gas per token (vs 10)
             let tokens_in_calldata = get_tokens_in_calldata(ctx.tx().input(), true);
             let calldata_storage_gas =
                 constants::mini_rex::CALLDATA_STANDARD_TOKEN_STORAGE_GAS * tokens_in_calldata;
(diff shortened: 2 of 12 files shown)


Troublor (approved):

flyq (approved):
## Minor suggestions (none blocking)

1. **Hand-maintained "Full Help Output" blocks (`run.md`, `tx.md`) will re-drift.** This is precisely the failure mode that created this PR (stale `--spec Rex4`, ghost `--rpc.chain-id`). Worth a follow-up to either generate these from `--help` in CI or trim them to a pointer — otherwise the next flag change silently re-stales them. Out of scope here.
2. **`high-precision-timestamp.md` documents a contract that lives in `mega-reth`, not this repo.** The in-repo `doc-correctness` tooling can't guard it, so this page is the most drift-prone in the set. A note/comment pointing at the upstream source file would help the next auditor.

PR #320, opened 2026-06-22
docs: clarify precompile table and limit hot-path comments

mark0-cn:
## Summary

Documentation/comment-only cleanup of three spots in the precompile provider and the EVM hot path.
No behavior change: no spec semantics, gas accounting, or control flow is touched.

## Changes

- `MegaPrecompiles::precompiles()` (`evm/precompiles.rs`): the doc claimed it "returns precompiles with custom gas cost overrides for MINI_REX spec" while the body just returned the inner table under a `// For now` note.
  Rewrote it to state accurately that the precompile gas-cost overrides (fixed-cost KZG point evaluation, Osaka ModExp schedule) are baked into the table by `mini_rex()` / `rex()`, while the per-spec compute-gas accounting and the REX5 forwarded-gas cap are applied separately in `run()`.
- `AdditionalLimit::after_frame_init` (`limit/limit.rs`): added a comment explaining why this site records the child frame's init growth without calling `check_limit()` itself.
  `before_frame_run` latches any exceed before the first metered opcode, so the deferral is sound under the fixed frame lifecycle order (per the Resource-Limit Check Protocol).
  This is the one frame-lifecycle hook that defers its latch, so the reasoning is now written down.
- `compute_gas_ext` module (`evm/instructions.rs`): replaced the `TODO: add more doc` placeholder with a real module doc describing the per-opcode compute-gas wrappers and the `default` / `@frame` / `SELFDESTRUCT` variants.

## Testing

- `cargo fmt --all --check`: clean.
- Doc/comment-only change with no compilation or runtime impact; existing tests are unaffected.

### DIFF
--- a/crates/mega-evm/src/evm/instructions.rs
+++ b/crates/mega-evm/src/evm/instructions.rs
@@ -1650,7 +1650,13 @@ pub mod storage_gas_ext {
     }
 }
 
-/// Compute gas recording implementation. TODO: add more doc
+/// Per-opcode compute-gas recording wrappers.
+///
+/// Each opcode is wrapped so the compute gas it consumes (`gas_before - gas_after`, minus any
+/// gas forwarded to a child frame for the call/create family) is recorded into the compute-gas
+/// dimension of `AdditionalLimit` after the inner instruction runs. The `wrap_op_compute_gas`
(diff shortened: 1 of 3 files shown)


mark0-cn:
@Troublor

PR #325, opened 2026-06-26
ci: add cargo-mutants mutation-testing pipeline + limit-subsystem tests

Troublor:
## Summary

Introduces a mutation-testing pipeline for `mega-evm` (built on `cargo-mutants`) and the first batch of mutation-killing tests for the `limit/` subsystem.

**Tooling & CI**
- `.cargo/mutants.toml` — canonical config: nextest runner, scoped to `mega-evm`, `--all-features` pinned, generated/test-util globs excluded, generous timeouts. cargo-mutants is pinned to `27.1.0` in CI so the generated mutant set is reproducible.
- `scripts/mutation_test.sh` — driver with `diff <base>` (PR-gate, `--in-diff`, merge-base semantics), `full`, and `file <glob>` modes; runs `--no-shuffle -vV` for deterministic, diagnosable runs.
- `scripts/mutation_gate.py` — scorer/gate with a two-tier suppression model (function-scoped excluded before generation, line-scoped filtered after) and a "no new survivors" exit code.
- `.github/workflows/mutation.yml` — diff-scoped PR gate that posts an auto-updating score comment and uploads `mutants.out/`, the scoped diff, and the report as artifacts.
- `mutants/` (README + suppressions), `.gitignore` (ignore regenerable run output), and a `REVIEW.md` note on keeping mutation-test location references current.

**Tests**
- Inline `#[cfg(test)]` unit tests in `src/limit/{compute_gas,data_size,kv_update,state_growth,mod}.rs` that kill survivors: strict-`>` detained-exceed boundary, `LimitKind` u8 round-trip, `within_limit` variant fidelity, originated data-size constant pinning (`+` vs `*`), `record_account_update`/state-growth accounting.
- Integration tests under `tests/mutation/` (the generated mutation-killing system-test target).
- A `tests/rex4/storage_call_stipend.rs` test pinning that `STORAGE_CALL_STIPEND` is excluded from rescued gas on a TX-level detention exceed (no system-granted gas leakage).

## Test plan
- `cargo nextest run -p mega-evm --all-features` — green (1033 tests).
- `cargo fmt --all --check` and `cargo clippy -p mega-evm --lib --tests --all-features --locked` — clean for changed files.
- Full-crate baseline run: 271/271 viable mutants killed (100%), 2 suppressed as equivalent.
- No behavior changes: all source edits are additive `#[cfg(test)]` modules.

---
*This PR was generated by an automated agent.*


vincent-k2026 (commented):
(superseded — see updated English review below)

scripts/mutation_gate.py
```diff
@@ -0,0 +1,145 @@
...
+    return [ln.strip() for ln in path.read_text().splitlines() if ln.strip()]
+
+
+def mutant_body(line: str) -> str:
+    """Strip the leading 'file:line:col: ' locator, leaving the mutation text."""
+    parts = line.split(": ", 1)
+    return parts[1] if len(parts) == 2 else line
+
+
+def cmd_report(args: argparse.Namespace) -> int:
```
vincent-k2026, line 70: If cargo-mutants doesn't produce `missed.txt`/`caught.txt` for any reason (wrong path, early abort, a future version renaming the files), `read_lines` returns `[]` for all of them, so `viable=0`, `score=100%`, `gate_pass=True` — the gate silently passes.

Suggest asserting up front that the run actually completed (e.g. `outcomes.json` exists) and aborting otherwise.

.github/workflows/mutation.yml
```diff
@@ -0,0 +1,102 @@
...
+    runs-on: ubuntu-24.04
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          fetch-depth: 0 # need base history for the diff
+
+      - name: Install Rust
+        uses: actions-rust-lang/setup-rust-toolchain@v1
```
vincent-k2026, line 42: No `Swatinem/rust-cache` step. cargo-mutants does a full compile + test cycle on every run; without a cache the cold start dominates wall-clock. Adding it after this step should roughly halve CI time:

```yaml
- uses: Swatinem/rust-cache@v2
  with:
    workspaces: ". -> target"
```

scripts/mutation_gate.py
```diff
@@ -0,0 +1,145 @@
...
+            "### Survivors needing attention",
+            "",
+            (
+                "Each mutation below changed the code but **no test failed**. Add a test "
+                "that kills it, or — if it is provably equivalent/dead — add a justified "
+                "entry to `mutants/suppressions.toml`."
+            ),
+            "",
+        ]
+        md += [f"- `{m}`" for m in real_survivors]
```
vincent-k2026: A PR introducing dozens of survivors will dump them all here. GitHub caps a single PR comment at 65536 chars, and past a certain length no one reads it. Suggest truncating to the first N (e.g. 20) and pointing at the artifacts for the rest.

mutants/suppressions.toml
```diff
@@ -0,0 +1,246 @@
...
+# EmptyExternalEnv is a unit struct. The real body constructs
+# ExternalEnvs { salt_env: EmptyExternalEnv, oracle_env: EmptyExternalEnv }; the mutant
+# `from(Default::default())` resolves to Default for ExternalEnvs<EmptyExternalEnv>
+# (mod.rs:70-74), yielding identical fields. Value-identical, no observable difference.
+[[suppress]]
+kind = "line"
+file = "crates/mega-evm/src/external/mod.rs"
+mutant = "replace <impl ExternalEnvFactory for EmptyExternalEnv>::external_envs -> ExternalEnvs<Self::EnvTypes> with ExternalEnvs::from(Default::default())"
+justification = "EmptyExternalEnv is a unit struct; real body builds ExternalEnvs{salt_env:EmptyExternalEnv, oracle_env:EmptyExternalEnv}; from(Default::default()) resolves to Default for ExternalEnvs<EmptyExternalEnv> producing identical fields. Value-identical, no observable behavioral difference."
+reviewer = "improve-mutation-score"
```
vincent-k2026: `reviewer` format is inconsistent: earlier entries use `"improve-mutation-score (William Aaron Cheung)"`, from this entry onward it's just `"improve-mutation-score"`. Pick one.

mutants/suppressions.toml
```diff
@@ -0,0 +1,246 @@
...
+# trivial passthrough to the inherent `FrameLimitTracker::tx_limit()`. Every internal
+# consumer (each tracker's own `check_limit()` / `current_call_remaining()`, and
+# `AdditionalLimit`) calls `self.frame_tracker.tx_limit()` directly; the *trait* method
+# on these three trackers is never invoked anywhere in the crate (verified by grepping
+# all `.tx_limit()` call sites — only `ComputeGasTracker`'s trait method is live, since
+# its `check_limit` calls `self.tx_limit()` and composes it with `detained_limit`).
+# Because the function is dead, replacing its body (`-> 0` / `-> 1`) cannot change any
+# observable behavior: the full mega-evm test suite passes with the mutation applied.
+# These are required only to satisfy the `TxRuntimeLimit` trait contract.
+[[suppress]]
```
vincent-k2026, line 65: Consider adding a `category` field to each `[[suppress]]`. In practice there are two distinct kinds — `equivalent` (mathematically/logically equivalent) and `dead` (unreachable / zero call sites) — currently mixed into the free-text `justification`. Tagging them makes it possible to track, e.g., the growth of `dead` suppressions over time (often a signal that code can simply be removed).

vincent-k2026 (commented):
A few small things — (1) and (2) affect gate reliability/speed; the rest are minor.

.github/workflows/mutation.yml
```diff
@@ -0,0 +1,102 @@
+name: Mutation Testing
+
+on:
+  workflow_dispatch:
+  pull_request:
```
RealiCZ, line 12: **Nothing automatically re-checks that these mutation-killing tests still kill their mutants.**

The PR gate only mutates changed lines, and cargo-mutants never mutates `#[cfg(test)]` code — and every source edit here lives in a `#[cfg(test)]` module. So this PR's own gate finds 0 mutants and reports `0/0 PASS`; it never exercises the ~1500 lines of tests added here. That's fine for bootstrap, but going forward, if a later change weakens a test so it stops killing its mutant, nothing catches it: the gate fires only on `src/` changes, and the only full re-verify (271/271) is `workflow_dispatch`-only — there's no `schedule:`.

Suggestion: add a periodic full run. The run step already sends every non-`pull_request` event to `mutation_test.sh full`, so it's basically:

```yaml
on:
  workflow_dispatch:
  pull_request:
  schedule:
    - cron: "0 6 * * 1"   # weekly full re-verify
```

mutants/suppressions.toml
RealiCZ, line 65: **Suppressions can silently rot into blind spots — and nothing flags it.**

Two related gaps:

1. `kind = "function"` entries become `--exclude-re`, so the function is never mutated again. These `tx_limit` trait methods are excluded as "dead today" — but if one later grows real logic, it's permanently unmutated and the gate stays green. For dead code, prefer deleting it or `#[mutants::skip]` over a permanent exclude.
2. If any suppressed function is renamed or moved, its entry just stops matching — a function exclude covers nothing, a line entry matches nothing — with no error. There's no check that a suppression still points at a real mutant, so stale entries accumulate invisibly.

`REVIEW.md` already asks reviewers to keep the `file:line:col` refs in `tests/mutation/` current; suppressions are the higher-stakes "disable a test" decision and deserve at least the same. Worth a small orphan check (every suppression must match a live mutant in a full run) so stale ones surface.

vincent-k2026 (approved):

RealiCZ (approved):

PR #327, opened 2026-06-29
ci(benchmark): paired interleaved A/B + bootstrap CI for drift-robust comparison

RealiCZ:
## Summary

The `/benchmark` harness compares two commits by running the whole feature suite, then the whole baseline suite, and diffing two single point estimates (`--output-format bencher`). The two passes are minutes apart, so the runner's slow drift — thermal throttling, frequency scaling, noisy neighbours — lands entirely on the feature-vs-baseline axis and reads as a ±2–5% phantom delta. That noise floor is large enough to invent regressions on code a PR never touched and to bury a genuine sub-5% optimization in the same band. It is not a sample-count problem (the heaviest benches already run with their tuned sample sizes) — it is drift between the two segments, which more samples within a segment cannot cancel.

This replaces the sequential single-point comparison with the standard methodology for cross-commit microbenchmarking: paired, order-interleaved A/B with a bootstrap confidence interval.

- Both commits' bench binaries are compiled up front (`cargo bench --no-run`); no rebuilds happen during measurement.
- For R rounds, the feature and baseline binaries run back-to-back on the same runner (`A B / B A / …`), each with a short criterion config, so the per-round paired Δ% cancels the common-mode drift the two-segment layout could not.
- Each bench's R paired Δ% are aggregated into a mean Δ% and a seeded 95% bootstrap CI; a change is flagged only when the CI excludes 0 AND |Δ| clears that bench's own A/A noise floor (its round-to-round jitter). Either gate alone is a known false-positive source.

What this buys, concretely: drift — the fixable, dominant term in the old ±2–5% floor — is removed, so untouched rows stop showing phantom regressions, and the residual floor drops to each bench's intrinsic measurement jitter (≈1% on the larger benches, higher on sub-µs ones). Every verdict becomes self-describing: a flag means a change resolved above that bench's noise, a non-flag means the effect is genuinely below it — neither is drift masquerading as signal. A real sub-5% change becomes detectable wherever the bench's intrinsic floor permits, which the old drift-dominated floor effectively prevented.

This is a measurement-methodology change, not one tailored to any single result: no benchmark definition is added or modified (`crates/mega-evm/benches/*` is untouched), only how two commits are compared — so every small-optimization PR benefits. The statistics (median / quantile / seeded bootstrap CI) are ported verbatim from ARO's significance gate. The per-target job matrix and the baseline-gap / regression tables are preserved; the verdict moves from a bare-% threshold to the CI-plus-floor gate, with added CI and A/A-floor columns and a significance legend. Rounds and per-round config (sample size / warm-up / measurement time) are tunable in the workflow's top-level `env:`, and the `bench` job timeout is raised to 180 min for headroom. An `aa_check` `workflow_dispatch` input runs the harness against HEAD versus itself as an on-runner noise-floor sanity check.

## Test plan

The A/A self-check is the correctness oracle: comparing a commit against itself must flag nothing, and the spread it then measures is the pure noise floor. It is wired in as a `workflow_dispatch` input (`aa_check`) and was validated locally by running the harness with the same `transact` bench binary on both sides on an idle machine.

Noise floor (A/A, same binary both sides, R = 20):

| Benchmark | Δ (mean) | 95% CI | verdict |
|-----------|----------|--------|---------|
| empty_transaction/equivalence | -0.2% | [-1.3%, +0.8%] | within noise |
| empty_transaction/mini_rex | +0.0% | [-1.0%, +1.0%] | within noise |
| empty_transaction/op_revm_latest | -0.7% | [-2.0%, +0.3%] | within noise |
| empty_transaction/op_revm_pinned | -0.5% | [-1.4%, +0.4%] | within noise |
| empty_transaction/revm_latest | -0.8% | [-1.9%, +0.2%] | within noise |
| empty_transaction/revm_pinned | +0.3% | [-0.6%, +1.3%] | within noise |
| empty_transaction/rex4 | +0.1% | [-0.5%, +0.7%] | within noise |

Every row reports |mean Δ| < 1% with a 95% CI that contains 0 — 0 significant changes, identically at R = 12 / 15 / 20. The old harness's ±2–5% drift floor collapses to ≈0% ± ~1% once the two sides are paired.

Detection power (same rounds; a known shift injected into the feature side only, so the effect is exact and the noise is unchanged):

- +6% slower → all rows flagged :warning: regression (CIs ≈ [+4%, +7%]).
- −8% faster → all rows flagged :rocket: improvement (CIs ≈ [−9%, −7%]).
- −3% faster → :rocket: on the benches whose A/A floor is < 3%, :eyes: noise-limited on the noisier sub-µs rows — i.e. it resolves a real sub-5% change exactly where the bench's precision allows, instead of drowning it in drift.

Robustness of the gate: under deliberate heavy CPU contention, a CI-only rule produced borderline false :rocket: (|Δ| ≈ 1.8%, CI just past 0); the A/A floor reclassifies those as :eyes: noise-limited (0 significant), which is why the gate is CI-excludes-0 **and** above-floor.

Note: GitHub runs `issue_comment` / `workflow_dispatch` workflows from the default branch, so the new harness takes effect after this merges; the `aa_check` dispatch can then confirm the noise floor on the real CI runner.


.github/scripts/bench_paired.sh
```diff
@@ -0,0 +1,62 @@
...
+for r in $(seq 1 "$rounds"); do
+  rr=$(printf '%02d' "$r")
+  feat_out="${out_dir}/feature__${target}__r${rr}.txt"
+  base_out="${out_dir}/baseline__${target}__r${rr}.txt"
+
+  # Alternate the within-pair order every round so no side is structurally
+  # favoured by always warming the cache / running on a hotter core first.
+  if [ "$baseline_bin" = "" ]; then
+    run_one "$feature_bin" "$feat_out"            # feature-only (no baseline)
+  elif [ $((r % 2)) -eq 1 ]; then
```
flyq, line 62: **1. `.github/scripts/bench_paired.sh:54` (driven by `.github/workflows/benchmark.yml:25`) — odd default `BENCH_ROUNDS=15` defeats the within-pair order-alternation it pairs with, biasing the mean and the A/A oracle.**
The alternation runs feature-first on odd rounds and baseline-first on even rounds. With 15 rounds, feature runs *first* on 8 rounds {1,3,…,15} and *second* on only 7 — an imbalance of one full round. The alternation exists specifically to cancel any first-vs-second execution-slot effect (frequency ramp-up, cold cache, scheduler placement), but with an odd count it cancels only 14/15 of it. Working the algebra on the A/A case (feature binary == baseline binary, slot penalty `s`): per-round Δ is `+s` on feature-first rounds and `−s` on baseline-first rounds, so the mean is `(8·(+s) + 7·(−s))/15 = s/15 ≠ 0`. So even the A/A self-check — the PR's stated correctness oracle, which must read ≈0 — carries a systematic residual whenever a slot effect exists. *Fix: make the default even (14 or 16) so feature-first and feature-second counts match exactly.* (Essentially confirmed — it's exact arithmetic on `seq 1 15` + `r % 2`.)

.github/scripts/bench_compare.py
```diff
@@ -0,0 +1,432 @@
...
+            data[phase].setdefault(name, {})[rnd] = ns
+    return data
+
+
+# ---------------------------------------------------------------------------
+# Aggregation
+# ---------------------------------------------------------------------------
+
+
+def noise_floor(series_list) -> float:
```
flyq, line 183: **2. `.github/scripts/bench_compare.py:175` — `noise_floor` conflates the deterministic slot-order effect with random jitter, so the gate at line 220 can suppress real sub-5% changes on a noisy runner.**
The floor is built from *consecutive same-phase* round deltas (`zip(vals, vals[1:])`, line 187). But because each phase alternates execution slot every round (finding 1's mechanism), the feature series is `[first, second, first, second, …]`, so every consecutive delta straddles a slot boundary and has magnitude ≈ the slot-order amplitude — not the random measurement jitter the docstring claims to estimate. The paired *mean* Δ% correctly cancels that slot effect, but the floor inflates to roughly its amplitude, and the significance gate `abs(mean_delta) > floor` (line 220) then reclassifies a genuine, correctly-measured 3% change as `:eyes:` noise-limited whenever the runner's slot effect exceeds it — directly undermining the "resolve a real sub-5% change" goal. The idle-machine test plan masks this because the slot effect is ~0 there; it would surface under the thermal/contention conditions the PR is built to be robust against. (Plausible — the mechanism is exact; the trigger magnitude depends on the real CI runner's slot-order effect, which the in-band A/A data can't separate from jitter.)

flyq (approved):

.github/scripts/bench_compare.py
```diff
@@ -0,0 +1,443 @@
...
+        return f"{ns / 1_000:.2f} µs"
+    return f"{ns:.0f} ns"
+
+
+def baseline_gap_section(feature) -> str:
+    # feature: {bench_name: {round: ns}} → median ns per row.
+    feature_med = {name: median(list(r.values())) for name, r in feature.items()}
+    groups = {}
+    for full_name, ns in feature_med.items():
+        slash = full_name.rfind("/")
```
vincent-k2026: This splits suffixed Criterion names incorrectly. For `group/revm_pinned/variant`, the baseline row becomes `variant`, so baseline-gap tables are silently omitted for many benchmark groups. Please parse the subject row before the variant.

.github/scripts/bench_compare.py
```diff
@@ -0,0 +1,443 @@
...
+
+
+def load_rounds(rounds_dir: str):
+    """Return {phase: {bench_name: {round: ns}}} for phase in feature/baseline."""
+    data = {"feature": {}, "baseline": {}}
+    for path in sorted(glob.glob(os.path.join(rounds_dir, "*.txt"))):
+        fm = _ROUND_FILE.match(os.path.basename(path))
+        if not fm:
+            continue
+        phase, _target, rnd = fm.group(1), fm.group(2), int(fm.group(3))
```
vincent-k2026: The target is parsed but then discarded. Since all matrix artifacts are merged into one directory, identical Criterion names from different targets can overwrite each other. Please include the target in the aggregation key.

vincent-k2026 (commented):
Two benchmark comparison parser issues.

vincent-k2026 (approved):

PR #330, opened 2026-06-30
bench: add VM-level mega-evm vs revm benchmarks

RealiCZ:
## Summary

Adds VM-level benchmarks that measure mega-evm's overhead over vanilla revm in areas the existing suite did not cover. All changes are additive: the existing subjects and prior bench numbers are unchanged.

## What this adds

- **SALT dynamic-gas isolation** — for SSTORE and CREATE workloads, three rows per spec: `revm_pinned` (floor), mega over `EmptyExternalEnv` (the SALT bucket path short-circuits to a constant), and mega over a crowded `TestExternalEnvs` (real FNV-hash + map-lookup + multiply per touched bucket). The active − baseline gap is the SALT bucket-multiplier path cost, which no prior bench exercised. LOG is intentionally excluded — its storage gas uses a fixed 10× multiplier unrelated to SALT buckets.
- **REX5 comparison rows** — REX5 is added to the shared spec list (so transact / revm_bench / mega_bench all gain a `rex5` vs revm row) and the three block-executor benches now run at both REX4 and REX5.
- **Precompile vs revm baselines** — every precompile group in `comp_cost` gains a vanilla-revm baseline row (revm at PRAGUE, which has BLS12-381 and predates the EIP-7825 tx-gas cap), so each precompile shows a mega-vs-revm gap. The revm row is assert-guarded, so a precompile that does not actually execute fails loudly.
- **Oracle real-data bench** — drives the oracle `get_oracle_storage` path with populated storage. The gap is the oracle SLOAD hit-vs-miss structural cost (hit early-returns; miss falls through to the journal SLOAD), not an isolated lookup — the path forks structurally so the lookup cannot be isolated further.
- **Test-utils enablers** — a test-utils-gated `MegaContext` constructor over a configurable external env (with a unit test), a `TestExternalEnvs` uniform-bucket-capacity setter, a `MegaWithEnv` subject, and the shared env-isolation registration helpers.

## Scope

VM-level only. Comparing against other clients (geth / reth / nethermind / erigon / besu / nimbus) and full-node load testing require the complete MegaETH node and are out of scope here. The comparison page (per-commit trigger, time-series storage, regression alerting) and its MGas/s data source are a follow-up — the per-row gas signal that was initially included here is deferred to that work, where the output format can be pinned to a real consumer.

## Testing

- All affected bench targets smoke-pass (`cargo bench -p mega-evm --bench <target> -- --test`): mega_bench, comp_cost, block_bench, transact.
- Lib unit tests pass, including the new `new_with_ext_envs` constructor coverage test.
- `cargo fmt --all --check` and `cargo clippy --workspace --benches --all-features --locked` are clean.


flyq (approved):
LGTM

vincent-k2026 (approved):

Troublor (approved):

PR #336, opened 2026-07-07
bench: rex5 shared lane + verdict-first comparison comment

RealiCZ:
## Summary
- `benches/common`: add `rex5` to the shared `SPEC_IDS` list. Every bench going through `register_all`/`register_mega` (`sstore_heavy`, `oracle_sload`, `create_deploy`, `call_value_empty_account`, `log_opcodes`, `mixed_workload`, `snailtracer`, `subcall_*`, …) topped out at `rex4`, so the latest spec had no lane exactly where storage/host optimizations land. Benches passing explicit lists (`system_contract`, `eip7702_authlist`, `comp_cost`, …) already carried `rex5`; this brings the shared list in line. Adds ~34 rows to the suite.
- `bench_compare.py`: restructure the `/benchmark` comment so one screen tells the verdict — headline counts first; significant changes worst-first with the top 15 expanded and the rest folded; the baseline-gap tables, the full comparison table, and the methodology prose each fold into their own `<details>`. ms-scale rows render as `143.40 ms` instead of `143403.2 µs`.
- New relink-floor hint: when significant rows land on upstream control lanes (`revm_*`/`op_revm_*`) that do not execute this repo's EVM code, the comment says so explicitly — significance there measures binary-layout/link-order noise between the two builds, and same-magnitude mega-lane findings must be read against that floor (see #335 for the probe quantifying it).

## Test plan
- `cargo check --benches -p mega-evm` green with the rex5 lane.
- `bench_compare.py` exercised against synthetic round files: verdict line, worst-first ordering, top-15/tail fold, control-lane hint, ms formatting, baseline-gap/full-table/methodology folds, and the A/A self-check mode (hint suppressed) all verified on rendered output.
- Comment rendering takes effect on the next `/benchmark` run after merge.

### DIFF
--- a/.github/scripts/bench_compare.py
+++ b/.github/scripts/bench_compare.py
@@ -362,11 +362,9 @@ def ci_str(c: Comparison) -> str:
 
 
 def row_str(c: Comparison, label: str) -> str:
-    b_us = f"{c.baseline_ns / 1000:.1f} µs"
-    f_us = f"{c.feature_ns / 1000:.1f} µs"
     sign = "+" if c.mean_delta > 0 else ""
-    return (f"| {label} | {b_us} | {f_us} | {sign}{c.mean_delta:.1f}% "
-            f"| {ci_str(c)} | ±{c.floor:.1f}% | {icon(c)} |")
+    return (f"| {label} | {format_time(c.baseline_ns)} | {format_time(c.feature_ns)} "
+            f"| {sign}{c.mean_delta:.1f}% | {ci_str(c)} | ±{c.floor:.1f}% | {icon(c)} |")
 
 
 TABLE_HEADER = (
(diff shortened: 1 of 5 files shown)


.github/workflows/benchmark.yml
```diff
@@ -12,8 +12,14 @@ on:
...
 
+# The group is suffixed with "is this a real trigger": issue_comment fires for
+# EVERY new comment (codecov, mutation bots, humans chatting), and with a single
+# group + cancel-in-progress an unrelated comment lands in the same group and
+# cancels a live benchmark mid-flight. Non-trigger comment runs now share their
+# own group and only ever cancel each other; real runs still debounce real
+# re-triggers.
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.issue.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.issue.number || github.ref }}-${{ github.event_name != 'issue_comment' || contains(github.event.comment.body, '/benchmark') }}
```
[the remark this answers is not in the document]
flyq: **1. Concurrency group key doesn't mirror the full trigger gate — an unauthorized comment mentioning `/benchmark` still cancels a live run.** `benchmark.yml:22` keys the group on `contains(github.event.comment.body, '/benchmark')` alone, but the prepare gate (`benchmark.yml:63-70`) additionally requires `issue.pull_request` and MEMBER/COLLABORATOR/OWNER association. A comment that passes the group test but fails the gate lands in the *true* group, cancels the in-flight run via `cancel-in-progress`, then no-ops. The most plausible collision is exactly the protected scenario: a fork-PR author (association `CONTRIBUTOR`/`NONE`) quote-replying the maintainer's `/benchmark` trigger or asking "is /benchmark done?" on the PR being benchmarked kills the 1–3 h run with no replacement. Note `contains` also matches a pasted workflow-file URL (`.../workflows/benchmark.yml` contains `/benchmark`). This is still a strict improvement over pre-PR behavior (any comment cancelled), but mirroring the full gate in the group expression closes it, e.g. suffix with `github.event_name != 'issue_comment' || (github.event.issue.pull_request && contains(github.event.comment.body, '/benchmark') && contains(fromJSON('["MEMBER","COLLABORATOR","OWNER"]'), github.event.comment.author_association))`. Same applies to `replay-bench.yml:25`.

.github/workflows/benchmark.yml
```diff
@@ -266,7 +309,12 @@ jobs:
     needs: [prepare, bench]
     runs-on: ubuntu-24.04
     timeout-minutes: 10
-    if: needs.prepare.outputs.compare == 'true'
+    # `always()` so a failed bench leg still updates the status comment (which
+    # would otherwise sit at "in progress" forever) — partial rounds are
+    # aggregated and the comment gets a warning banner instead.
+    if: >-
+      always() && needs.prepare.result == 'success' &&
+      needs.prepare.outputs.compare == 'true'
```
flyq: **2. `always()` on the compare job resurrects cancelled runs and overwrites the fresh status comment with stale partial results.** Per GitHub semantics, a job whose `if` contains `always()` still runs when the workflow run is cancelled — that's the documented reason to prefer `!cancelled()`. Sequence: `/benchmark` re-trigger cancels run A; run B's prepare posts "in progress" within seconds; run A's compare job (`benchmark.yml:315-317`) then still executes, aggregates whatever partial artifacts exist, and overwrites the shared comment with a `bench: cancelled` partial report that sits there for run B's entire 1–3 h duration. `!cancelled() && needs.prepare.result == 'success' && needs.prepare.outputs.compare == 'true'` achieves the stated goal (failed bench legs still update the comment) without running the compare job of a run that was cancelled precisely because a newer one owns the comment.

.github/workflows/benchmark.yml
```diff
@@ -128,6 +134,43 @@ jobs:
...
+              '',
+              `Comparing \`${short(process.env.BASELINE_SHA)}\` (baseline) → ` +
+                `\`${short(process.env.FEATURE_SHA)}\` (feature). ` +
+                'This comment is updated in place when the run finishes.',
+            ].join('\n');
+            const { owner, repo } = context.repo;
+            const issue_number = context.payload.issue.number;
+            const comments = await github.paginate(github.rest.issues.listComments,
+              { owner, repo, issue_number, per_page: 100 });
+            const mine = comments.filter(c => c.body && c.body.includes(marker)).pop();
```
flyq: **3. Marker matching has no author check — a quote-reply can get its comment hijacked.** GitHub's quote-reply copies raw markdown *including* hidden HTML comments, so a human quoting the report comment reproduces `<!-- criterion-benchmark-report -->` in their own comment. Both lookups (`benchmark.yml:167`, `benchmark.yml:385`) filter on body alone and take `.pop()` (newest match), so the workflow would then edit the human's comment — the write-scoped token can do that. Add an author guard to both filters: `c.user && c.user.login === 'github-actions[bot]' && c.body.includes(marker)`.

.github/scripts/bench_compare.py
```diff
@@ -438,28 +468,70 @@ def render(comparisons, feature_rounds, feature_sha, baseline_sha, repo_url,
...
+            "there measures binary-layout / link-order noise between the two "
+            "builds, and same-magnitude findings on mega lanes should be read "
+            "against that floor.\n\n"
         )
-        body += TABLE_HEADER + "\n".join(rows) + "\n\n</details>\n\n"
-        body += "**Significant changes:**\n\n"
-        body += TABLE_HEADER + "\n".join(row_str(c, label_of[id(c)]) for c in significant) + "\n"
+
+    if significant:
+        by_impact = sorted(significant, key=lambda c: -abs(c.mean_delta))
```
flyq: **4. "Worst first" ranks by |Δ|, so big improvements can push real regressions below the fold.** `bench_compare.py:498` sorts by `-abs(c.mean_delta)`; with more than 15 significant rows, a +6% regression folds into the tail beneath a stack of −20% improvements — the one thing a reviewer must not miss. `key=lambda c: (c.improved, -abs(c.mean_delta))` puts regressions (worst first) ahead of improvements and matches the header's wording.

flyq (approved):

Troublor (approved):

PR #337, opened 2026-07-09
ci(codspeed): adopt instruction-count benchmarks via CodSpeed simulation

RealiCZ:
## Summary

- Adopt CodSpeed simulation mode as the per-PR performance verdict: retired-instruction counts are layout-insensitive, so they can resolve changes that the wall-clock A/B harness buries under the multi-percent relink/layout noise floor measured in #335.
- Swap the `criterion` dev-dependency to `codspeed-criterion-compat` 5.0.1 — a package rename only: bench sources are unchanged, plain `cargo bench` keeps the wall-clock path, and the `html_reports`/`plotters` features are preserved so local HTML reports under `target/criterion` keep working.
- Add `.github/workflows/codspeed.yml`: one job per bench target (7-target matrix) running `cargo codspeed` under valgrind on PRs (path-filtered to changes that can move instruction counts), pushes to main, and manual dispatch. Informational only — not a required check. `cargo-codspeed` installs as a prebuilt binary and the matrix legs share one build cache.
- Revert the disproven binary-layout stabilization: `[profile.bench] codegen-units = 1` (workspace `Cargo.toml`) and `-C llvm-args=-align-all-functions=6` (benchmark workflow). Two probe rounds in #335 showed it ineffective (9 → 16 spurious significant lanes, worst +8.4%), with CGU=1 a suspected net negative. The bench workflow keeps its explicit opt-out of the setup action's default `-D warnings` (measurement builds should not fail on unrelated warnings), and the codspeed workflow opts out for the same reason.
- Document the local instruction-count repro command in `AGENTS.md`.

## Verification

- Wall-clock path unchanged with the compat dep: `cargo bench -p mega-evm --bench mega_bench -- 'rex4/sload_100' --sample-size 10` runs green and `target/criterion/**` — including the HTML report and distribution plots — is written as before, so both `/benchmark` (stdout) and downstream estimate consumers are unaffected.
- `cargo fmt` / `clippy` / `cargo sort` / riscv64 `no_std` check all green (the new dependency is dev-only).
- The exact build/run invocation was validated end-to-end on a fork dry-run, and the workflow has since produced real reports on this PR: CodSpeed detected all 306 benchmarks and posts the per-PR instruction-count comment.

## Notes

- The CodSpeed GitHub App is installed and uploads work (OIDC, no token secret). The install wizard auto-opened #338, which duplicates this onboarding with a slimmer setup; it can be closed in favor of this PR.
- Timing from the first full runs: a cold matrix leg finishes in ~12 minutes and a warm one in 2–7 minutes, so none of the heavy targets (`block_bench`, `attack_replay`, `ctt`) needs to be dropped. The per-target timeout is set to 30 minutes accordingly.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -227,13 +227,6 @@ jobs:
 
       - name: Install Rust
         uses: actions-rust-lang/setup-rust-toolchain@v1
-        with:
-          # 64-byte (2^6) function alignment: cache-line-aligned function
-          # starts remove the largest layout-noise term a relink can inject
-          # into the A/B comparison (complements the bench profile's single
-          # codegen unit; replaces the action's default `-D warnings`, which
-          # measurement builds don't need).
-          rustflags: -C llvm-args=-align-all-functions=6
 
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -227,6 +227,12 @@ jobs:
 
       - name: Install Rust
         uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          # Empty rustflags overrides this action's default `-D warnings`.
+          # Measurement builds should not fail on unrelated warnings; the
+          # previous align-all-functions flag is gone (layout stabilization
+          # was ineffective), but the opt-out of `-D warnings` stays.
+          rustflags: ""
 
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -53,7 +53,7 @@ jobs:
   # cannot make different jobs measure different code (or mislabel the result).
   prepare:
     name: Resolve commits
-    runs-on: ubuntu-24.04
+    runs-on: blacksmith-2vcpu-ubuntu-2404
     timeout-minutes: 10
     # Reacts to the triggering comment; does not execute PR code.
     # `issues: write` is required to react to the (issue) comment — PR
@@ -184,11 +184,11 @@ jobs:
   bench:
     name: Bench ${{ matrix.target }}
     needs: prepare
-    runs-on: ubuntu-24.04
+    runs-on: blacksmith-32vcpu-ubuntu-2404
     # Paired interleaving runs each round's measurement ~1.5x the old single
     # pass, and re-pays per-bench setup once per round; give heavy-setup targets
-    # (block_bench / attack_replay / comp_cost) generous headroom. Well under the
-    # 6h hosted-runner job cap. Lower BENCH_ROUNDS in env to trade back time.
+    # (block_bench / attack_replay / comp_cost) generous headroom. Lower
+    # BENCH_ROUNDS in env to trade back time.
     timeout-minutes: 180
     # This job compiles and runs untrusted PR code — keep it read-only so there
     # is no write-scoped token to exfiltrate. `pull-requests: read` lets
@@ -317,7 +317,7 @@ jobs:
   compare:
     name: Compare and comment on PR
     needs: [prepare, bench]
-    runs-on: ubuntu-24.04
+    runs-on: blacksmith-2vcpu-ubuntu-2404
     timeout-minutes: 10
     # `!cancelled()` (NOT `always()`): a failed bench leg still updates the
     # status comment (partial rounds aggregated under a warning banner), but a
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -21,7 +21,10 @@ permissions:
 
 concurrency:
   group: codspeed-${{ github.ref }}
-  cancel-in-progress: true
+  # PR runs coalesce (only the newest push matters), but every main push is a
+  # baseline datapoint — cancelling one mid-flight leaves a gap CodSpeed must
+  # bridge from an older ancestor. Queue main runs instead of cancelling.
+  cancel-in-progress: ${{ github.ref_name != 'main' }}
 
 jobs:
   codspeed:
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -38,10 +38,20 @@ jobs:
           submodules: recursive
       - name: Install Rust
         uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          # The explicit rust-cache step below owns caching (matrix-shared key).
+          cache: false
+      - name: Cache cargo
+        uses: Swatinem/rust-cache@v2
+        with:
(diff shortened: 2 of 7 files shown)


.github/workflows/benchmark.yml
```diff
@@ -227,13 +227,6 @@ jobs:
 
       - name: Install Rust
         uses: actions-rust-lang/setup-rust-toolchain@v1
```
Troublor, line 229: Removing the `with:` block reverts the `-align-all-functions=6` flag **and** brings back this action's implicit `-D warnings` default on measurement builds. The old block's inline comment explicitly opted out of `-D warnings` for measurement builds ("replaces the action's default `-D warnings`, which measurement builds don't need").

Is that side-effect intentional as a tightening, or should we preserve the previous behavior? If the latter:

```suggestion
      - name: Install Rust
        uses: actions-rust-lang/setup-rust-toolchain@v1
        with:
          rustflags: ""
```

Otherwise, a one-line comment here noting that `-D warnings` is now back on the bench workflow would help the next reader.

.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,51 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - name: Install Rust
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+      - name: Install cargo-codspeed
+        run: cargo install cargo-codspeed --locked --version 5.0.1
```
Troublor: Two CI-cost cleanups worth stacking here before the first full run — both matter more once the matrix is 7 targets × every PR:

1. **Prebuilt `cargo-codspeed`.** `cargo install --locked --version 5.0.1` compiles the CLI from source on every matrix job. `taiki-e/install-action` fetches a prebuilt binary and cuts several minutes per job:

   ```suggestion
      - name: Install cargo-codspeed
        uses: taiki-e/install-action@v2
        with:
          tool: cargo-codspeed@5.0.1
   ```

2. **Cargo cache.** No `Swatinem/rust-cache` step — each of the 7 jobs cold-builds the whole workspace under valgrind instrumentation. Given the 90-min per-target timeout you called out in the PR body, one bad target lands close to the ceiling on a cold build. Adding

   ```yaml
      - name: Cache cargo
        uses: Swatinem/rust-cache@v2
        with:
          shared-key: codspeed
   ```

   right after the Rust install would materially reduce that.

crates/mega-evm/Cargo.toml
```diff
@@ -43,7 +43,11 @@ thiserror.workspace = true
 tracing.workspace = true
 
 [dev-dependencies]
-criterion = { workspace = true, features = ["html_reports", "plotters", "cargo_bench_support"] }
+# CodSpeed drop-in: `codspeed-criterion-compat` re-exports criterion's API, so
+# `use criterion::...` in the benches is unchanged. Under a normal `cargo bench`
+# it behaves exactly like criterion; under `cargo codspeed` it runs each bench
+# once under valgrind and reports layout-insensitive instruction counts.
+criterion = { package = "codspeed-criterion-compat", version = "5.0.1", default-features = false, features = ["cargo_bench_support"] }
```
Troublor: Dropping `html_reports` + `plotters` removes `target/criterion/report/**` from local `cargo bench` runs. `AGENTS.md` still tells contributors 'Always run benchmarks locally before committing' as the pre-commit smoke test — the HTML report with distribution plots is a real ergonomic loss for that workflow.

Could we update `AGENTS.md` to spell this out beyond the one-line comment already added under the Benchmarks block? Specifically:

- Under the 'Always run benchmarks locally before committing' bullet, note that local wall-clock is now stdout-only (no HTML), and point contributors at the CodSpeed PR report for instruction-count deltas.
- Optionally: mention the escape hatch (temporarily re-enabling `html_reports` locally) for cases where a contributor wants distribution plots while investigating a regression.

Troublor (changes requested):

.github/workflows/codspeed.yml
RealiCZ: Both landed in f6d189c, with two notes from digging into the details:

1. `taiki-e/install-action` has no native manifest for `cargo-codspeed`, so it goes through its cargo-binstall fallback — CodSpeedHQ/codspeed-rust publishes prebuilt musl binaries for v5.0.1, and the runs on this branch confirm the install now takes seconds instead of minutes.
2. On the cache: `setup-rust-toolchain` already embeds Swatinem/rust-cache by default (that's why warm legs were already 1.6–7 min), but with per-job keys — 7 near-identical dependency caches. f6d189c disables the built-in cache and adds an explicit `Swatinem/rust-cache@v2` with `shared-key: codspeed` so the legs share one; per-leg bench binaries are workspace artifacts rust-cache never caches anyway, so sharing loses nothing.

Also lowered the per-target timeout 90 → 30 min in 4ed5a1a based on measured runs (cold leg ~12 min, warm 2–7 min).

crates/mega-evm/Cargo.toml
RealiCZ: Went one step further than documenting the loss: `codspeed-criterion-compat` does expose `html_reports`/`plotters` as forwarded features — dropping them was over-caution on my side, not a constraint of the compat crate. c64a046 restores them on the dep, so local `cargo bench` writes `target/criterion/**` (HTML report and distribution plots included) exactly as on main; verified by wiping `target/criterion` and re-running a bench. The AGENTS.md caveat and escape hatch are dropped accordingly, keeping one line that points at the CodSpeed PR report for instruction-count deltas.

Troublor (approved):

.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,65 @@
...
+      - ".github/workflows/codspeed.yml"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  id-token: write   # OIDC auth — public repo, no token secret needed
+
+concurrency:
+  group: codspeed-${{ github.ref }}
+  cancel-in-progress: true
```
flyq: **`cancel-in-progress: true` applies to `main` pushes too** (`.github/workflows/codspeed.yml:22-24`). Rapid back-to-back merges will cancel baseline runs, leaving gaps CodSpeed has to bridge by falling back to older ancestors. Consider `cancel-in-progress: ${{ github.ref_name != 'main' }}` so PR runs still coalesce but every main commit keeps its baseline.

flyq (commented):
**The Blacksmith swap in the two wall-clock workflows is unexercised.** Both are comment-triggered (`/benchmark`), so this PR's green checks only prove the 32-vcpu runner works for the codspeed job. The `blacksmith-2vcpu-ubuntu-2404` tag and the 180-minute bench legs haven't run on Blacksmith at all. Suggest triggering one `/benchmark` smoke run from this branch before merge — also to re-establish the noise floor on the new hardware, since #335's calibration no longer applies.

flyq (approved):
LGTM

PR #340, opened 2026-07-16
docs: reference system-contract bytecode via versioned artifacts

RealiCZ:
## Summary

- Replace the inline deployed-bytecode hex blobs on the system-contract spec pages (Oracle v1.0.0/v1.1.0/v2.0.0, KeylessDeploy v1.0.0, MegaAccessControl v1.0.0, MegaLimitControl v1.0.0, SequencerRegistry v1.0.0) with references to the canonical versioned artifacts under `crates/system-contracts/artifacts/`.
- Each `Deployed bytecode:` entry now shows a truncated prefix, a link to the artifact JSON (`deployedBytecode` field), and a `cast keccak`/`jq` one-liner that re-derives the code hash; the code hash stated on each page remains the normative identity of the bytes.
- This removes ~11 KB of hex that duplicated data already hash-verified at build time by `crates/system-contracts/build.rs`, following the OP Stack specs precedent: the code hash is the normative anchor, full bytes live outside the prose page.
- Docs only; no behavior change.

## Verification

- Every replaced blob was compared byte-for-byte against its artifact's `deployedBytecode` before deletion, and every stated code hash equals the artifact's `codeHash`.
- The verification command shown on each page was executed for all seven artifacts and reproduces the stated code hash.
- All seven artifact links point to files present on `main`.


flyq (commented):
1. **State the working directory for the verification command.**
   `jq -r .deployedBytecode crates/system-contracts/artifacts/…` assumes the reader runs it from a repo checkout root; the spec pages are published on docs.megaeth.com where that context is not obvious.
   A short lead-in such as "From the repository root:" above the code block (once per page) would remove the ambiguity.
2. **Consider one sentence making the hash's normative role explicit.**
   The PR body says "the code hash stated on each page remains the normative identity of the bytes", but no page says this in so many words.
   A single sentence on the system-contracts overview page (e.g. "The code hash is the normative identity of a contract version; the linked artifact JSON is the canonical source of the full bytes.") would make the intent survive future edits.
   The truncated 12-byte prefixes are cosmetic either way — all seven start with the same Solidity prologue (`0x6080604052…`), so they carry no discriminating value and could equally be dropped.

RealiCZ:
@flyq 
Thanks for the review — both suggestions taken:

1. The verification lead-in on every page now reads "To verify the code hash, from the repository root:". It is attached to each code block (rather than once per page) so that each version section on the Oracle page stays self-contained for readers who deep-link to a specific version.
2. Added to the system-contracts overview (Deployment Semantics): "The code hash stated on each contract page is the normative identity of that contract version. The artifact JSON linked on each page is the canonical source of the full deployed bytecode."

On the truncated prefixes: kept them intentionally. The format follows the OP specs convention, where the prefix is likewise the shared Solidity prologue plus a full-bytecode link:

https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/ecotone/derivation.md#gaspriceoracle-deployment

It serves as a visual cue that the value is bytecode, not as a discriminator.

flyq (approved):

Troublor (approved):

PR #350, opened 2026-07-21
fix(rex6): charge the creator once across reverted CREATE retries

RealiCZ:
## Summary

- Follow-up to #348 (Rex6 preparation): one Rex6 accounting fix plus one spec clarification, both found in review of the migration.
- **CREATE retry double-charge (Rex6)**: the creator's nonce-bump account update survives a child CREATE's revert (revm bumps the nonce before the create checkpoint), and under Rex6 its charge lives on the parent's lane — but the frame unwind also reset the parent's dedup flag, so a contract that retries CREATE after a constructor revert was charged the same creator account update twice toward data-size / KV-update accounting. `push_create_frame` no longer arms the unwind under Rex6; Rex5 keeps its frozen behavior (its charge lived on the child lane and was correctly dropped with it).
- **Fee-reward accounting semantics (docs)**: the Rex6 post-execution fee-reward writes are recorded after the transaction's execution result is final, so the spec now states explicitly that this usage feeds the transaction's reported totals and the block-level cumulative counters only, and MUST NOT retroactively change the transaction's outcome.

## Test plan

- New regression test: a contract CREATEs, the constructor reverts, and the same frame retries — asserts the creator account update is charged exactly once (red on the pre-fix code, green after).
- Full mega-evm test suite, clippy / fmt / Prettier all clean; Rex≤5 is byte-identical (the change is spec-gated).

## Notes

- Based on #348's branch; will be retargeted to `main` once #348 merges.


Troublor (approved):

flyq (approved):
LGTM

vincent-k2026 (approved):
Approve — concur with @Troublor and @flyq.

**Fix (traced).** `charged_parent_update` is consumed only by `pop_frame_unwind_parent`, so gating it off under Rex6 keeps the parent's `target_updated` set across the child revert, and a CREATE retried after a constructor revert no longer re-charges the creator account-info write. The CALL path is correctly left armed — a value-transferring CALL's balance move *is* rolled back with the callee revert, so its charge belongs on the child lane. Rex5 stays byte-identical (`&& !self.rex6_enabled` is a no-op when Rex6 is off).

**Discrimination.** `test_rex6_nested_create_revert_then_retry_charges_creator_once` asserts `rex6 − rex5 == 1` write / 1 KV; pre-fix rex6 charges twice (delta 2 → red), post-fix delta 1 (green), with rex5 the frozen baseline — and the diff mutation gate (3/3) mutates exactly this line. Pairs cleanly with the single-revert test as a positive control.

**Non-blocking nit.** The test uses rex5 as a *moving* baseline; an absolute assertion on the rex6 value would be strictly more discriminating and independent of any future rex5 drift. The fee-reward doc clarification correctly pins the "a tx-level limit crossed solely by these post-final writes does not retroactively fail the tx" semantics.

PR #351, opened 2026-07-22
feat: harden SequencerRegistry sequencer rotation (Rex6, contract v2.0.0)

RealiCZ:
## Summary

- Ported from megaeth-labs/mega-evm-internal#32 onto the public Rex6 line (base: `cz/feat/rex6-preview`, will be retargeted to `main` once #348 merges). The `#340` docs cherry-pick from the internal branch is dropped — it is already in the base.
- Hardens `SequencerRegistry.scheduleNextSequencerChange` (Rex6, contract v2.0.0) against scheduling a sequencer rotation to a key nobody holds — a liveness deadlock that cannot be fixed on-chain once triggered.
- The new sequencer key must now sign an EIP-712 proof over the exact `(newSequencer, activationBlock)` being scheduled, and the activation block must be at least a config-seeded `minRotationDelay()` blocks in the future.
- Ships as an in-place, storage-preserving bytecode upgrade at Rex6 activation (following the Oracle's versioned-bytecode precedent) — slots 0–12 are preserved byte-for-byte, only the new `_minRotationDelay` slot (13) is written.
- The system-address role's scheduling path is untouched; symmetric hardening there is deferred to a follow-up PR.

## Port notes (vs the internal branch)

`transact_deploy_sequencer_registry` was rebased onto this branch's generic `SystemContractSpec` / `transact_deploy` refactor: version selection feeds the spec (`target_code` / `target_code_hash`), the fresh-deploy seed gains the `_minRotationDelay` slot, and the Rex6 v1→v2 upgrade stays an explicit early path — it must preserve storage and record the slot's true original value, which the generic force-create deploy intentionally does not express.

## Replay protection (why no nonce)

The signature commits to `(chainId, verifyingContract, newSequencer, activationBlock)` via the EIP-712 domain and struct, and combined with `onlyAdmin` gating this is sufficient:

- Only the admin can call `scheduleNextSequencerChange`, so a non-admin cannot replay a captured signature.
- A signature cannot be reused for a different `newSequencer` or `activationBlock` — the digest would differ.
- Replaying the identical `(newSequencer, activationBlock)` pair is idempotent and harmless — it reinstates the same intended state.
- A stale signature naturally expires as `block.number` approaches the old `activationBlock`, since a new schedule call must satisfy `activationBlock >= block.number + minRotationDelay()` at call time.

## `minRotationDelay` — deliberate tradeoff, needs sign-off

This PR does not set a production value — this branch has no `chain.rs` mainnet/testnet hardfork schedule yet, so `SequencerRegistryRex6Config` is not wired to real chain IDs.
**Before Rex6 is scheduled on a real network, governance/ops must pick and sign off on a `rex6_min_rotation_delay` value.**
The design's proposed production value is `21_600` blocks (~6 hours at ~1s/EVM block), applied uniformly with **no emergency fast-path** — a bad-registration reaction window matters most exactly when the sequencer key rotation is urgent (e.g. suspected compromise), so a bypass would reopen the race this feature exists to close.
Devnets / e2e harnesses should use a small value (5–10 blocks) for fast iteration.

## Spec gating

Rex5 is already live on both mainnet and testnet with real seeded admin/sequencer addresses, so this cannot edit the Rex5 initial deployment — it lands as Rex6, gated by `MegaSpecId::is_enabled(REX6)`.

## Downstream impact

- mega-reth: bump the mega-evm dependency tag; no code changes required for this PR alone.
- test-client / chain-ops: rotation-scheduling calls now require the new key's signature (`rotationDigest()` returns the exact message to sign); tooling and runbooks need a follow-up in their own repos before Rex6 goes live anywhere.
- megaeth-chainspec (mega-reth side): once mega-reth actually schedules Rex6 activation, it needs the `rex6_min_rotation_delay` value from the sign-off above.

## Test plan

- [x] Foundry: valid/invalid signature, wrong-signer, signature-over-different-target, malformed/wrong-length, high-s malleability, cross-chain replay, delay boundary, cancel-without-proof, re-schedule-after-cancel, overwrite-requires-fresh-proof (92 tests pass)
- [x] Rust: version-dispatch (V1/V2), storage-preservation across upgrade, grandfathered V1-scheduled rotation activating both at and after the upgrade block, full end-to-end rotation via block executor (`--test rex6`: 114 pass)
- [x] `cargo fmt` / `cargo clippy --workspace --all-targets` / `cargo check --no-default-features` — all clean


Troublor (approved):

flyq (approved):
LGTM

vincent-k2026 (approved):
Approve — concur with @Troublor and @flyq.

**Security.** The possession proof binds `chainId + verifyingContract` (EIP-712 domain) and `newSequencer + activationBlock` (struct); `_recoverRotationSigner` fails closed on wrong length / high-`s` malleability (`s > secp256k1n/2`) / `v ∉ {27,28}` / `ecrecover == 0`, with `ecrecover` last. Cancel (`activationBlock == type(uint256).max`) deliberately ignores the signature so a lost new key can never block a cancel — the right liveness valve. Replay analysis holds: `onlyAdmin` + full digest binding + the `activationBlock >= block + delay` recheck ⇒ a captured signature can only reinstate the same intended `(seq, block)` (idempotent) and naturally expires.

**Migration.** The in-place upgrade writes only slot 13 (`mark_touch`, not created, so live roles/pending/history survive), and `new_changed(original=0, delay)` records the true original for an honest state diff. The executor apply-then-deploy reorder at the Rex6 block is correct and well-commented: the v2 upgrade must commit **last**, or an apply outcome carrying the pre-upgrade v1 `code_hash` would clobber it. Grandfather upgrade-block case is e2e-tested; the fallback schedule now attaches `SequencerRegistryRex6Config` and fail-fasts if it's ever missing. The Foundry negative-control matrix (wrong-signer, different-target, high-`s`, invalid-`v`, cross-chain replay, reschedule-after-cancel, overwrite-requires-fresh-proof, delay-boundary) is thorough.

**Follow-up (non-blocking, worth a tracked next PR).** `scheduleNextSequencerChange` is hardened but `scheduleNextSystemAddressChange` is not — the same "schedule to a key nobody holds" deadlock this PR closes for the sequencer remains open for the system-address role (which authorizes system txs + Oracle). The PR body defers this; the symmetric hardening is the structural fix. And `rex6_min_rotation_delay` needs a governance-approved value before Rex6 is scheduled on a real network (fallback is 1 block for dev only).

PR #358, opened 2026-07-28
docs: specify compute gas accounting and pin it with a cross-spec test harness

RealiCZ:
## Summary

Compute gas is one of the two dimensions of the dual gas model and has its own per-transaction limit, but it had no real specification — roughly 40 lines split across `resource-accounting.md` and `dual-gas-model.md`, saying little beyond "same as vanilla EVM gas". Meanwhile the implementation records it at nine distinct sites, and the measurement itself changes across five specs.

This PR adds `docs/spec/evm/compute-gas.md` as the authoritative page, migrates the scattered content into it, and adds a test harness that pins the specification against the implementation.

The motivating use case is the upcoming per-opcode metering hot-path optimisation: that work must preserve recorded compute gas byte-for-byte on every frozen spec, and until now there was neither a written contract nor a regression net for it.

## The specification

Compute gas is defined as a **derived measurement** of the EVM gas inherited from Optimism Isthmus / Ethereum Prague, not as a gas schedule of its own. Restating 256 opcode costs would duplicate the inherited spec, rot on every upstream bump, and still leave the consensus-critical question unanswered — how much of an operation's inherited cost is recorded, and when.

What the page defines instead:

- **The measurement window** — `recorded = (gas_before − gas_after) − storage_gas_charged − forwarded_child_gas`, with the window boundaries stated normatively. The opening point is consensus-visible: it determines how much has been recorded when an opcode halts partway through, and therefore which limit is reached first.
- **Three exclusions** — storage gas (with the two equivalent implementation forms, and the rule that they must not be combined), forwarded child gas (including the Rex5 `CALL_STIPEND` adjustment and the frozen MiniRex quirk where three CALL-family opcodes are not subject to the 98/100 cap), and refunds.
- **Seven opcode metering classes**, with per-spec assignment expressed as inheritance plus overrides, so every one of the 256 opcodes is covered without restating EVM gas costs.
- **Every non-opcode recording site** — transaction intrinsic, precompiles, contract-creation code deposit, keyless-deploy sandbox, and system contract interception (which records none). These are closed by an accounting identity, so a missing or extra site does not balance.
- **Limits and exceed behaviour** — the transaction limit, the detained limit, the Rex4+ per-call-frame budget, and the frame-local-revert vs transaction-level-halt split.

## What moved

The compute-gas section of `resource-accounting.md` and the metering *amount* from `dual-gas-model.md` are replaced by pointers, so there is a single normative source. `dual-gas-model.md` keeps the metering *order*, which is genuinely a storage↔compute interaction.

`dual-gas-model.md`'s total-gas equation is corrected: it was stated as an identity between the receipt's `gas_used` and `compute + storage`, which does not hold once an EVM refund is settled.

## The test harness

`crates/mega-evm/tests/compute_gas` contains two layers:

**A golden snapshot** — 32 programs covering every metering class and every non-opcode recording site, executed under all nine specs, with the resulting compute-gas readings checked in. Any change to a recorded amount on any spec surfaces as a diff. Regenerate with `UPDATE_COMPUTE_GAS_SNAPSHOT=1 cargo test -p mega-evm --test compute_gas`.

**Seven rule assertions** — the snapshot records numbers but not *why* they are what they are; a refactor that preserves a total while breaking the underlying rule would still pass it. These tests assert the rules directly: the `CALL_STIPEND` exclusion (with a negative case for each of its three conditions), the CREATE2 memory-expansion window split, the KZG fixed cost, interception costing zero, refund exclusion, the frozen MiniRex forwarding quirk, and Equivalence recording nothing.

Reviewers should read the snapshot diff as part of the review. **Rows for Equivalence through Rex5 must never change** — a diff there is a replay-breaking regression, not a value to bless.

## Verification

The specification was checked against the implementation by two independent automated reviews, which found and this PR fixes:

- **KZG point-evaluation cost was written as 50,000 "inherited unchanged".** MegaETH overrides it to 100,000 from MiniRex onward (`precompiles.rs:98`), as `precompiles.md:39` already documented. An implementer following the new page alone would have disagreed with the chain on every block touching `0x0a`.
- **The gas-rescue rule was stated unconditionally.** The keyless-deploy dispatch path only rescues from Rex6 onward; pre-Rex6 the sender loses the whole envelope. The exceed matrix now carries that case explicitly.

Both errors survived the initial test suite because the corpus contained no KZG case and no keyless-deploy case. The corpus now covers KZG, and the rule is pinned by a test that hard-codes 100,000 rather than importing it, so a change to the implementation constant fails rather than being silently followed.

Local gate: `cargo sort`, `cargo fmt --all --check`, `cargo clippy --workspace --lib --examples --tests --benches --all-features --locked`, `cargo test --workspace`, and `mise run lint` (lychee + markdownlint + prettier) all pass.


flyq (approved):
# Review: PR #358 — compute gas specification + cross-spec test harness

## Overview

The PR adds `docs/spec/evm/compute-gas.md` (550 lines) as the single normative source for compute-gas accounting, converts the scattered content in `resource-accounting.md` and `dual-gas-model.md` into pointers, corrects the previously wrong receipt identity (`gas_used` ≠ `compute + storage` once refunds settle), and pins the whole thing with a new test target: a 32-program × 9-spec golden snapshot plus 8 rule-assertion tests. No `src/` changes — the `spec:unchanged` / `api:unchanged` labels are accurate.

**Verdict: this is a high-quality PR and I found no correctness errors in it.** I independently verified the page's normative claims against the implementation (three parallel deep-verification passes covering ~30 claim clusters), ran the new tests, and checked every cross-page reference. Everything confirmed. The remaining findings below are letter-level wording issues and coverage suggestions — none replay-affecting, none blocking.

## Specification correctness (verified against implementation)

All confirmed, each traced to code:

- **Measurement window** — `recorded = (gas_before − gas_after) − storage_gas_charged − forwarded_child_gas` with saturating subtraction matches `record_storage_compute_gas!` (`instructions.rs:464-514`) and the plain wrappers; every abort path (stack underflow, storage-charge OOG, inner-instruction error) skips the trailing record exactly as the single-record rule states.
- **Forwarded-gas exclusion** — the three-branch pseudocode matches `instructions.rs:473-493`, including the Rex5 `CALL_STIPEND` subtraction gated on scheme+value, and the pre-inflation claim holds: Rex4 stipend inflation happens in `before_frame_init`, strictly after the parent's window closes.
- **MiniRex forwarding quirk** — real: the mini_rex table wires only `CALL` through `forward_gas_ext` (`instructions.rs:691-694`); Rex swaps all three.
- **CREATE2 window movement** — MiniRex–Rex4 second window skipped on inner failure (`instructions.rs:1740-1743`), Rex5 eager record (`:1632-1636`), Rex6 single window, and Rex6 does halt on oversized initcode before any expansion (`:1613-1626`).
- **Non-opcode sites** — intrinsic records revm's pristine `initial_gas` before the MegaETH storage additions (`execution.rs:697-719`); the Rex5+ precompile formula, the *structural* KZG branch condition (a limit comparison, not an error-variant match — `precompiles.rs:234-251`), and the 100,000 override at `precompiles.rs:98` all match; keyless overhead is charged from Rex2 but recorded only from Rex3 (`sandbox/execution.rs:174-181`), rescue Rex6-gated (`:200-202`).
- **Exhaustiveness** — the accounting identity's site list is complete: every mutation of the live compute counter routes through the five listed sites. Notably, the Rex6 post-execution fee-reward accounting merges data-size/KV/state-growth only, no compute gas (`execution.rs:952-955`), so the "MUST NOT record at any other site" claim survives.
- **Limits & exceed** — 1B/200M limits, strict `>` at both levels, 98/100 frame share, Rex4-only pre-inflation cap, frame-check-before-tx-check with no short-circuit, `MegaLimitExceeded(uint8,uint64)` frame revert, rescue excluding the stipend portion, and the detention halt's distinct `VolatileDataAccessOutOfGas` variant all match.
- **Class assignment** — the MiniRex base + per-spec override tables match the instruction tables exactly, including the SELFDESTRUCT reported-dimension priority (data size → KV → compute → state growth, `limit.rs:358-380`) and the Rex1 detention reset (`compute_gas.rs:172-180`).

The PR body's two self-reported fixes (KZG 100,000; keyless rescue matrix) are correctly incorporated, and the test that hard-codes 100,000 instead of importing it is the right pinning choice.

## Findings (minor, worth a follow-up commit)

1. **Rex6 window-boundary claim doesn't literally match SELFDESTRUCT** — `compute-gas.md:73` says every storage-affecting opcode's window "MUST open before the opcode's storage-gas charge", but Rex6 SELFDESTRUCT still charges storage gas (`instructions.rs:2033`) before its window opens (`:2297`). The amounts are identical under the page's own two-form equivalence rule, so this is unobservable — but a normative MUST that mega-evm itself doesn't literally satisfy invites a false "non-compliance" report from a future auditor. Suggest carving SELFDESTRUCT out or restating the Rex6 block in terms of recorded amounts.
2. **Keyless frame-local branch is currently unreachable** — `compute-gas.md:415` specifies the frame-local `InsufficientComputeGas` revert, but KeylessDeploy interception is depth-0-only (`intercept.rs:301`) and the tracker frame stack is empty there, so `check_limit` can only produce a transaction-level exceed today. The code branch exists as written (defensive), so the spec isn't wrong, but a one-line hint that only the transaction-level branch is reachable on current dispatch would save implementers hunting for a repro.
3. **"Current frame's remaining budget" is ambiguous** — in the Rex5+ precompile bound (`compute-gas.md:281`), the frame in question is the precompile invocation's *own* pushed frame (the forwarded 98/100 share), not the caller's frame. Worth one clarifying clause; an implementer bounding by the caller's frame would compute a different cap.
4. **Convention deviation should be explicitly blessed** — `docs/spec/AGENTS.md` says main content describes the latest stable spec only and unstable behavior goes in `<details>`. The page argues its cross-spec structure well in Rationale, but the Rex6 rescue rule additionally sits in main prose (`compute-gas.md:401`, `:418-421`; `resource-limits.md:101-103`) while `keyless-deploy.md` wraps the same content in a `<details>` block — inconsistent placement for identical content. Either wrap those or codify this page's exception in `AGENTS.md`. (Related nit: two `{% hint info %}` blocks carry normative MUST NOTs; AGENTS.md reserves hints for non-normative notes.)

## Test harness

- All 12 tests pass locally; `cargo fmt`, `clippy` (only the pre-existing MSRV-config warning), and Prettier on docs are clean. Cargo auto-discovers `tests/compute_gas/main.rs`; no manifest change needed.
- The claim tests are well-designed: differential pairs isolate exactly one variable (byte-identical programs, same push widths), each `CALL_STIPEND` condition gets a negative case, and the KZG constant is deliberately hard-coded so an implementation change fails instead of being followed.
- I spot-checked the snapshot arithmetically: intrinsic 21000/60000, the +2,300 Rex5 stipend deltas, the +100,000 Rex5 KZG delta, the refund row, pre-Rex2 `InvalidFEOpcode` SELFDESTRUCT rows, and the Rex5 eager-expansion value 763,907 (which works out exactly: 21,012 + quadratic expansion of 18,750 words). Rows are internally consistent with the stated rules.
- **Coverage gap worth closing:** the corpus contains no case that actually *halts on a compute-gas or detention limit*, yet halt position ("which limit is reached first, at which opcode") is the page's stated reason window placement is consensus-visible. A corpus entry that trips gas detention with default limits (volatile read, then a burn loop) would pin the recorded-at-halt amount and the rescue arithmetic in the snapshot. Similarly, per-frame-exceed (`MegaLimitExceeded`) recording isn't exercised here. Fine as a follow-up.
- Nit: `main.rs:103-196` has three near-identical `transact*` helpers; a single core function with the envs/output variants layered on would drop ~50 lines. Optional.

## Docs quality

- Structure follows the spec-page template exactly (frontmatter with `spec: Rex5`, Motivation → Specification → Constants → Rationale → Security Considerations → Spec History); every constant referenced in the body appears in the Constants table; all 14 cross-page anchors I checked resolve; `SUMMARY.md` placement is sensible.
- Cross-page consistency is genuinely good: the Spec History bullets match the upgrade pages (including the subtle Rex2-introduce vs Rex3-record split for the keyless overhead), `keyless-deploy.md`/`resource-limits.md`/`glossary.md` all agree with the new normative source, and no page still claims the old receipt identity.
- The `dual-gas-model.md` receipt correction is right, and the metering-order edits sit properly inside the Rex6 `<details>` block.

## Recommendation

Approve. Items 1–3 are small wording fixes I'd fold in before merge since this page will be treated as normative; item 4 and the halt-path corpus case can be follow-ups.

Troublor (approved):

Troublor (approved):

PR #360, opened 2026-07-30
feat: seal REX6 and open REX7

RealiCZ:
## Summary

REX6 is frozen and REX7 opens as the unstable spec.

Freezing REX6 is a documentation and tooling operation, not a behavior change: no execution path changes for any spec, and the compute-gas snapshot diff is purely additive — 32 new Rex7 rows and one header line, with every frozen-spec row byte-identical.

The spec pages carry the weight. Every one of the 39 `Rex6 (unstable)` `<details>` blocks is lifted into main prose, replacing the behavior it superseded, and that superseded behavior moves into each page's Spec History. All 22 spec pages advance to `spec: Rex6`.

Doing that surfaced a distinction the docs had no way to express. The conventions used "stable", "activated", and "not-yet-activated" interchangeably, which works only while every frozen spec is also live. REX6 is frozen with no activation timestamp on either network, so the two properties are now stated separately — maturity (`unstable` / `frozen`) and network status (`not scheduled` / `scheduled` / `active`, per network) — in `docs/spec/AGENTS.md`, `AGENTS.md`, and the `doc-readability` skill's matching check. Both upgrade pages carry "Not yet scheduled" activation tabs.

A second gap in the same conventions is now written down: the `<details>` rule for unstable content was stated without an exception for the pages whose subject is the spec ladder itself, even though those pages had always listed the unstable spec in main prose (this PR's base carries a full `### REX6` section on `hardfork-spec.md` while declaring `spec: Rex5`). A folded block cannot express that a rung exists, so hiding it would show a reader looking up which specs exist a ladder that stops one short. The exception is now stated alongside the glossary and replay-page ones, and requires each such entry to label the spec unstable where it is described.

REX7 inherits REX6 unchanged: its instruction table, runtime limits, and precompile set all delegate, and no spec gate references it yet.

## Deliberately not done

Rex6 and Rex7 have **no** mainnet or testnet activation timestamps — those are not yet decided. A test pins that: both resolve to `ForkCondition::Never` on both canonical schedules, and both chains still resolve to `REX5` at `u64::MAX`, so an accidental schedule entry fails rather than silently forking a live chain at that timestamp.

Scheduling REX6 is a separate change: it sets the timestamps and, at that point, the concept pages' network-status notes.

## Spec gaps closed

Sealing makes the concept pages normative for REX6, which exposed content that was only ever informative or outright wrong.

Three REX6 behaviors existed only in `upgrades/rex6.md`, a page that declares itself informative, so no `<details>` block carried them and the lift could not reach them. Each is now specified where it belongs: the EIP-7702 authorization-list skip on a pre-frame limit exceed (consensus-visible — it decides which authority writes survive the halt) in `resource-limits.md`, the KeylessDeploy deploy-address occupancy read through the journal in `keyless-deploy.md`, and the creation-opcode halt ordering for oversized initcode and static frames in `contract-limits.md`.

Two pages were wrong. `selfdestruct.md` said a `SELFDESTRUCT` whose target already exists incurs no charges; under REX6 a non-zero balance credit to an existing distinct target records data size and a KV update, which `resource-accounting.md` already stated — an implementer following the `SELFDESTRUCT` page would have under-counted. `compute-gas.md` claimed both keyless-deploy exceed branches rescue the outer transaction's gas; only the transaction-level branch does.

One overclaim was load-bearing enough to matter at the freeze. `hardfork-spec.md` and `upgrades/rex6.md` describe REX6 as bringing the source-side `SELFDESTRUCT` — one whose executing contract is the beneficiary — under *both* beneficiary detention and `disableVolatileDataAccess`. Only the second half exists: the REX6 check sits in the volatile-access-disabled revert guard, detention needs no `SELFDESTRUCT`-specific hook because the beneficiary is already marked when reached as a transaction recipient or call target, and `gas-detention.md`'s trigger list correctly omits it. Sealing would have promoted that to a normative `MUST` on `selfdestruct.md` that mega-evm does not implement, so an alternative implementer following the page would add the trigger and diverge on a halt-visible path. The two are now stated separately, with the EIP-7702-delegated CALL case — which does engage both — kept intact. Relatedly, `selfdestruct.md` was the only page naming the 40-byte target-write charge `SELFDESTRUCT_BENEFICIARY_DATA_BYTES`; it now uses `ACCOUNT_UPDATE_DATA_SIZE`, the name `resource-accounting.md` defines and states the same rule with.

## Tooling

`MegaHardforkConfig::with_all_activated_through(spec)` states "a chain running spec N" once, walking the fork ladder the `hardfork!` macro already emits as `MegaHardfork::VARIANTS` — generated from the same variant list that declares the enum, so it is exhaustive by construction. Previously each caller subtracted the forks above N by hand, which every new spec silently widened — the `rex4` / `rex3` test configs had been resolving one spec too high since Rex6, and three helpers named for Rex6 became Rex7 chains the moment Rex7 existed. The builder states the whole ladder rather than only adding to it: a later fork already registered is removed, so the resolved spec is a function of `spec` alone and not of builder call order. A test pins `spec_id(0) == spec_id(u64::MAX) == spec` for every spec, from an empty config and from a fully-activated one, so a missing entry fails once, centrally.

The mutation spec-gate freezes REX6, with two corrections to how the lists are used:

- `FROZEN_SPECS` had been serving as both the mutable-source set and the adjacency universe, which dropped `is_enabled(REX6) ==> is_enabled(REX7)` — the "gated one fork too late" mutant that silently disables frozen REX6 behavior, and precisely what freezing REX6 is meant to catch. Split into `ALL_SPECS` (adjacency) and `FROZEN_SPECS` (sources).
- `rex6` stays out of `INSTRUCTION_MODULES`: `rex6::instruction_table` returns `rex5::instruction_table` verbatim, so a misroute mutant there is provably equivalent and survives by construction. REX6 expresses every difference as `is_enabled` dispatch inside the shared handlers, which boundary shift already probes.

Spec lists that had gone stale several specs back and would now read as contradicting the progression are refreshed: the root README (REX4), `ARCH.md` (REX2), the `mega-evm` and `mega-evme` READMEs (REX4 / `Rex2` default), `mega-t8n`'s `--state.fork` help (Rex2), `evm/AGENTS.md`'s instruction-module list (rex4), and the spec progressions in `docs/spec/overview.md` and `glossary.md`, which stopped at REX5 and never gained REX6.

`bench_compare.py`'s `ROW_ORDER` gains `rex6` and `rex7`, and `SPEC_IDS` — the list the benches actually register — gains `rex7`. Each had drifted from the other: `ROW_ORDER` stopped at `rex5`, so the `rex6` rows were dropped from the comparison table, and `SPEC_IDS` stopped at `rex6`, so no bench emitted a `rex7` row to drop. The filter did either silently, so a `ROW_ORDER` entry that produced no rows is now named in the table output.

## Review guidance

The snapshot diff is the regression net for the claim that nothing moved. Read `crates/mega-evm/tests/compute_gas/snapshot.txt`: **every row from Equivalence through Rex6 must be unchanged**, and the only additions are the 32 Rex7 rows.

The snapshot alone does not pin Rex7 ≡ Rex6 — its rows differ by the spec-name column, so a drifted Rex7 row would still render as a well-formed snapshot and could be blessed by a regeneration. `test_rex7_matches_rex6_on_every_program` compares the readings directly instead.

For the doc diff, the useful check is whether a lifted block's prose describes REX6's behavior and its predecessor's went to Spec History, not the reverse. Two such inversions were caught in review and are fixed above.

## Test plan

- `cargo test --workspace` — 33 suites green
- `cargo check -p mega-evm --target riscv64imac-unknown-none-elf --no-default-features` — no_std
- `cargo clippy --workspace --lib --examples --tests --benches --all-features --locked` — clean
- `cargo fmt --all --check`, `cargo sort --check --workspace --grouped --order package,workspace,lints,profile,bin,benches,dependencies,dev-dependencies,features`
- `mise run lint` — lychee 0 errors, markdownlint 0 issues, prettier clean
- `python3 mutants/operators/spec-gate/generate.py` — regenerated; `spec-gate.rules` gains 3 lines (`REX5→REX6`, `REX6→REX5`, `REX6→REX7`), of which `REX6→REX7` is the one a naive single-list freeze would have missed
- No benchmarks: no execution path changes.

New tests: canonical-schedule lockdown (Rex6/Rex7 never scheduled), `with_all_activated_through` resolution for every spec, `SpecName::from_mega_spec` coverage of the latest spec (its wildcard arm means a new spec falls through to `Unknown` without a compile error), and Rex7 ≡ Rex6 across the compute-gas corpus.

## Labels

`spec:new`, `comp:core`, `api:compatible` — `MegaHardforkConfig::with_all_activated_through` is additive.


docs/spec/overview.md
```diff
@@ -70,6 +70,8 @@ Contracts deployed under a given spec will continue to behave identically, regar
...
 - **REX4** — Per-call-frame resource budgets, relative gas detention, [storage gas stipend](glossary.md#storage-gas-stipend), MegaAccessControl and MegaLimitControl system contracts.
 - **REX5** — SequencerRegistry system contract, Oracle v2.0.0 with dynamic system address, caller-account update deduplication, storage-gas-stipend separated-allowance model, value-transfer CALL/CALLCODE parent compute-gas attribution, CREATE code-deposit compute-gas atomicity, EIP-2935/EIP-4788 pre-block gas floor with fail-closed block rejection, CREATE2 empty-initcode short-circuit, KeylessDeploy trailing-bytes rejection and empty-code log forwarding.
+- **REX6** — Unified per-opcode gas metering order, consolidated EIP-7702 authorization accounting, CREATE-frame accounting corrections, KeylessDeploy sandbox hardening, post-execution fee-reward accounting, system-originated transaction metering exemption, extended beneficiary detention coverage, and SequencerRegistry v2.0.0 rotation hardening. Frozen, but not yet scheduled on any network.
+- **REX7** — The spec currently open for development. No behavioral change over REX6 yet.
```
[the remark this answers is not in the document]
RealiCZ: Not taking the `<details>` move, but you found a real hole in the convention — fixed in 904abfa.

The Rex7 entry carries no behavior to hide. It states that there is none yet:

> **REX7** — The spec currently open for development. No behavioral change over REX6 yet.

And `<details>` cannot express "this rung exists". Folding it means a reader looking up which specs exist is shown a ladder that stops one short of reality — on the two pages whose entire job is to answer that question.

The convention already worked this way, it just never said so. On this PR's base, `hardfork-spec.md` declares `spec: Rex5` and carries a full `### REX6` section in main prose — status line plus every REX6 behavior bullet — while REX6 was the unstable spec. So the rule you cite has been read as scoped to concept pages ("Main content in **concept pages** MUST describe the latest frozen spec's behavior only"), with the roster pages as an unwritten exception, next to the two exceptions that *are* written down (glossary, replay page).

Leaving it unwritten is the actual defect, so it is now written:

> **Roster-page exception**: a page whose subject is the spec ladder itself (`overview.md`, `hardfork-spec.md`, `upgrades/overview.md`, and the progression line in `glossary.md`) MUST carry the unstable spec in its progression and its per-spec summary, outside `<details>`.
> A reader looking up which specs exist must not be shown a ladder that stops one short of reality, and a `<details>` block cannot express "this rung exists".
> Every such entry MUST label the spec unstable where it is described, and MUST stay a summary — the normative text lives in the concept pages, still gated by `<details>` there.

That last clause is what addresses your underlying concern — that unstable behavior could read as part of the frozen Rex6 specification. The guard is the label, not the folding: `hardfork-spec.md` and `upgrades/overview.md` already marked Rex7 unstable, and `overview.md` now does too. Normative Rex7 text still goes in `<details>` on the concept pages, where an implementer would actually read it.

Your P2 on `with_all_activated_through` was right and is fixed in a9b34d7 — the builder now removes what it does not activate, and the test pins the contract starting from a non-empty config.

docs/spec/evm/selfdestruct.md
```diff
@@ -32,6 +32,7 @@ If the executing contract was not created in the same transaction, `SELFDESTRUCT
 - preserve the contract's code and storage.
 
 `SELFDESTRUCT` targeting the [beneficiary](../glossary.md#beneficiary) MUST trigger beneficiary [gas detention](gas-detention.md).
+A `SELFDESTRUCT` whose **executing contract** is the beneficiary MUST also trigger it: the operation reads and zeroes the executing contract's balance, so it observes beneficiary state regardless of its stack target.
```
flyq: **[Major]** This sentence makes a source-side **detention** trigger normative, but the frozen implementation doesn't have one.

The REX6 source-side check exists only in the volatile-access-**disabled** revert guard; the wrapper comment (`crates/mega-evm/src/evm/instructions.rs:1091-1096`) states that for detention "the beneficiary is already balance-marked before its own code runs (it is reached as the tx recipient or a CALL target, both of which mark it), so detention engages without a SELFDESTRUCT-specific hook."

That invariant has a hole: a CREATE/CREATE2 **constructor** executes at the derived address with no beneficiary marking — tx-level marking covers only `TxKind::Call` (`crates/mega-evm/src/evm/context.rs:562`), and the SELFDESTRUCT host hook marks the *target* only (`crates/mega-evm/src/evm/host.rs:104`). If the block beneficiary equals a predictable CREATE2 address, a constructor `SELFDESTRUCT` to a distinct target reads and zeroes beneficiary state with **no detention** — mega-evm does not trigger it, while this sentence says a node MUST. A conforming alternative implementer following this page would add the trigger and diverge from mega-evm on that path; detention is halt-visible, so this is consensus-relevant. Note that `gas-detention.md` (the normative trigger list) correctly **omits** this trigger, so the two pages currently disagree.

The `disableVolatileDataAccess` half is implemented and tested (`test_rex6_selfdestruct_source_eq_beneficiary_disabled_reverts`); only the detention half overclaims.

Suggested fix (doc-only): scope this sentence — and the Spec History bullet at line 104 — to the disabled-state guard, matching `mega-access-control.md`. `upgrades/rex6.md` Summary item 7 and its description frontmatter bundle "beneficiary detention and disableVolatileDataAccess" the same way; that text predates this PR, but this PR seals it, so it deserves the same trim. If the detention trigger is actually *wanted*, that is a REX6 behavior change and must land before this PR freezes the spec (or go to REX7).

docs/spec/evm/selfdestruct.md
```diff
@@ -42,7 +43,17 @@ A node MUST meter this account creation identically to account creation by any o
 - record the creation against the [data size](resource-accounting.md#data-size) (`+SELFDESTRUCT_BENEFICIARY_DATA_BYTES`), [KV updates](resource-accounting.md#kv-updates) (`+1`), and [state growth](resource-accounting.md#state-growth) (`+1`) resource lanes.
 
 A `SELFDESTRUCT` whose transferred balance is zero MUST NOT incur any of these charges, because a zero-value transfer does not create the target account.
-A `SELFDESTRUCT` whose target already exists in state MUST NOT incur these charges.
+
+### Beneficiary Balance Credit to an Existing Account
+
+When `SELFDESTRUCT` transfers a non-zero balance to a target that already exists in state and is **distinct** from the executing contract, no account is created, but the target's balance is written.
+A node MUST record that write against the [data size](resource-accounting.md#data-size) (`+SELFDESTRUCT_BENEFICIARY_DATA_BYTES`) and [KV updates](resource-accounting.md#kv-updates) (`+1`) resource lanes.
```
flyq: **[Minor]** Constant naming drift: this rule charges the existing-target credit as `SELFDESTRUCT_BENEFICIARY_DATA_BYTES`, while `resource-accounting.md` ("SELFDESTRUCT Existing-Beneficiary Data Size") states the same rule as `ACCOUNT_UPDATE_DATA_SIZE`. Both are 40, so there is no numeric conflict, but it's two names for one charge across the two normative statements of the same rule.

Related: the constants-table description below (line 77, "Data-size bytes recorded for the beneficiary account write **on beneficiary creation**") is now under-inclusive — the same constant covers this existing-account write too.

Worth one canonical name (or an explicit equivalence note) plus a widened description before freeze.

crates/mega-state-test/src/types/spec.rs
```diff
@@ -184,8 +188,22 @@ mod tests {
...
             MegaSpecId::REX5,
             MegaSpecId::REX6,
+            MegaSpecId::REX7,
         ] {
             assert_eq!(SpecName::from_mega_spec(spec).to_spec_id(), Ok(spec));
         }
     }
+
+    #[test]
+    fn from_mega_spec_maps_the_latest_spec() {
```
flyq: **[Nit]** `REVIEW.md` mandates the `test_` prefix for test names; this module is consistently unprefixed and `AGENTS.md` permits aligning with local style, so this is a known tension — flagging for an explicit choice (`test_from_mega_spec_maps_the_latest_spec` if REVIEW.md should win).

flyq (commented):
## Review: seal REX6 and open REX7

**Verdict:** the sealing work is solid — I verified the central claims directly on the branch and they hold. One **[Major]** doc-spec divergence should be fixed before merge (inline comment on `selfdestruct.md:35`), plus two smaller inline suggestions.

### Verified

- **Snapshot purely additive** — programmatic diff: exactly 1 removed line (the header comment) + 33 added (32 Rex7 rows + replacement header); every frozen row byte-identical. `test_compute_gas_snapshot_matches` and `test_rex7_matches_rex6_on_every_program` pass; the latter correctly closes the snapshot's spec-name-column blind spot.
- **REX7 wiring complete and purely delegating** — instruction table (`rex7 → rex6 → rex5` verbatim), runtime limits, precompiles, names/`FromStr`, hardfork ladder + trait method, `BlockLimits`, `SpecName`.
- **Canonical-schedule lockdown** pins `ForkCondition::Never` for Rex6/Rex7 on both networks and REX5 terminal at `u64::MAX` — exactly the right pin for "frozen but unscheduled".
- **`with_all_activated_through` is sound** — `insert()` preserves params, `without()` drops the entry; the result is a function of `spec` alone (order-independence pinned by the downgrade assertion). It also fixes the latent rex4/rex3 test-config bug: the old `with_all_activated().without(Rex5)` left Rex6 registered and resolved one spec too high.
- **The three spec-gap closures match the implementation precisely** — the type-4 all-or-nothing authorization skip (`crates/mega-evm/src/evm/execution.rs:611`, gated on `limit_exceeded()`, refund forgone), the keyless occupancy read (`crates/mega-evm/src/sandbox/execution.rs:366` — journaled, cold, code-hash-only, REX6-gated with the load-bearing pre-REX6 fallback), and the creation halt ordering (static-first at `crates/mega-evm/src/evm/instructions.rs:1800`, size-check before prework at `:1634`).
- The two admitted inversion fixes (CALLCODE cold-charge table, keyless-deploy rescue direction) now read correctly, and `dual-gas-model.md` / `resource-limits.md` agree with them.
- **Mutation split is right** — the `ALL_SPECS`/`FROZEN_SPECS` separation restores the `REX6 ⇒ REX7` mutant; excluding `rex6` from `INSTRUCTION_MODULES` is provably correct (verbatim alias ⇒ equivalent mutant).
- **Conventions applied consistently** — all 22 pages at `spec: Rex6`, zero `<details>` blocks remain in `docs/spec/`, tabs/hint styles follow the new maturity vs. network-status rules.
- **Tests run locally** — `compute_gas` (23), `mega-evm --lib` (262), `rex4` (216), `rex5` (209), `rex6` (115), `mega-state-test` (37): all green.

### Not attachable inline

- **[Nit]** PR description says "the rules diff is +1 rule", but `spec-gate.rules` gains 3 lines (`REX5→REX6`, `REX6→REX5`, `REX6→REX7`). Presumably meant "+1 beyond what the naive single-list freeze would have produced" — worth clarifying so the review trail stays accurate.
- The trim suggested for `selfdestruct.md:35` also applies to `upgrades/rex6.md` Summary item 7 and its description frontmatter ("beneficiary detention and disableVolatileDataAccess" bundling) — that text predates this PR (so it is outside the commentable diff), but this PR seals it, so it deserves the same fix. Details in the inline comment.

Specific change suggestions are attached as inline comments.

docs/spec/evm/selfdestruct.md
RealiCZ: Confirmed and fixed doc-side in edf2d52 — the code stays as it is, and here is why.

Every fact in this finding checks out against the branch:

- The sentence is mine, added in `bc9b0cb` on this PR.
- The REX6 source-side check lives only in the volatile-access-disabled revert guard, and the wrapper comment says so explicitly (`crates/mega-evm/src/evm/instructions.rs:1091-1096`).
- `host.rs:104` marks the **target** (`check_and_mark_beneficiary_balance_access(&target)`), never `address`.
- `context.rs:562` marks only the caller and `TxKind::Call(recipient)`, so a CREATE transaction's derived address is not marked.
- `gas-detention.md:73` states only the target trigger, so the two pages did disagree.

So the marking invariant the wrapper relies on does have the hole you describe, and the doc stated a `MUST` that mega-evm does not implement.

**The code is not changing.** That corner — a constructor executing at an address that happens to be the block beneficiary — is a previously adjudicated won't-fix (2026-07-21; a fix had been written and was reverted). Reaching it requires a CREATE/CREATE2 deployment to land on a chosen 160-bit address, which is not computationally feasible, or the sequencer configuring the beneficiary to be a pending-create address. It is not attacker-reachable, and every reachable path to a beneficiary contract's own code — direct transaction recipient, CALL-family target — already marks the account.

**The doc is changing**, because that ruling covers the implementation, not the specification. A normative `MUST` on an unreachable path is still a divergence surface: a second implementer follows the page, adds the trigger, and now differs from mega-evm on a halt-visible behavior. The spec should describe what MegaETH does.

Scoped in five places, since the bundling was wider than the one commentable line — and three of them predate this PR, which is what made it worth catching now rather than after the freeze:

| Location | Was |
| --- | --- |
| `evm/selfdestruct.md` Specification | mine, this PR — the normative `MUST` |
| `evm/selfdestruct.md` Spec History | mine, this PR |
| `hardfork-spec.md:141` | pre-existing |
| `upgrades/rex6.md` item 7 | pre-existing |
| `upgrades/rex6.md` detail bullet | pre-existing — "under `disableVolatileDataAccess` it now reverts **and otherwise engages detention**" |

The `MUST` now reads as a disabled-state guard and says explicitly that it is not to be read as an additional detention trigger, pointing at `gas-detention.md#beneficiary-access` as the complete list. The EIP-7702-delegated CALL case is kept as engaging both — I verified that one separately: `load_account_delegated` marks the resolved delegate on the enabled path, and `gas-detention.md` already specifies the one-hop resolve as a MUST.

One thing I would not adopt from the suggested wording: my first draft justified the absence with "on every path that can reach a beneficiary contract's own code, the beneficiary is already marked" — which is exactly the absolute your CREATE case falsifies. It now states the mechanism (marked as transaction recipient or call target) without claiming completeness.

Your other two are also fixed: the constant is unified on `ACCOUNT_UPDATE_DATA_SIZE` with a widened description (same commit), and the `test_` prefix is applied to all five tests in that module rather than one, since AGENTS.md asks for alignment (caad6b1). The `+1 rule` line in the PR description was wrong as written and now names the three added lines and which one the naive freeze would have missed.

flyq (approved):
LGTM

.github/scripts/bench_compare.py
```diff
@@ -279,6 +279,7 @@ def compare(data, iters: int):
 ROW_ORDER = [
     "revm_pinned", "revm_latest", "op_revm_pinned", "op_revm_latest",
     "equivalence", "mini_rex", "rex", "rex1", "rex2", "rex3", "rex4", "rex5",
+    "rex6", "rex7",
```
Troublor, line 282: Adding `rex6` here fixes the reported symptom, but `rex7` introduces the same drift in the opposite direction: `crates/mega-evm/benches/common/mod.rs:30-38` (`SPEC_IDS`) still ends at `("rex6", MegaSpecId::REX6)`, and that file isn't in this PR. No benchmark emits a `rex7` row, so `rows.get("rex7")` is always `None` and the row is filtered out at line 333.

Net effect: REX7 development starts with zero instruction-count coverage, which is what `AGENTS.md` requires benchmarks for on hot-path changes — and nothing reports the absence.

These two rosters have now been out of sync in both directions within a single PR, which seems like the argument for making the sync mechanical rather than editing both by hand each time. Minimum fix is `("rex7", MegaSpecId::REX7)` in `SPEC_IDS`.

Beyond that, `split_spec` returning `None` for an unrecognized segment (line 304) is what made the original `rex6` gap invisible — a bench under an unknown spec is dropped from the table entirely with no diagnostic. Worth either matching spec segments by pattern (`^(equivalence|mini_rex|rex\d*)$`) so unknown-but-well-formed rows fall into the existing `extras` bucket, or warning when a `ROW_ORDER` entry produced no rows.

crates/mega-evm/src/block/hardfork.rs
```diff
@@ -285,17 +316,32 @@ impl MegaHardforkConfig {
...
+    /// it, so the config would resolve to the newest fork still registered rather than to `spec`,
+    /// and it would silently drift again the next time a spec is introduced.
+    ///
+    /// The result is a function of `spec` alone, not of what the config held before: a later fork
+    /// already registered is removed rather than left in place, so the resolved spec does not
+    /// depend on builder call order.
+    ///
+    /// Patch hardforks are included by their spec, not their position: `MiniRex1` maps back to
+    /// [`MegaSpecId::EQUIVALENCE`], so it is registered for every `spec`.
+    pub fn with_all_activated_through(mut self, spec: MegaSpecId) -> Self {
```
Troublor, line 316: This is the right fix for the call sites, but it patches the symptom rather than the model. `MegaHardforkConfig` represents a chain as a *set of independent fork switches*, when the domain is a *point on a linear ladder*: activating spec N means every spec below N is active, by definition. The set model makes holes representable, and `with_all_activated_through` adds a correct way to build a config without removing the incorrect ones.

The holes aren't hypothetical. Five sites still write the old pattern and now silently resolve to the unstable REX7:

- `src/system/oracle.rs:309-312` — commented "Activate MiniRex only (pre-Rex2, pre-Rex5)", actually `REX7`
- `src/system/oracle.rs:331`, `:378`, `:503` — commented "Rex2 active, Rex5 not active", actually `REX7`
- `src/system/keyless_deploy.rs:135`

They pass only because the deploy gates read per-fork `is_rex_N_active_at_timestamp` predicates rather than the resolved spec — which is the more serious half of the problem. Spec resolution goes through `hardfork()` (highest *registered* active fork), but `block/executor.rs:179,267`, `system/oracle.rs:69`, and `system/sequencer_registry.rs:195,209` gate independently. A config with `Rex7` active and `Rex6` off — writable today through the public `with(fork, condition)` — resolves the spec to `REX7` while the Rex6-gated SequencerRegistry v2.0.0 deploy never runs. That's a consensus-relevant divergence expressible through the published API.

The same modeling gap is why `chain.rs:87`'s fallback silently tracks the unstable spec. `with_all_activated()` is now `with_all_activated_through(MegaSpecId::default())`, and `hardfork_schedule` routes every chain ID that isn't 4326 or 6343 there — so every devnet, local chain, and integration harness runs REX7 at genesis, and will auto-advance to REX8 with no code change. Free today (REX7 ≡ REX6); on the first REX7 behavior change it rewrites those chains' semantics at genesis with no fork boundary, breaking replay of history they already produced. Under a rung-based API the fallback would have to name its rung explicitly instead of inheriting it from `Default`.

What I'd suggest, in increasing order of effort:

1. **Validate the invariant.** Add a `validate()` on `MegaHardforkConfig` rejecting a config whose Mega activation conditions are not non-decreasing along `MegaHardfork::ALL`, called at chain-config load. This matches the repo's existing convention (`HardforkParams::validate()` — "panics loudly at chain-config load time rather than allowing the error to surface at the first block where the fork activates") and is cheap now that `ALL` exists.
2. **Remove the hole-making tools.** Drop or privatize `without(MegaHardfork)`, and make `with_all_activated_through` the only way to set the Mega portion. Sites that genuinely want a non-monotonic config to isolate a single deploy gate — `oracle.rs:309` and `keyless_deploy.rs:135` are the only two — can construct it explicitly and say so in a comment, rather than it being the default idiom.
3. **Make it unrepresentable.** Model the Mega portion as an ordered ladder of activation timestamps rather than a bag of entries, so "spec N is active" and "everything below N is active" are the same statement, and there is nothing to validate.

(1) alone would have caught all five sites above at test setup.

crates/mega-evm/src/block/hardfork.rs
```diff
@@ -35,10 +35,33 @@ hardfork! {
...
 
 impl MegaHardfork {
+    /// Every `MegaETH` hardfork, oldest first.
+    ///
+    /// This is the single source of the fork ladder. Everything that needs to walk it —
+    /// [`MegaHardforkConfig::with_all_activated`],
+    /// [`MegaHardforkConfig::with_all_activated_through`] — derives from this array, so
+    /// introducing a hardfork means adding one entry here rather than editing every place that
+    /// enumerates forks.
+    pub const ALL: [Self; 11] = [
```
Troublor: `ALL` is a real improvement, but the doc comment claims more than the code enforces: "introducing a hardfork means adding one entry here rather than editing every place that enumerates forks." Nothing makes that true. `[Self; 11]` is a plain array literal whose length is only checked against its own contents — add `Rex8` to the `hardfork!` invocation above and omit it here, and it compiles. `with_all_activated()` then silently doesn't activate it. The guard at line 505 iterates a *second* hand-written list to assert membership in the first, so a variant missing from both fails neither.

The broader ask, which I think is worth doing properly rather than patching: **a new spec or a new hardfork should be declared in exactly one place, and every other site should either derive from it or fail to compile.** Right now the exhaustive matches force some sites (`spec_id()`, the instruction table, `EvmTxRuntimeLimits::for_spec`, precompiles) while these stay silent:

- `MegaHardfork::ALL` (here)
- the `hardfork()` newest-first `if/else` chain and the `is_rex_N_active_at_timestamp` predicates — hand-written per fork
- `tests/compute_gas/main.rs::ALL_SPECS`, `evm/spec.rs`'s test `ALL_SPECS`, the spec list in `test_with_all_activated_through_resolves_to_that_spec`
- `mutants/operators/spec-gate/generate.py::ALL_SPECS` / `FROZEN_SPECS`
- `benches/common/mod.rs::SPEC_IDS` and `.github/scripts/bench_compare.py::ROW_ORDER`
- `mega-state-test`'s `SpecName` wildcard arm (now partly guarded by the new test)

Spec and hardfork are genuinely different things and need separate declarations — several hardforks map to one spec (`MiniRex1` → `EQUIVALENCE`, `MiniRex2` → `MINI_REX`), so the mapping has to be an input to the declaration, not inferred from position. Sketch:

- `mega_specs! { EQUIVALENCE => "Equivalence", …, REX6 => "Rex6", REX7 => "Rex7" (unstable) }` emitting the `MegaSpecId` enum, `name::*`, `FromStr`/`Display`, `MegaSpecId::ALL`, and `MegaSpecId::LATEST_FROZEN` alongside `default()`.
- `mega_hardforks! { MiniRex => MINI_REX, MiniRex1 => EQUIVALENCE, …, Rex7 => REX7 }` emitting the enum, `spec_id()`, `ALL`, the per-fork `is_*_active_at_timestamp` predicates, and the newest-first `hardfork()` chain.

Everything in-crate then consumes `::ALL` at runtime instead of re-listing; the two non-Rust consumers (`generate.py`, `bench_compare.py`) read a generated manifest instead of a hand-copied list. That also makes `FROZEN_SPECS = ALL_SPECS[:-1]` derivable and retires the "which spec is unstable" fact currently duplicated across `AGENTS.md`, `generate.py`, and `docs/spec/`.

Probably out of scope for this PR — but if it's deferred, the minimum that makes *this* const's doc comment true is a compile-forced completeness guard:

```rust
/// Position in [`Self::ALL`]. Exhaustive by construction: a new variant
/// fails to compile here, which is what keeps `ALL` complete.
const fn ladder_index(self) -> usize {
    match self { Self::MiniRex => 0, /* … */ Self::Rex7 => 10 }
}
```

with a test asserting `ALL[i].ladder_index() == i`. That subsumes the hand-written list at line 505 and makes the comment at line 527 (which claims the spec loop below it fails on a new spec — it doesn't) true as written.

Troublor (changes requested):

crates/mega-evm/src/block/hardfork.rs
RealiCZ: Agreed, and the model fix is its own PR: **#362**. It splits the executing spec from a monotone activated-spec floor, makes the spec-introducing `is_<fork>_active_at_timestamp` predicates projections of that floor, adds `validate_schedule`, and makes `without` private. It also carries the spec-side half of your `MegaHardfork::ALL` comment — `MegaSpecId::ALL` with an exhaustive-match guard, since `MegaSpecId` has no macro to generate it from. #362 merges first and this PR rebases onto it, so for the spec-gates code that PR is authoritative; this one stays on the documentation.

Two notes on the suggestions. Your validate as literally stated would reject MegaETH's own testnet schedule — `MiniRex1`/`MiniRex2` are `Never` after `MiniRex` at `Timestamp(0)`, and `Never` sorts as infinitely late — so it is scoped to spec-introducing forks. And privatizing `without` narrows an idiom, not a capability: `with(fork, ForkCondition::Never)` is equivalent and has to stay public for that same testnet schedule.

The call sites are all converted, including six you did not list (`system/sequencer_registry.rs`, `tests/rex5/system_tx_replay.rs`) and three Rex6-named helpers that the new default silently turned into Rex7 chains.

On the bench roster: fixed here in `43df9b7`. Of your two options for the invisibility I took the second — a `ROW_ORDER` entry that no bench emitted is now named in the table output — since pattern-matching surfaces an unexpected row, but the defect was a missing one, which has no segment to match.

Troublor (approved):

PR #362, opened 2026-07-31
refactor(block): derive hardfork gating from single-source spec resolution

RealiCZ:
## Summary

- Make the fork→spec mapping 1:1 and monotone: every hardfork schedules its own spec rung, and rollbacks are expressed by **alias specs** — `MINI_REX_1` (behavior: `EQUIVALENCE`) and `MINI_REX_2` (behavior: `MINI_REX`) — rungs of their own whose `behavior()` projects to an earlier spec. The resolved `spec_id` only ever climbs; mainnet's rollback window resolves to `MINI_REX_1` and executes `EQUIVALENCE` semantics through the projection.
- One resolved spec, two projections: `spec.is_enabled(X)` compares **behavior** (both sides project, so alias windows roll execution semantics back — the ~30 existing behavior gates in `limit/`, `intercept`, etc. needed zero changes), and `spec.reaches(X)` compares **position** (one-way chain setup: predeploys, bytecode versions, pre-block EIP-2935/4788 rules — never rolled back). The EVM dispatch sites (instruction tables, runtime limits, precompiles, block limits) give each alias a first-class arm grouped with its target (`EQUIVALENCE | MINI_REX_1 => ...`), and generic reconciliation tests (driven off `MegaSpecId::ALL` × `is_alias()`) assert each site's grouping agrees with `behavior()` — stating the mapping twice with an automatic cross-check, so a mis-grouped arm fails CI instead of shipping a hybrid spec.
- Three structural invariants are pinned at compile time by const assertions whose checkers are themselves tested against malformed inputs: `MegaSpecId::ALL` is a gapless ladder prefix, the `behavior()` projection is flat (no alias chains or cycles, no upward or duplicate-keyed targets), and the fork→spec map climbs the ladder strictly (strictness doubles as the 1:1 proof, and the assertion message carries the remedy: express a rollback as a new alias rung, never by reusing an earlier spec). `crates/mega-evm/src/block/AGENTS.md` gains an invariant map indexing every guard by the layer it fires at (compile / config-load / block-execution / test / CI).
- The monotone ladder collapses the machinery a reversible spec needed: `max_activated_spec_id` is gone — the maximum over activated forks IS the latest activated fork's spec, so `spec_id` is the one resolution and one-way intent lives at the comparison (`reaches`). Pre-block setup reads the executing spec from the EVM cfg — the same source `resolve_system_address` reads — so the executor is coherent by construction rather than by an assert, and the orphan-patch validation rule folds into the general skipped-rung check (an alias scheduled without its base *is* a skipped rung, since the alias's rung sits above its base's).
- The `is_<fork>_active_at_timestamp` predicates are position projections for behavior-introducing forks (kona-style cascading: a schedule that omits a predecessor still reports it active, keeping downstream gates additive) and raw event queries for the two alias forks, whose occurrence is not recoverable from ladder position — testnet never scheduled them while its ladder climbs past their rungs.
- Add `MegaHardforks::validate_schedule` (op-node-style `Check()`): a published schedule must climb the ladder without skipped rungs (alias forks may be omitted, but an alias without its base is itself a skipped rung), in activation order, by timestamp, with required per-fork params attached (`Rex5` → `SequencerRegistryConfig`, `Rex6` → `SequencerRegistryRex6Config`); `hardfork_schedule` debug-asserts it and node startup can call it to fail fast while block execution stays tolerant of malformed schedules. A pairing test pins that the load-time params roster and the pre-block fail-closed checks cover the same requirements on the same config.
- Derive `hardfork()` resolution from `MegaHardfork::VARIANTS` in reverse declaration order and declare `MegaSpecId::ALL` as the single spec enumeration with an exhaustive-match compile guard — porting REX7 (merged from #360) exercised exactly this forcing: the merge would not compile until the new spec was placed on the ladder. Hand-written spec rosters derive from `ALL` where possible (`--bench-spec` accepted-value list and its test).
- Express "a chain running spec N" as `with_all_activated_through(MegaSpecId::N)` (`without` is now private), and pin the unknown-chain fallback to a named rung — advanced to `REX7` in this PR to match the fallback behavior #360 shipped, as an explicit constant rather than an inherited default.
- Seal the deploy layer: each contract's `<name>_spec()` builder and `transact_deploy_sequencer_registry_for` take the resolved spec (plus typed params) instead of a hardfork config, so a per-fork activation gate cannot be reintroduced.

## Replay safety

- Alias specs execute their target's behavior through a total `behavior()` projection shared by every gate — no duplicated instruction tables, no second frozen surface. The mainnet `MiniRex1` window is covered end-to-end: a full block executes under `CfgEnv.spec = MINI_REX_1`, produces `EQUIVALENCE` semantics, and keeps emitting the MiniRex predeploys' read-only witness entries.
- On the canonical mainnet/testnet/fallback schedules, position projections coincide with raw activation events for every behavior-introducing fork, pinned by parity tests sweeping every activation timestamp ±1.
- mega-reth contains zero direct references to `MegaSpecId` variants (no exhaustive matches, no numeric ordinal use, no numeric serialization) — the ladder renumbering and the alias value flowing through `CfgEnv.spec` are transparent to it; predicates keep their semantics through the trait defaults.

## Performance

- CodSpeed reports zero instruction-count change across all pre-existing benchmarks — the alias projection is const-foldable and the per-block resolution is a single reverse scan (the separate floor computation no longer exists); the new `hardfork_resolution` group pins absolute costs.
- Downstream components query predicates once per block, not per transaction (mega-reth's txpool caches results per head into atomics; state-sync resolves per block).
- The block benchmarks now pair each spec row with its coherent schedule (`hardforks_for(spec)`), so a row measures one complete per-spec world; the `rex4` rows previously ran Rex5-scheduled pre-block setup and their CodSpeed baselines shift accordingly.

## API changes

- Breaking: `MegaHardforkConfig::without` is no longer public (use `with_all_activated_through`, or `with(fork, ForkCondition::Never)` for a deliberate gap).
- Behavioral (release-note): the `is_<fork>_active_at_timestamp` predicates for behavior-introducing forks are now position projections rather than raw per-fork event queries — identical on the canonical schedules (pinned by ±1-sweep parity tests), but on non-canonical configs a scheduled fork now implies its predecessors' answers; `mega_fork_activation` remains the raw event query.
- Behavioral (tooling): `mega-evme replay --override.spec` now applies the override to pre-block setup as well — setup and execution move to the override spec together (a coherent what-if, matching `run`/`tx`) instead of setup following the chain schedule while execution follows the override; replay without an override is unchanged. Block-level limits still derive from the chain schedule under an override; aligning them is queued as a mega-evme follow-up.
- Behavioral: `MegaSpecId` gains `MINI_REX_1`/`MINI_REX_2` (and `REX7` via #360); `is_enabled` compares behavior projections (unchanged truth table for all non-alias pairs); `MegaHardfork::spec_id()` is 1:1 and `const`. `"MiniRex1"`/`"MiniRex2"` become valid spec names in `FromStr`/`--override.spec`.
- New: `MegaSpecId::{ALL, behavior, is_alias, reaches}`, `MegaHardforks::validate_schedule`, `ScheduleError`, `MegaHardforkConfig::with_all_activated_through`.

## Testing

- End-to-end tests for partial-ladder pre-block setup, the mainnet rollback window under the alias spec (behavior vs. position asserted separately), the pre-REX5 30M pre-block system call budget (killing the diff-scoped spec-gate mutants), and the load-time/pre-block params pairing.
- Unit tests for position/event parity on all canonical schedules, `with_all_activated_through` resolution per rung (driven off `MegaSpecId::ALL`, including alias rungs), the resolved spec equalling the naive maximum over activated forks on every schedule shape, all `validate_schedule` error classes, alias reconciliation at every dispatch site (instruction tables, precompiles, runtime limits, block limits), malformed-input rejection for all three const checkers, and the latest-spec anchor.
- Full workspace suite (1398 tests), clippy (all features), rustfmt, Prettier, cargo sort, and the riscv `no_std` check pass.


crates/mega-evm/src/block/chain.rs
```diff
@@ -81,9 +81,19 @@ pub fn testnet_hardforks() -> MegaHardforkConfig {
...
+/// must therefore not move them: a devnet that produced history under this fallback
+/// would otherwise replay differently — through [`hardfork_schedule`], which is what
+/// `mega-evme replay` resolves an unknown chain ID with — against the same binary
+/// that produced it. Advancing this rung is a deliberate edit made when a spec is
+/// sealed, and forgetting it leaves these chains where they are, which is the safe
+/// direction.
 pub fn all_activated_hardforks() -> MegaHardforkConfig {
     MegaHardforkConfig::new()
-        .with_all_activated()
+        .with_all_activated_through(MegaSpecId::REX6)
```
flyq: **[Major — coordination, not a code defect]** This pin makes the opposite policy decision from open PR #360, and the two PRs overlap on 8 files.

#360 keeps the unknown-chain fallback **following the latest spec** — it builds this config with `with_all_activated()` (through `MegaSpecId::default()` = REX7 there), asserts `hardfork_schedule(1).spec_id(0) == MegaSpecId::REX7`, and updates `replay.md` to "currently equivalent to `Rex7`". This PR pins the fallback at `REX6` and the doc comment above argues precisely why following the latest is a replay hazard. Both cannot merge as-is; whichever lands second must resolve the *policy*, not just the text.

The overlap also includes duplicate infrastructure: #360 introduces a hand-rolled `MegaHardfork::ALL` constant (plus a test asserting it stays in sync) and builds its `with_all_activated_through` on it; this PR builds the same builder on the macro-generated `MegaHardfork::VARIANTS`, which is strictly better — the `hardfork!` macro emits `VARIANTS` from the same variant list that declares the enum, so there is no second list to forget. On the merits I'd suggest: this PR's pin + `VARIANTS` win, and #360 drops its `ALL` constant, its sync test, and its fallback/`replay.md`/`test_schedule_dispatch_by_chain_id` changes on rebase.

For the record, #360 landing after this one needs no `validate_schedule` change (the params checks key to Rex5/Rex6 only), and `with_all_activated()` = `through(default)` picks up Rex7 automatically — consistent with #360's all-activated tests.

Overlapping files: `chain.rs`, `hardfork.rs`, `sequencer_registry.rs`, `rex4/deployment.rs`, `rex5/system_tx_replay.rs`, `AGENTS.md`, `docs/mega-evme/commands/replay.md`, `docs/spec/hardfork-spec.md`.

crates/mega-evm/benches/block_bench.rs
```diff
@@ -446,12 +446,42 @@ fn bench_rex5_pre_block(c: &mut Criterion) {
     group.finish();
 }
 
+/// Benchmark hardfork-config resolution on the real mainnet schedule.
+///
+/// The floor-projected predicates (`is_rex_5_active_at_timestamp`) and
+/// `max_activated_spec_id` are what downstream node components call once per block; `spec_id`
+/// is the executor's per-block resolution; `validate_schedule` is a once-per-config-load cost.
+/// The floor's descending early-exit scan keeps the per-query cost at one or two activation
```
flyq: **[Minor]** This comment (and the PR description's Performance section) describes an implementation that does not exist.

`max_activated_spec_id` (`crates/mega-evm/src/block/hardfork.rs:179-186`) is a plain full scan — `.iter().filter(...).map(...).max()` over all of `VARIANTS` — with no descending order and no early exit. The description's claim that "a reference test pins the early-exit scan against the naive maximum over all activated forks across every schedule shape" also has no counterpart in the diff (no such test exists anywhere in the repo).

The measured numbers confirm it: on my machine the projected predicate costs ~364 ns ≈ the floor's ~366 ns — the full-scan shape. A genuine descending early-exit (mainnet: Rex6 inactive → Rex5 active, 2 lookups) would land near `spec_id`'s ~64 ns, not 6× above it.

Not a correctness or real performance problem — downstream calls this once per block — but a comment describing nonexistent code will misdirect the next perf investigation. Either:

- implement the early exit (descending scan, return at the first activated *spec-introducing* fork; fold patch-fork specs into a running max so a `MiniRex2`-without-`MiniRex` partial ladder still resolves correctly) plus the claimed reference test, or
- fix this comment and the PR description to say the naive scan over 10 forks is the deliberate choice.

crates/mega-evm/tests/mutation/block.rs
```diff
@@ -211,36 +211,63 @@ fn staged_config() -> MegaHardforkConfig {
         .with(MegaHardfork::Rex3, ForkCondition::Timestamp(700))
         .with(MegaHardfork::Rex4, ForkCondition::Timestamp(800))
         .with(MegaHardfork::Rex5, ForkCondition::Timestamp(900))
+        .with(MegaHardfork::Rex6, ForkCondition::Timestamp(1000))
 }
 
-/// Each `is_*_active_at_timestamp` returns `true` at/after activation and `false` before. This
-/// kills the `-> false` mutants on the MiniRex1/MiniRex2/Rex1/Rex3 predicates (and pins the rest).
+/// Each `is_*_active_at_timestamp` returns `true` at/after activation and `false` before, on a
```
flyq, line 218: **[Minor]** Hand-edit of a machine-maintained mutation test drops its mutant linkage.

The rework itself is sound and necessary — the predicate bodies changed shape, and the two-sided probes are exactly what kills the new rung-shift mutants (the doc comment's analysis of which side catches which shift is good). But the per-mutant location references (`MiniRex1 (line 144)`, `(line 149)`, `(line 159)`, `(line 169)`) were deleted rather than updated, and `tests/mutation/main.rs` declares these tests machine-generated/machine-maintained ("do not rename, reword, or refactor them by hand"), while REVIEW.md requires the `file:line` mutation-location references to be kept up to date when the referenced source moves — a reference that no longer points at its mutant defeats the linkage.

Suggestion: restore location references pointing at the new predicate bodies in `hardfork.rs` (or regenerate via the `improve-mutation-score` flow) so each assertion pair stays keyed to a concrete mutant.

crates/mega-evm/src/block/hardfork.rs
```diff
@@ -108,88 +127,268 @@ pub trait MegaHardforks: OpHardforks {
...
+    ///   strictly before an earlier-declared one.
+    /// - No skipped rungs: a spec-introducing fork may be unscheduled only if no scheduled fork
+    ///   maps to an equal-or-higher spec. Patch forks are exempt — testnet legitimately never
+    ///   schedules `MiniRex1`/`MiniRex2`.
+    /// - Per-fork parameters required by a scheduled fork are present (`Rex5` requires
+    ///   [`SequencerRegistryConfig`](crate::SequencerRegistryConfig), `Rex6` requires
+    ///   [`SequencerRegistryRex6Config`](crate::SequencerRegistryRex6Config)). This surfaces a
+    ///   missing config at load time rather than at the first block of the fork. A new
+    ///   [`HardforkParams`] type must be registered here.
+    fn validate_schedule(&self) -> Result<(), ScheduleError> {
```
flyq, line 298: **[Minor]** `validate_schedule` accepts a patch fork scheduled without its base.

A schedule containing only `MiniRex1 @ Timestamp(0)` (no `MiniRex`) validates: `MiniRex1`'s spec is `EQUIVALENCE`, so it never triggers the skipped-rung check for `MiniRex` (that check fires only when a *scheduled* fork maps to an equal-or-higher spec, and `EQUIVALENCE < MINI_REX`). Scheduling the rollback of a fork that never happened is a configuration mistake of exactly the class this function exists to reject — it executes harmlessly (spec stays `EQUIVALENCE`), but as a published schedule it is nonsense.

Cheap tightening: a scheduled patch fork requires the fork introducing the spec it patches (equivalently, its immediate predecessor in declaration order) to be scheduled. `MiniRex2`-without-`MiniRex` is already caught by the existing check (equal spec), so this only closes the rollback-fork case.

flyq (commented):
## Review: derive hardfork gating from single-source spec resolution

**Verdict:** the design and implementation are strong — I verified the central claims directly on the branch and they hold. Before merge: one **coordination-level** issue with open PR #360 (inline comment on `chain.rs:96`) needs an explicit decision, plus two Minors (a nonexistent early-exit implementation claimed by the PR description and a bench comment; a hand-edited mutation test that dropped its mutant linkage).

### Verified

- **Floor/event parity** — `test_floor_matches_per_fork_activation_for_spec_introducing_forks` sweeps every canonical schedule's activation timestamps ±1 and pins floor == raw event for every spec-introducing fork; behavior differs only on partial ladders, which no canonical chain has.
- **Mainnet `MiniRex1` rollback window covered end-to-end** — the executing spec rolls back to `EQUIVALENCE` while the witness hook still receives the Oracle predeploys' accounts, asserted through a real `OnStateHook` recorder rather than inferred.
- **`hardfork()` rewrite is equivalent** — `VARIANTS.rev().find(...)` preserves the old if-else ladder's order and equal-timestamp tie-break (later-declared wins), and reads raw activation events, correctly bypassing the predicates' semantic change.
- **`VARIANTS` is macro-generated** (alloy-hardforks 0.2.13 `hardfork!`, `src/hardfork/macros.rs:15`) — no second hand-written ladder to keep in sync.
- **`validate_schedule`** — all four error classes tested, including the `with_all_activated_through(EQUIVALENCE)` patch-fork edge and the skipped middle rung (0f92b95); the `hardfork_schedule` debug-assert passes on all three canonical schedules.
- **No leftover callers** — no consensus path in `src/` or `bin/` still consumes the projected predicates; the deploy layer is sealed (spec builders take a resolved `MegaSpecId`, so a per-fork activation gate cannot be reintroduced).
- **Partial-ladder e2e tests are high quality** — the fail-closed test asserts the *specific* error string and documents why matching the variant alone would pass under either gating style; exactly the "would this test still pass if the behavior regressed" discipline REVIEW.md asks for.
- **Run locally** — `mega-evm --lib` (276), `block_executor` (38, incl. the 4 new partial-ladder tests), `mutation` (47), `rex4` (216), `rex5` (209): all green. riscv `no_std` check passes (`ScheduleError`'s `derive_more` derives are no_std-clean). `cargo fmt` clean; clippy shows only the pre-existing MSRV warning. The new bench group runs; measured numbers match the description's.
- Labels are accurate: `api:breaking` (signature + `without` privatization), `spec:stable` (zero behavior change on canonical schedules, pinned by the parity tests).

### Not attachable inline

- **[Nit]** The PR's Testing section omits the riscv `no_std` check (`cargo check -p mega-evm --target riscv64imac-unknown-none-elf --no-default-features`) — a standard item for core-crate changes. I ran it; it passes. Worth adding to the test plan for the record.
- **[Nit]** `max_activated_spec_id` takes `BlockTimestamp` while every `is_*_active_at_timestamp` predicate takes `u64` (the same underlying alias) — worth aligning on one spelling.
- The docs updates check out against the implementation: the new "Executing Spec vs. Highest Spec Reached" section in `hardfork-spec.md` matches the code's semantics, and the `replay.md` / three `AGENTS.md` rule updates all correspond to real code.

Specific change suggestions are attached as inline comments.

crates/mega-evm/src/block/chain.rs
RealiCZ: Good catch on the overlap, but there is no competing policy to resolve — this is the intended sequence. This PR is the spec-gates mechanism work and lands first; #360's changes on this surface predate it and were always going to be superseded on rebase, with #360 keeping its actual scope: sealing REX6 and opening REX7, primarily documentation. Concretely, on rebase #360 drops its hand-rolled `MegaHardfork::ALL`, the sync test for it, and its fallback/`replay.md`/`test_schedule_dispatch_by_chain_id` changes — the spec-gates code here is authoritative. The pin stays at `REX6`, which after #360 is the newest sealed spec; advancing it to `REX7` would be a separate deliberate edit, updating the `RUNG` constant in `test_unknown_chain_fallback_pins_its_rung` in the same change. #360's REX7 addition will also hit the new `ladder_index` exhaustive match and `MegaSpecId::ALL`, so the new spec joins the enumeration under compile-time forcing rather than by convention.

crates/mega-evm/src/block/hardfork.rs
```diff
@@ -39,6 +39,27 @@ hardfork! {
 }
 
 impl MegaHardfork {
+    /// Whether this fork introduces a new spec — its spec is strictly higher than every
+    /// earlier-declared fork's. Patch forks (`MiniRex1`, `MiniRex2`) do not.
+    pub(crate) fn introduces_spec(self) -> bool {
+        let declared = self.declaration_index();
+        Self::VARIANTS[..declared].iter().all(|fork| fork.spec_id() < self.spec_id())
+    }
```
Troublor: Suggest declaring skippability rather than deriving it from the spec map.

`introduces_spec()` is being used to answer "may a published schedule omit this fork?" — that's what gates the `SkippedRung` exemption, `base_fork()`, and the `is_*_active_at_timestamp` split. But the property it actually needs is *why the fork exists*: `MiniRex1` is omittable because it rolled back a defect only mainnet hit, and `MiniRex2` because it undid that rollback. Testnet never hit the defect, so it schedules neither. The spec map doesn't carry that — it selects the right two forks today by coincidence of the current schedule.

Geth treats these as independent axes and declares both. Its skippable forks carry a hand-written `optional: true` (`daoForkBlock`, the Glaciers, `mergeNetsplitBlock` — all situational fixes only some networks needed), and the two axes deliberately don't line up: the Glaciers *do* change consensus rules and are optional, while Petersburg — which removes an EIP Constantinople added, the closest analogue to `MiniRex1` — is **not** optional and gets a hand-written special case in `IsPetersburg`. You can't derive one from the other.

Concretely, I'd suggest replacing this with an exhaustive match, so a new fork is a compile error until someone decides — the same forcing function `ladder_index` already uses:

```rust
impl MegaHardfork {
    /// Whether a published schedule may omit this fork.
    ///
    /// `MiniRex1` rolled back a `MiniRex` defect on mainnet and `MiniRex2` restored it
    /// afterwards; testnet never hit the defect and schedules neither. Every other fork
    /// is required — see `validate_schedule`.
    pub const fn optional(self) -> bool {
        match self {
            Self::MiniRex1 | Self::MiniRex2 => true,
            Self::MiniRex | Self::Rex | Self::Rex1 | Self::Rex2
            | Self::Rex3 | Self::Rex4 | Self::Rex5 | Self::Rex6 => false,
        }
    }
}
```

Three things fall out, which is what makes me think this is the right cut rather than just a rename:

1. **`introduces_spec` currently misclassifies a supported shape.** `AGENTS.md` states multiple hardforks may map to one spec. The strict `<` means a future `Rex7b` sharing `REX7` with `Rex7a` is silently treated as a patch fork — inheriting the skip exemption and an `OrphanPatch` dependency it shouldn't have. `test_floor_matches_per_fork_activation_for_spec_introducing_forks` skips non-introducing forks, so nothing catches it.

2. **`OrphanPatch` disappears.** If the no-skip rule is stated in *declaration* order rather than spec order, "`MiniRex1` scheduled but `MiniRex` not" is already a skipped-fork violation. That also fixes a misdiagnosis in the current rule: a schedule with only `MiniRex2` reports `SkippedRung { missing: MiniRex, scheduled: MiniRex2 }`, and `MiniRex2` is not "scheduled above" `MiniRex` — it's a patch of the same rung, so `OrphanPatch` is unreachable for the shape it was added for. The validator then needs no `spec_id` at all.

3. **The term "patch fork" can go.** It currently collides with the vocabulary twelve lines up in this same file — the enum documents `Rex1`–`Rex6` as *"first patch to Rex"*, *"second patch to Rex"*, which under the new definition are exactly the forks that are **not** patch forks. It's also load-bearing in `docs/spec/hardfork-spec.md:38`, where a MUST/MAY rule references "patch hardforks" three times without defining the term anywhere in the document — a third-party implementer can't implement that rule, and guessing from the enum's doc comments gives the wrong set.

`with_all_activated_through` doesn't need `base_fork` either — a declaration-order prefix is simpler and is what it actually wants: walk `VARIANTS`, register until the first fork whose spec isn't enabled, drop the rest.

Not blocking on its own, but I'd want it in this PR rather than a follow-up: once mega-reth calls `validate_schedule` at genesis load, `ScheduleError` becomes a cross-repo contract, and reshaping it afterwards is a breaking change that also propagates this vocabulary into a second repo.

RealiCZ:
@codex review

crates/mega-evm/src/evm/spec.rs
```diff
@@ -106,22 +142,91 @@ impl MegaSpecId {
...
+    /// Derived from [`behavior`](Self::behavior), so a future alias needs only its projection
+    /// arm; there is no second list to extend.
+    pub const fn is_alias(self) -> bool {
+        self.behavior() as u8 != self as u8
+    }
+
+    /// Returns `true` if `other`'s BEHAVIOR is enabled under `self` — the gate for execution
+    /// semantics. Both sides project through [`behavior`](Self::behavior) first, so an alias
+    /// spec enables exactly what its behavior target enables (`MINI_REX_1` does NOT enable
+    /// `MINI_REX`).
```
flyq, line 169: **[major — pre-merge ask]** This double projection is my best explanation for the red CodSpeed check: `is_enabled` now runs `behavior()` on both sides — a 12-variant match — and the per-frame paths call it heavily (`evm/execution.rs` alone has 31 call sites; interceptor dispatch and the limit sub-tracker frame hooks all gate through it). The two regressed benches (`subcall_1000_transfer_1wei::mini_rex` −5.33%, `subcall_1000_nested::mini_rex` −5.17%) are exactly the 1000-frame workloads where a few extra instructions per gate multiply, and `mini_rex` is the row where that overhead is largest *relative* to its light per-frame work (rex4+ rows absorb the same absolute cost below CodSpeed's threshold). CodSpeed's per-bench flamegraph diff should confirm or kill this hypothesis in minutes. If confirmed: either cache a pre-projected behavior where hot (e.g. store the projected `behavior` alongside `spec` in the interpreter-adjacent structs), or acknowledge on CodSpeed with a rationale (synthetic frame-density microbench; block-level benches unaffected). Either way the red check should not merge unexplained.

crates/mega-evm/src/evm/spec.rs
```diff
@@ -40,6 +42,12 @@ pub enum MegaSpecId {
     EQUIVALENCE,
     /// The EVM version for the *Mini-Rex* hardfork of `MegaETH`.
     MINI_REX,
+    /// Alias spec for the `MiniRex1` hardfork: scheduled as its own rung, executes
```
flyq, line 45: **[major — pre-merge ask]** Inserting the alias rungs here renumbers every discriminant from `REX` (2→4) through `REX7` (9→11). In-repo this is safe — I checked: serde on this enum is name-based (JSON), and there is no `as u8` persistence, no bincode/postcard usage, and no numeric serialization anywhere in the workspace. The remaining exposure is downstream: any variant-*index* codec (bincode-style) of `MegaSpecId` — or of a `CfgEnv`/`EvmEnv` containing it — would silently misread old data (index 2 used to mean `REX`, now means `MINI_REX_1`). The PR body says mega-reth was audited for exactly this; please attach the grep evidence or link the companion mega-reth PR/CI run here — I could not verify it independently. Cheap belt-and-braces: a `test_ladder_positions_are_pinned` asserting the exact `as u8` value of every variant, so any future renumbering becomes a loud diff in this repo, where the review attention is.

crates/mega-evm/src/block/hardfork.rs
```diff
@@ -111,95 +134,252 @@ pub trait MegaHardforks: OpHardforks {
...
+    /// [`MegaHardfork::VARIANTS`] so a newly declared fork joins resolution without a second
+    /// hand-written ladder here.
+    ///
+    /// Resolution reads raw activation events
+    /// ([`mega_fork_activation`](Self::mega_fork_activation)); with the 1:1 ascending fork->spec
+    /// map, the resolved spec is monotone in time on any ordered schedule.
+    fn hardfork(&self, timestamp: BlockTimestamp) -> Option<MegaHardfork> {
+        MegaHardfork::VARIANTS
+            .iter()
+            .rev()
```
flyq, line 178: **[minor]** The rev-scan (and the skipped-rung `find` in `validate_schedule`) are correct only while `MegaHardfork::VARIANTS` declaration order maps to strictly ascending `spec_id()` — that invariant is what makes "latest-declared activated fork" ≡ "max activated spec". Today it is guaranteed indirectly (`test_mega_hardfork_spec_ids_match_expected_specs` and `test_floor_early_exit_matches_naive_reference` would both fail), but a one-liner asserting `VARIANTS` maps to strictly increasing `spec_id()` would fail *at* the mistake (a misplaced future variant) instead of two tests away.

crates/mega-evm/src/block/hardfork.rs
```diff
@@ -300,22 +480,28 @@ impl MegaHardforkConfig {
...
+    /// and it would silently drift again the next time a spec is introduced. On top of the wrong
+    /// resolved spec, the leftover later forks also keep the scheduled spec high, so every
+    /// pre-block setup gate below them stays open.
     ///
-    /// The result is a function of `spec` alone, not of what the config held before: a later fork
-    /// already registered is removed rather than left in place, so the resolved spec does not
-    /// depend on builder call order.
+    /// The *activations* in the result are a function of `spec` alone, not of what the config
+    /// held before: a later fork already registered is removed rather than left in place, so the
+    /// resolved spec does not depend on builder call order. Per-fork params are the exception:
```
flyq, line 509: **[nit]** Honest doc — but the params-drop asymmetry is the one stateful surprise left in the builder: params on a removed fork vanish and do not come back on a later climb. No current caller hits it (the canonical schedules attach params after the final ladder shape). A `debug_assert!` (or a log line) when a dropped entry carries params would catch surprising builder orders before they become a puzzling test failure.

flyq (commented):
## Review summary

**Verdict: approve with two pre-merge asks.** The monotone 1:1 fork→spec ladder with alias specs and the behavior/position dual projection is carried through consistently. What I verified before signing off: `is_enabled`'s truth table is unchanged for every pre-existing spec pair (both sides project through `behavior()`, identity for non-alias specs); the alias value flowing through `CfgEnv.spec` is safe in-repo (no raw ordinal comparisons anywhere except two pre-existing, position-correct `>=` in `bin/mega-evme/src/common/state.rs:569-570` — worth migrating to `reaches` in a follow-up for grep-ability); all three EVM dispatch sites (instruction table, runtime limits, precompiles) route through `behavior()` with unreachable alias arms; the deploy layer gates on `reaches`, so the mainnet MiniRex1 window keeps its predeploys and witness entries (pinned end-to-end in `tests/block_executor/partial_ladder.rs`); `validate_schedule` covers all four error classes including the elegant alias-without-base collapse into the skipped-rung rule; the REX7 fallback pin has a drift test in both directions; and mutation testing is 100% on the diff.

Two asks before merge, both as inline comments below: the unexplained CodSpeed regression on the two `mini_rex` subcall benches, and re-verifying the mega-reth serialization claim against current HEAD (I could not verify it independently).

One release-note bullet worth adding: the behavior-fork predicates (`is_mini_rex_active_at_timestamp` … `is_rex_7_active_at_timestamp`) changed from raw per-fork event queries to floor projections. On the canonical schedules this is a no-op (pinned by the ±1-sweep parity test) and it is the right gating semantics — but downstream callers using them as "did this fork's activation event occur" on non-canonical configs now get cascaded answers; `mega_fork_activation` is the raw query. Deserves an explicit line in whatever changelog downstream consumes, beyond the `api:breaking` label.

Nit on the PR body: `resolve_system_address`'s signature did not actually change (it already took `(hardforks, spec, db)` on main — only the internal Rex6 gate moved from a timestamp predicate to `spec.reaches`), so "takes a single spec instead of an exec/setup pair" slightly oversells it.

Merge-order note: landing this before #365 keeps the cheaper rebase on #365's side — the cross-merge conflicts only on `executor.rs` + `instructions.rs`, and #365 does not touch `spec.rs`.

RealiCZ:
@flyq 

Thanks — all items handled; the first one was resolved by events between your review base and the current tip. Point by point:

**CodSpeed (major).** Resolved, and the double-projection hypothesis is refuted by the timeline: the two flagged benches went green at `4c580eb` — a push that touched no per-frame path (`is_enabled` and `behavior()` are byte-identical before/after; the commit only reshaped the three construction-time dispatch matches). A projection cost cannot explain a regression that disappears while the projection stays. A local wall-clock A/B (main@`396b0ca` baseline vs HEAD, same machine, criterion saved-baseline) also failed to reproduce the flagged pair (`transfer_1wei/mini_rex` +2.4%, p=0.08; `nested/mini_rex` −0.9%, p=0.31) while showing ±20–30% wide-CI swings on untouched variants (`transfer_1wei/equivalence`, `nested/rex4`) — instruction-count layout sensitivity, not real cost. The check is green on the current tip; nothing left to acknowledge.

**Discriminant renumbering (major).** Both asks done:

- `test_ladder_positions_are_pinned` (evm/spec.rs) pins every variant's exact `as u8` value, so any future renumbering is a loud diff in this repo.
- mega-reth evidence, against `d6dc8236c` (develop):
  - `grep -rn "MegaSpecId::[A-Z]" --include="*.rs" crates/ bin/` → **0 matches** (no variant references).
  - All `MegaSpecId` mentions: 6 lines in 4 files, every one in type position (imports and `EvmEnv<MegaSpecId>` generics — ephemeral runtime values; no persisted struct contains the type).
  - No numeric casts (`as u8`/`to_*_bytes`) involving the type.
  - `cargo tree -i bincode` → **empty**: bincode sits in Cargo.lock only behind reth's optional `serde-bincode-compat` feature on `reth-{optimism,ethereum}-primitives` (block/tx primitives that do not contain `MegaSpecId`); the feature is not in the active graph and mega-reth's own code never invokes bincode.

**VARIANTS ↔ ladder order (minor).** Added `test_variants_declaration_order_climbs_the_ladder`: `windows(2)` over `VARIANTS` asserting strictly ascending `spec_id()`, so a misplaced future variant fails at the mistake.

**Params-drop `debug_assert!` (nit).** Declined, with a concrete counterexample: truncating a params-carrying config — e.g. `mainnet_hardforks().with_all_activated_through(MegaSpecId::REX4)` to exercise a pre-Rex5 rung — is exactly the documented use of the builder, and dropping the Rex5/Rex6 params along with their forks is the desired outcome there. A `debug_assert!` would panic on defined behavior. The docstring states the asymmetry and the re-attach path (`with_params`).

**One correction to the summary:** "all three EVM dispatch sites route through `behavior()` with unreachable alias arms" describes the tip you first reviewed. Since `4c580eb` the sites give each alias a first-class arm grouped with its target (`EQUIVALENCE | MINI_REX_1 => ...`), and a per-site reconciliation test (driven off `MegaSpecId::ALL` × `is_alias()`) asserts the grouping agrees with `behavior()` — the mapping is stated twice with an automatic cross-check, and the mis-grouping mutants are verified killed. `is_enabled` itself is unchanged.

**Also done:** the two `>=` in `bin/mega-evme/src/common/state.rs` are migrated to `reaches` in this PR (not deferred), and the predicate-semantics release note is added to the PR body ("Behavioral (release-note)" bullet, pointing at `mega_fork_activation` as the raw event query).

flyq (approved):

crates/mega-evm/src/evm/spec.rs
```diff
@@ -106,22 +142,91 @@ impl MegaSpecId {
...
-    /// or below it, so a lower-or-equal version is always enabled under a higher one.
+    /// The behavior this spec executes: alias specs project to the spec whose behavior they
+    /// reuse; every other spec is its own behavior.
+    pub const fn behavior(self) -> Self {
+        match self {
+            Self::MINI_REX_1 => Self::EQUIVALENCE,
+            Self::MINI_REX_2 => Self::MINI_REX,
+            other => other,
+        }
+    }
```
Troublor, line 157: `behavior()` is now the most load-bearing table in this design — every `is_enabled` routes through it twice, `is_alias`/`introduces_behavior` derive from it, and the three alias reconciliation tests in `instructions.rs`, `precompiles.rs`, and `limit.rs` all check *against* it. But nothing pins any property of the table itself: `spec.rs` has tests for names, ladder positions, order, and `is_ladder_prefix`, and none for this.

Two shapes are currently expressible and would be caught by nothing:

- **A chain** — a future `MINI_REX_3 => MINI_REX_1`. `behavior()` returns `MINI_REX_1`, an alias, so `is_enabled` compares against a rung that has no behavior of its own. Silently wrong, not an error.
- **A cycle** — `A => B, B => A`. Single-step lookup, so it terminates and just returns the wrong answer.

Could resolution be made transitive, with cycle detection? i.e. follow the projection to a fixed point, and reject (or fail to compile) if the chain doesn't terminate. That makes `MINI_REX_3 => MINI_REX_1 => EQUIVALENCE` resolve correctly rather than half-way, and makes a cycle a loud failure instead of a silent wrong answer.

Worth noting the tension so it's a deliberate choice either way: `behavior()` is `const fn` and sits on the EVM gate path, so a loop-to-fixed-point turns a table cycle from "wrong answer" into "hang" (at const-eval or at runtime), and detection then needs a bound or a visited set inside `no_std` + `const fn`. So transitive resolution and cycle detection have to land together — the first without the second is worse than today.

If you'd rather keep the projection a single lookup, the same two hazards can be made unrepresentable at compile time, following the `is_ladder_prefix` pattern already in this file:

```rust
/// A behavior target must itself be concrete, and must not sit above the spec projecting
/// onto it.
///
/// Idempotence rules out chains *and* cycles in one property: a chain `A→B→C` gives
/// `behavior(behavior(A)) == C != B`; a cycle `A→B→A` gives `A != B`.
const fn behavior_table_is_flat(list: &[MegaSpecId]) -> bool {
    let mut i = 0;
    while i < list.len() {
        let target = list[i].behavior();
        if target.behavior() as u8 != target as u8 {
            return false; // chain or cycle
        }
        if target as u8 > list[i] as u8 {
            return false; // alias projecting upward
        }
        i += 1;
    }
    true
}

const _: () = assert!(
    behavior_table_is_flat(MegaSpecId::ALL),
    "behavior() targets must be concrete specs at or below the projecting rung"
);
```

The upward check is worth having regardless of which route you take: if an alias projected *up*, a chain resolved to that rung would execute higher semantics while `reaches` reported a lower position, so the setup for those semantics would never have run — `is_enabled` and `reaches` would be describing incompatible worlds. That's currently true by inspection and by nothing else.

Either way, the checker itself should get a test fed malformed tables — as `test_is_ladder_prefix_rejects_malformed_lists` already does for the ladder — since the real table always satisfies the property and a weakened guard would pass silently.

crates/mega-evm/src/block/hardfork.rs
```diff
@@ -413,13 +610,28 @@ mod tests {
...
+        // misplaced variant instead of two derived tests away.
+        for pair in MegaHardfork::VARIANTS.windows(2) {
+            assert!(
+                (pair[0].spec_id() as u8) < (pair[1].spec_id() as u8),
+                "{:?} -> {:?} must climb the spec ladder",
+                pair[0],
+                pair[1],
+            );
+        }
+    }
```
Troublor, line 648: Good that this invariant is pinned at all — but it's guarded more weakly than the ones it underwrites.

The three structural invariants of this design are enforced inconsistently:

| invariant | guard |
| --- | --- |
| `MegaSpecId::ALL` is in ladder order without gaps | `const _: () = assert!(is_ladder_prefix(...))` — compile time |
| every spec has a `ladder_index` | exhaustive match — compile time |
| **fork → spec is strictly ascending** | this `#[test]` — runtime |

The third is the one the others rest on. `max_activated_spec_id` collapsing to `spec_id` is *only* correct because of it: if a future fork mapped to a lower existing rung, the latest activated fork would no longer be the maximum over activated forks, and one-way setup would start retracting inside that window — the exact bug class this PR exists to prevent, re-enterable through the single invariant the compiler doesn't check.

It's also the one most likely to be broken by a well-intentioned edit. Someone adding a rollback fork the old way — `MiniRex3 => MINI_REX`, reusing an existing rung rather than minting a new alias rung — trips this test, but only when it runs, and the message ("must climb the spec ladder") doesn't tell them the remedy.

Suggest promoting it to a const assertion, same shape as `is_ladder_prefix`. It needs only `spec_id` to become `const fn` — the body is already a match over unit variants, so it qualifies as-is; the signature would move from `&self` to `self`:

```rust
const fn variants_climb_the_ladder() -> bool {
    let mut i = 1;
    while i < MegaHardfork::VARIANTS.len() {
        if MegaHardfork::VARIANTS[i - 1].spec_id() as u8 >= MegaHardfork::VARIANTS[i].spec_id() as u8 {
            return false;
        }
        i += 1;
    }
    true
}

const _: () = assert!(
    variants_climb_the_ladder(),
    "each hardfork must map to a strictly higher spec rung than the one before it — \
     express a rollback as a new alias rung, not by reusing an earlier spec"
);
```

Two things that buys beyond earlier failure: the assertion message can carry the *remedy* rather than just the violated property, which is what the next person adding a rollback fork actually needs; and strict `<` pins injectivity at the same time, so the 1:1-ness the docs rely on stops being a separate claim.

Together with the `behavior()` flatness check I left on `spec.rs`, that would put all three load-bearing properties on the compiler: the ladder is ordered, forks climb it strictly, and alias projections are flat and point downward.

crates/mega-evm/src/block/hardfork.rs
```diff
@@ -111,95 +134,252 @@ pub trait MegaHardforks: OpHardforks {
...
+    /// [`spec_id`](Self::spec_id).
+    ///
+    /// Kept as an alias so call sites can state "one-way setup" intent explicitly; pair it with
+    /// [`MegaSpecId::reaches`] (position) rather than `is_enabled` (behavior).
+    fn max_activated_spec_id(&self, timestamp: BlockTimestamp) -> MegaSpecId {
+        // With the 1:1 ascending fork->spec map, the latest activated fork IS the maximum:
+        // the floor coincides with `spec_id` on every schedule. Kept as an alias for the
+        // transition; call sites can migrate to `spec_id` + `reaches`.
+        self.spec_id(timestamp)
+    }
```
Troublor: Suggest deleting this method — the problem is not that it is redundant, it is that "max" names a concept the model no longer has.

Each hardfork maps to exactly one spec, and the map is strictly ascending in declaration order (now pinned by `test_variants_declaration_order_climbs_the_ladder`). Under those two facts there is no maximum to take: the set of activated forks has a *latest* member, and its spec *is* the answer. A maximum is only a meaningful quantity if the map could descend — which is precisely the shape the 1:1 alias-rung redesign eliminated by giving `MiniRex1`/`MiniRex2` their own ascending rungs.

So the name imports the old mental model. A reader who meets `max_activated_spec_id` on a public trait will reasonably infer that activation order and spec order can disagree and that the code is defending against it. That inference is now false, and it is exactly the confusion this redesign set out to end. The body being `self.spec_id(timestamp)` is the symptom; the name is the defect.

Two reasons to do it in this PR rather than later:

- It is new API introduced here, on `MegaHardforks`, which downstream implements. Removing it now costs nothing; removing it after merge is a breaking release, and "already published" is how transitional shims become permanent. The doc says *"Kept as an alias for the transition; call sites can migrate"* — but nothing migrated, so the transition has no end condition.
- The stated justification (letting call sites "state one-way setup intent explicitly") is already served by `reaches`, at the comparison, which is where the intent actually lives:
  ```rust
  spec_id(t).reaches(MegaSpecId::REX5)                 // "the ladder reached REX5"
  max_activated_spec_id(t).reaches(MegaSpecId::REX5)   // states it twice, once via a name that no longer describes a computation
  ```

Removal is mechanical — 17 production call sites, all inside `mega-evm`: the ten predicates below, plus `executor.rs:145`, `deploy.rs:150`, `oracle.rs:49,111`, `control.rs:53`, `limit_control.rs:32`, `keyless_deploy.rs:51`, `sequencer_registry.rs:195`. The "floor" vocabulary in the surrounding test names and in `tests/block_executor/partial_ladder.rs`'s module doc is the same vestige and can go with it.

One test is worth keeping rather than deleting, under a new name. `test_floor_early_exit_matches_naive_reference` asserts that the resolved spec equals the maximum over activated forks across canonical, partial, rollback, block-numbered, and empty schedules. The early-exit scan it was written for is gone, but what it now checks is the ascending invariant itself, verified dynamically across schedule shapes — a good complement to the static check, since one proves declaration order climbs and the other proves the resolved spec really is the maximum on every shape a schedule can take. Something like `test_resolved_spec_is_the_maximum_over_activated_forks` would say what it now does.

crates/mega-evm/src/block/executor.rs
```diff
@@ -56,6 +58,15 @@ pub struct MegaBlockExecutor<H, E, R: OpReceiptBuilder> {
...
     system_caller: SystemCaller<H>,
+    /// The scheduled spec for this block's timestamp, resolved once at construction.
+    ///
+    /// Every pre-block setup gate derives from this single value through `reaches` (position),
+    /// which keeps setup additive by construction and immune to alias windows — an alias rung
+    /// rolls back behavior, not the setup below it.
+    ///
+    /// Cached because the block env is fixed for an executor's lifetime — the constructor
+    /// already reads `block().timestamp` for its hardfork-coherence asserts.
+    setup_spec: MegaSpecId,
```
Troublor: Suggest dropping this field and resolving from one source instead.

Its value is `hardforks.spec_id(block_timestamp)`, and it is read only inside `pre_execution_changes` and its one helper (lines ~196-346). Nothing later in the executor's lifetime touches it, so the caching rationale is thin — the value is resolved once per block either way.

The name is the same residue as `max_activated_spec_id`, one layer up. The field doc now correctly locates the setup-vs-execution distinction in `reaches` ("through `reaches` (position)"), which is the argument against also encoding it in the value's name: if the comparison carries the distinction, the binding does not need to. `scheduled_spec` would at least match the doc's own words, but inlining is cleaner.

The substantive part is that there are now **two sources of the same fact inside one pre-execution chain**:

```rust
let setup_spec = self.setup_spec;                    // :196 — hardforks.spec_id(ts)
...
let spec = self.evm.ctx().mega_spec();               // :695 — cfg_env.spec
resolve_system_address(&self.hardforks, spec, ...)   // :697
```

Both answer "what spec is this block", and they are reconciled only by the coherence assert in `new()`:

```rust
#[cfg(not(any(test, feature = "test-utils")))]
assert_eq!(evm.spec_id(), hardforks.spec_id(block_timestamp), ...);
```

which is compiled **out** under `test` and `test-utils`. `bin/mega-evme/Cargo.toml` enables `test-utils` unconditionally, so the two can diverge there: under `replay --override.spec rex5` on a pre-Rex2 block, pre-block setup follows the *schedule* and installs the v1.0.0 Oracle while `resolve_system_address` follows the *override* — REX5 semantics executing against pre-REX2 predeploys.

To be clear about scope: that divergence is pre-existing, not introduced here (setup resolved from `hardforks` before this PR too), so I would not block on it. It is worth surfacing because this PR is what made "which spec gates what" explicit, and because it points at the fix for both issues at once: use one source throughout. `resolve_system_address` already takes `self.evm.ctx().mega_spec()`; having `pre_execution_changes` read the same value would make the executor coherent by construction rather than by an assert that is absent from precisely the builds where divergence is reachable.

Worth deciding deliberately either way — if the schedule (not the cfg) is intended to be authoritative for setup even under an explicit spec override, that is a defensible choice, but it should be stated in the field doc rather than left as a consequence of which of the two bindings a given call site happened to reach for.

Troublor (changes requested):

crates/mega-evm/src/block/hardfork.rs
RealiCZ: This comment was written against the earlier revision, where forks mapped many-to-one onto specs and `introduces_spec()` was derived from that map — that code no longer exists. The alias-rung rework replaced it wholesale, so the latest push contains no direct change for this thread; what follows is how the current design answers the same question.

Instead of declaring skippability as an independent per-fork axis (`optional()`), we made the fork→spec map 1:1 and moved the one hand-written declaration to the spec side: `MegaSpecId::behavior()`, one arm per alias. "May a schedule omit this fork" is then derived as `is_alias()` — an alias rung has no setup and no behavior of its own, so omitting it is safe by construction. The two axes geth keeps independent are deliberately collapsed here: an omittable behavior-introducing fork or a mandatory alias is unrepresentable rather than something a validator has to cross-check. MegaETH has neither of geth's diverging cases (no situational forks like the Glaciers, no EIP-level removals like Petersburg — rollbacks here are whole-behavior projections), so the collapse costs nothing today; if a situational behavior fork ever appears, an independent `optional()` axis comes back exactly as proposed in this comment.

The consequences predicted here did land along the way: `OrphanPatch` folded into `SkippedRung` (`test_validate_schedule_rejects_alias_without_base`), `base_fork` is gone, "patch fork" gave way to *alias* with a normative definition in `docs/spec/hardfork-spec.md`, and the 1:1 strict ascent the derivation rests on is now a compile-time const assertion (`climbs_the_spec_ladder`).

Troublor (approved):

vincent-k2026 (approved):
**Theme:** the alias-rung model turns "the fork→spec map must not descend" from a maintained convention into something structurally unrepresentable, and — unusually — the guards that enforce it are themselves tested. Approving; the items below are non-blocking.

**Verified against HEAD:**

```
cargo test -p mega-evm  →  1253 passed / 0 failed
```

Then two deliberate mutations, to check the guards actually discriminate rather than merely exist:

| mutation | result |
| --- | --- |
| `behavior()` → `MINI_REX_2 => MINI_REX_1` (an alias chain) | **compile error**: `error[E0080]: behavior() targets must be concrete specs at or below the projecting rung` |
| `instructions.rs` regroups `MINI_REX_1` under the `mini_rex` table | **test red**: `MINI_REX_1 table diverges from its behavior target EQUIVALENCE at opcode 0x00` |

Both fail as designed.

### Prior rounds

Four of the five open threads are addressed at HEAD and can be resolved: `behavior()` flatness is now pinned by the `is_flat_projection` const assertion (verified firing above); the ladder-climb check is a const assertion (`climbs_the_spec_ladder`); `max_activated_spec_id` is gone (zero references repo-wide); and the executor's `setup_spec` field is gone in favour of `self.evm.ctx().mega_spec()`. The fifth (`optional()` as an independent axis) has RealiCZ's design answer and was superseded wholesale by the alias-rung rework.

### Genuinely good

1. **The model is the fix.** Making fork→spec 1:1 and monotone, and expressing rollbacks as alias rungs, deletes the machinery a reversible map needed — `max_activated_spec_id` disappears because there is no longer a maximum to take. That is a structural elimination, not a refactor.
2. **The checkers are tested.** `is_ladder_prefix`, `is_flat_projection`, and `climbs_the_spec_ladder` all take `&[T]` so the real table goes through a const assertion while malformed tables go through unit tests. Most codebases test the guarded thing and not the guard, which is exactly how a weakened guard ships silently.
3. **`is_flat_projection` covers chains, cycles, upward targets, and duplicate keys as one fixed-point property** rather than four separate checks.
4. **"State it twice with an automatic cross-check"** at every dispatch site: the hand-written alias grouping in `instructions.rs` / `precompiles.rs` / `evm/limit.rs` / `limit.rs` is reconciled against `behavior()` by a generic test driven off `ALL × is_alias()`. Instruction tables compared per opcode with `fn_addr_eq`, precompile sets by pointer identity — the comparison forms are well chosen.
5. **`resolve_system_address` reads one spec through two projections** with the reason stated at the call site: `is_enabled` gates whether dynamic resolution applies, `reaches` selects the installed bytecode version.
6. **`test_missing_params_fail_at_load_time_and_pre_block_alike` is genuinely two-sided** — it asserts the `validate_schedule` error variant *and* drives a real `pre_execution_changes()` failure, rather than restating one side.
7. **`without` made private with a doc that names the idiom it removes**, not just the API: `with_all_activated().without(fork)` reads as "the spec below `fork`" and is not.
8. **The unknown-chain fallback is pinned to a named rung**, and its test comment is honest that a new spec leaves it green — it pins drift, not the decision.

### Non-blocking

**1. `validate_schedule` never runs in a release build.** Its only production call site is the `debug_assert_eq!` in `chain.rs:116`, which is compiled out. The trait doc, `block/AGENTS.md`'s invariant map, and the README all say node startup *should* call it, and nothing does — here or in mega-reth. So the fail-fast half of this design ships inert: a published schedule with a skipped rung or missing params still reaches block execution and leans on the fail-safe (position-compared setup stays additive). `mega-evme` and `mega-t8n` both resolve chain configs and could be the first real caller; failing that, please link a tracked mega-reth follow-up in the PR body rather than leaving a "should".

**2. The required-params roster is the one hand-maintained parallel list left.** `validate_schedule` hardcodes `Rex5 → SequencerRegistryConfig` / `Rex6 → SequencerRegistryRex6Config`; `executor.rs` and `transact_deploy_sequencer_registry_for` hardcode the same two rules independently; the pairing test hardcodes them a third time. A third `HardforkParams` type added without touching all three goes unnoticed — the pairing test does not grow on its own. The doc concedes this ("A new `HardforkParams` type must be registered here"). A single const table — `&[(MegaHardfork, &'static str, fn(&dyn MegaHardforks) -> bool)]` — iterated by both `validate_schedule` and the pairing test would put this on the same footing as everything else in the PR.

**3. Nothing forbids an alias rung that crosses a one-way setup boundary.** `is_flat_projection` only requires the target be downward and a fixed point. A future alias at, say, rung 12 projecting to `EQUIVALENCE` satisfies both, and yields a hybrid: `reaches(REX5)` installs the Rex5+ Oracle v2.0.0 and the SequencerRegistry, while `resolve_system_address`'s `is_enabled(REX5)` gate (`REX5.behavior()=9 <= EQUIVALENCE.behavior()=0` → false) returns `MEGA_SYSTEM_ADDRESS`. The executor then uses the static address while the on-chain Oracle reads the registry — the hybrid-spec class this PR exists to prevent, re-enterable through the one axis nothing checks. Unrepresentable today (both aliases sit below `REX`), so not blocking. Worth either a normative rule in `docs/spec/hardfork-spec.md` ("an alias rung must not cross a one-way setup boundary; the lowest is `MINI_REX`") or a named test that enumerates the `reaches`-gated rungs and asserts every alias sits below the lowest one.

**4. The replay-safety audit covers mega-reth but not mega-kona.** `mega-kona`'s `crates/proof/megaevm/src/lib.rs` also implements `MegaHardforks`, over a `MegaChainSpec` with one `Option<u64>` per fork — precisely the partial-ladder shape where the `is_*_active_at_timestamp` change (raw event → cascading position projection) is observable — and it never calls `validate_schedule`. It is pinned at tag `v1.7.0`, so nothing breaks today; but its `mega_fork_activation` is an exhaustive match over `MegaHardfork` with no `Rex7` arm, so the next bump touches that code anyway, and that is the moment to re-derive its predicates. This is a fault-proof component, so a predicate semantics change is consensus-relevant — worth naming it in the release-note section.

**5. State the discriminant-renumbering audit as a command, not a conclusion.** `MegaSpecId` is `#[repr(u8)]` with derived `Serialize`/`Deserialize`, and this PR moves `REX`..`REX7` from 2..9 to 4..11 — an old `REX`(2) decodes as `MINI_REX_1`(2), i.e. `EQUIVALENCE` semantics, under any variant-index codec. `test_ladder_positions_are_pinned`'s own comment names this hazard. I checked what I could reach: `EvmEnv<MegaSpecId>` is never serialized in mega-reth or stateless-validator, and stateless-validator's bincode witness format is `(SaltWitness, MptWitness)` with no spec field — so the claim holds there. But `gh search code MegaSpecId --owner megaeth-labs` does not even return stateless-validator's four local references, so it under-reports and cannot support a "zero references" conclusion. Either list the repos and the command used, or make `MegaSpecId`'s serde name-based (JSON behaviour unchanged, compact codecs immune to renumbering from here on).

### Nits

- **`is_enabled(alias)` is unconditionally true and perfectly legal.** `spec.is_enabled(MINI_REX_1)` projects the argument to `EQUIVALENCE`, so any behavior gate written against an alias rung is a silent no-op. `reaches(alias)` is meaningful; `is_enabled(alias)` never is. Worth a line in the `is_enabled` doc (which currently only says both sides project), or a test that pins the property so it reads as deliberate.
- **mega-reth's `revm_spec()` is a second statement of `spec_id`**: `chain_spec.hardfork(ts).map(|h| h.spec_id()).unwrap_or(EQUIVALENCE)`, with its own fallback. Now that `MegaHardforks::spec_id` is trait API, that call site should just use it so the fallback cannot diverge — a line in the PR description would be enough to route it.
- **`BlockLimits::from_hardfork_and_block_gas_limit` still keys on `MegaHardfork`**, the only dispatch site where a *fork* selects behavior; the other three moved to `MegaSpecId`. Making it `from_spec` would let the same generic reconciliation cover it.
- **The Performance section contradicts itself**: "zero instruction-count change across all pre-existing benchmarks" versus "the `rex4` rows ... their CodSpeed baselines shift accordingly". The second is the accurate one; the first needs narrowing.

PR #365, opened 2026-08-04
chore: upgrade to revm 40.0.3 / alloy-evm 0.36.0

RealiCZ:
## Summary

- Upgrades the workspace from revm 27 to **revm 40.0.3**, with `op-revm` / `alloy-op-evm` pinned to the Optimism monorepo revision wired to reth v2.3.0, and `alloy-evm` 0.36.0.
- revm 40 reworked interpreter and handler internals underneath the frozen MegaETH specs; this PR re-pins the frozen semantics so observable behavior stays identical for every existing spec.
- Includes one semantic fix surfaced while rebasing onto the cross-spec compute-gas suite (#358): Rex6 CALL-family first-touch pricing of preload-warm addresses (second commit).

## Frozen-semantics re-pinning

- **Static-gas pre-charge**: revm 40's `step()` deducts static gas from a per-spec gas table before the handler runs. Volatile-guarded opcodes get zeroed table entries and charge inside the wrapper at the revm-27-equivalent position; all other opcodes add the pre-charge back into the compute-gas measuring window (`STATIC_GAS_TABLE`).
- **Halt handling**: limit-triggered halts go through `set_halt_action!`, which preserves remaining gas for sender refund instead of revm 40's `spend_all`-on-OOG and explicitly replaces pending frame actions.
- **Host entry points**: the oracle SLOAD customizations (env override, forced-cold pricing, access marking) moved into `sload_skip_cold_load`; resident journal entries keep their own cold pricing; access-list warming only applies to entries with storage keys; CALL-family account loads distinguish the raw operand from its EIP-7702 delegate.
- **System calls**: pre-REX5 system calls keep the 30M gas limit via `PRE_REX5_SYSTEM_CALL_GAS_LIMIT` (the upstream constant drifted to 30M + EIP-8037 reservoir) and continue to skip `post_execution`.
- **Precompiles**: error paths normalize spent gas back to zero; the KZG fixed-rate compute-gas branch keys on `bytecode_address`.
- **CfgEnv**: spec changes go through `set_spec_and_mainnet_gas_params` so `GasParams` follow the spec; `tx_chain_id_check` keeps its revm-27 default for untouched configurations (`with_cfg_unpinned` opts out); EIP-8037 is pinned off in every conversion.
- **Block executor**: `commit_transaction` failures are latched and surfaced in `finish()`; state hooks moved onto revm's `State`; results carry `tx_type` instead of a cloned envelope.

## Rex6 fix (second commit)

Before revm 40, the CALL-family host entry resolved the raw operand's EIP-7702 delegate before loading it, and that resolution materialized the operand's journal entry (cold, code hydrated) as a side effect, so a Rex6 first touch of a preload-warm address (precompile, coinbase, address-only access-list entry) was priced cold.
revm 40 resolves the delegate inside the CALL instruction; the migration's phase-based marking lost the materialization side effect, pricing such first touches warm — 2,500 gas below frozen Rex6 behavior in both EVM gas and compute gas.
The fix reproduces the materialization in `load_account_info_skip_cold_load` for Rex6 CALL-family raw-operand loads.
Caught by `test_callcode_cold_first_touch_follows_the_spec_arc` from #358.

## Accepted deviation: the frozen detention window (debug-tripwired)

- revm 27 loaded a CALL-family (and EXTCODECOPY) target before charging the opcode's own costs, so a frame that ran out of gas on those charges had already marked beneficiary access and the rest of the transaction ran detained; revm 40 charges first, so the same frame halts unmarked and the rest of the transaction runs undetained.
- The halting frame itself is byte-identical either way — out-of-gas consumes the frame's remaining gas, the load rolls back with the frame, no logs — so the divergence is exactly the tracker mark and the detention cap it would have set for the rest of the transaction.
- Normal traffic cannot sit in the window (a frame holding under 100 gas — under 9,100 with value — still calling the beneficiary), but sampling is not the guarantee: the release gate is a full-history replay run with debug assertions enabled.
- This branch ships a debug-build tripwire (`debug_check_frozen_detention_window`) on every plain out-of-gas exit of the guarded wrappers: it fires exactly when a halt leaves the beneficiary — or, from REX6, its one-hop delegate — unmarked where revm 27 marked it. Release builds compile it out (verified absent from the release rlib; CodSpeed should report zero delta).
- If replay ever trips it, the wrapper backfill archived in the migration notes restores the revm-27 marking exactly; until then the revm 40 order is the accepted behavior. Known blind spot: pre-REX4 CALL frames die in `step()`'s table pre-charge before any wrapper runs, so that era is covered by the replay gate alone.
- Writing the tripwire tests surfaced and fixed one spec-doc error: `gas-detention.md` listed EXTCODECOPY among the opcodes that register volatile access before charging; its copy cost is charged before the load, so it belongs with the CALL family on the excluded side.
- **Gate executed (2026-08-10→17): zero tripwire hits across all of mainnet history.** All Rex4+ blocks (13,862,189→23,718,840, ~270M txs) replayed in full, and the pre-REX4 era (blocks 1→13,862,188) — including the blind-spot window noted above — replayed in full except a ~11B-tx homogeneous swap-spam region covered by ~1.2B txs of full chunks plus spatial slice samples, with `debug_assertions` + `overflow_checks` armed throughout. The revm-40 charge order is confirmed as the accepted behavior; the archived wrapper backfill remains unused.

## Output format changes (no execution change)

- revm-inspectors 0.27 → 0.40 changes struct-log serialization: every log now carries `"refund": 0` and memory words gain a `0x` prefix. Gas and gasCost values in the trace fixtures are byte-identical to main — execution did not move, serialization did. Anyone diffing `mega-evme run --trace` output against archived traces (or geth) will see the field-level drift.

## Verification

- `cargo test --workspace`: 1,468 passed / 0 failed (includes the EEST state-test suite, the replay corpus, and the cross-spec compute-gas snapshot suite from #358).
- Mainnet replay hard gate (2026-08-10→17): all Rex4+ history (~270M txs) plus the pre-REX4 era re-executed against on-chain receipts (status, gasUsed, logs). Every observed divergence was mechanically attributed — an EIP-7702 fresh-authority refund artifact of the RPC replay tooling (since fixed by mega-evme `36081e7`) and the halt-receipt log leak this gate caught (fixed in this PR, `7156d05`) — and a 92/92-chunk re-verification with both fixes confirmed zero remaining divergence.
- `cargo clippy` with `-D warnings`, `cargo fmt --check`, `cargo sort --check`, and the `riscv64imac-unknown-none-elf` no_std check all pass.

## Status

Marked [WIP]: a manual semantics review of the migration is still in progress; no further code changes are expected unless the review finds issues.


crates/mega-evm/src/evm/precompiles.rs
```diff
@@ -98,17 +100,22 @@ pub mod kzg_point_evaluation {
     /// Gas cost for the KZG point evaluation precompile.
     pub const GAS_COST: u64 = 100_000;
 
+    /// Runs the upstream KZG point evaluation and overrides the reported gas with `MegaETH`'s
```
flyq, line 176: **[blocker — frozen-spec drift]** This fixed-cost wrapper pattern is needed one screen up for modexp too: `mini_rex()` installs `revm::precompile::modexp::OSAKA` unwrapped (line 81 — unchanged in this diff, which is exactly the problem: upstream changed the semantics underneath it). revm-precompile 25.0.0 returned `Ok(min_gas)` for `base_len == 0 && mod_len == 0` *before* reading `exp_highp` or computing gas; 36.0.3 moved that special case *after* the `gas_cost > gas_limit → OutOfGas` check and now charges `gas_cost` instead of `min_gas` (verified in both sources). Concrete replay divergence on any MINI_REX…REX6 chain: a call to 0x…05 whose 96-byte header declares `base_len=0, mod_len=0, exp_len=64` (no body) was a flat 500-gas success at any limit ≥ 500; it now charges 8,192 — or halts PrecompileOOG burning the entire forwarded gas when 500 ≤ limit < 8,192. Gas used, receipt status, refunds and downstream state all move. Same flip for `exp_len ≤ 32` with exponent ≥ 2^32. EQUIVALENCE is unaffected (BERLIN modexp's multiplication complexity is 0 in this case, so both orderings coincide). Fix: wrap `modexp::OSAKA` the way this KZG wrapper does, reproducing the old early-return position + `min_gas` charge for all current specs (REX7 included, until a spec deliberately adopts the new EIP-7883 ordering), and add zero-base/zero-mod tests — `grep modexp` finds no mega test today.

crates/mega-evm/src/evm/instructions.rs
```diff
@@ -1201,28 +1593,47 @@ pub mod volatile_data_ext {
...
+    /// This family is the one that cannot put its static gas back where revm's body charges it.
+    /// The body charges it as part of the call cost, after the target load — but it then derives
+    /// the child's gas limit from what the frame has left, and `forward_gas_ext` re-derives its
+    /// 98/100 cap from the same quantity. A charge taken after the body would therefore hand the
+    /// child ~98 gas more than the schedule allows. Charging it before the body keeps every
+    /// forwarded amount identical, because the same total is deducted before either split is
+    /// computed; only the order of the deductions within the body differs, which is unobservable
+    /// for a frame that can afford them.
+    ///
+    /// A frame that *cannot* afford the charge is the case where that order shows: the body never
```
flyq, line 1708: **[major — needs explicit sign-off, currently silent]** This comment accurately describes the mechanics but quietly accepts a frozen-spec drift the PR body never mentions. Under revm 27 the CALL family loaded the target *before* charging (revm-interpreter 24 `contract.rs:122-143`), and the load marked beneficiary access in the non-journaled tracker — so a frame that then OOG'd still tightened the tx-wide detention cap. After this PR the charge comes first on every spec ≥ MINI_REX: via the kept 100-gas table entry pre-charged by `step()` (MINI_REX…REX3), via this wrapper's charge-before-body (REX4+), and via upstream's 9,000 value-transfer cost now sitting before the load. A frame CALLing the beneficiary (or its REX6 delegate) with gas in the `[M, M+100)` window (`[M, M+9,100)` with value) historically marked-then-OOG'd; now it OOGs unmarked, and the *rest of the transaction* runs with an un-detained compute budget — a tx that historically halted on the detained limit (with rescued-gas refund) can now run to completion with different state, gas, and logs. The new gas-detention.md section canonizes this by listing only BALANCE/EXTCODE*/SLOAD/SELFDESTRUCT as still-marking-on-OOG. Asks: (a) disclose this as an accepted deviation in the PR body; (b) scan the replay corpus / chain history for the trigger window — the empirical count is the decisive input for accepting vs. vendoring revm's call body; (c) make the doc say explicitly that the CALL family is excluded and why, rather than only listing the included opcodes.

crates/mega-evm/src/evm/instructions.rs
```diff
@@ -275,10 +332,18 @@ mod rex2 {
...
 
-        table[SELFDESTRUCT as usize] = compute_gas_ext::selfdestruct;
+        table[SELFDESTRUCT as usize] = Instruction::new(compute_gas_ext::selfdestruct);
 
         table
     }
+
+    /// Returns the static gas table for the `REX2` spec.
+    ///
+    /// `SELFDESTRUCT` is re-enabled above without a volatile guard (Rex4 adds one), so the zeroed
```
flyq, line 351: **[major — cheap fix available]** The zeroed set is about *where* gas is charged, but for specs whose handler is `control::invalid` the kept entry also changes *what a failing frame reports*: MINI_REX/REX/REX1 keep revm's 5,000 SELFDESTRUCT table entry while wiring the handler to `control::invalid`, and revm 40's `step()` debits the entry before dispatch — so a disabled SELFDESTRUCT reached with < 5,000 remaining now halts `OutOfGas` where revm 27 (which charged nothing before the handler) reported `InvalidFEOpcode` on every historical execution. Gas (all consumed) and state are identical, so receipts do not move — but the halt reason is frozen surface: replay tooling, debug traces, mega-evme outcomes, and state-test exception mapping all see it. Same mechanism in tiny-gas windows for CLZ (entry 5, `NotActivated`→`OutOfGas`) and SLOTNUM/DUPN/SWAPN/EXCHANGE (`OpcodeNotFound`→`OutOfGas`), including under EQUIVALENCE via upstream's own pre-charge. Zeroing the disabled-opcode entries in the affected spec tables restores the frozen reason for the cost of a few table writes — fixing seems strictly better than signing off the drift.

crates/mega-evm/src/evm/context.rs
```diff
@@ -311,16 +385,80 @@ impl<DB: Database, ExtEnvTypes: ExternalEnvTypes> MegaContext<DB, ExtEnvTypes> {
...
+    /// [`with_cfg_unpinned`](Self::with_cfg_unpinned): both apply the same pins, and differ only
+    /// in whether the untouched-default inference is consulted for `tx_chain_id_check`.
+    fn apply_cfg(mut self, mut cfg: CfgEnv<MegaSpecId>, intent: CfgIntent) -> Self {
+        // Normalize before classifying. EIP-8037 is pinned off on every configuration that
+        // reaches the EVM, so it is not a field the caller gets to choose and must not be part of
+        // the shape the default-form test reads — otherwise two configurations that normalize to
+        // the same thing would come out with different `tx_chain_id_check`.
+        pin_amsterdam_eip8037_off(&mut cfg);
+
+        let pin_revm27_chain_id_check =
```
flyq: **[major — needs a decision; overlaps the open codex P1]** The inference makes the effective `tx_chain_id_check` depend on whether an embedder touched *any unrelated* cfg field. The production path is the sharp end: `MegaEvmFactory::create_evm → with_cfg(evm_env.cfg_env)` (factory.rs:161) hands over a configured env (real chain id, size limits), so it is taken at face value — i.e. whatever reth v2.3.0 puts in `CfgEnv`, and revm 40 defaults the flag to `true` where revm 27 ran mainnet with `false`. The gate rejects `Some(wrong_id)` on any tx and `None` on non-legacy txs. Deposits skip validation and pool-validated txs carry correct ids, so sequencer-built history *should* contain no tripping tx — but "should" is not the frozen-spec standard, and mega-evme/t8n inherit the same cliff the moment their cfg differs from the default shape in any field. Ask: make it a decision, not an inheritance — pin `false` explicitly on the node/replay construction path (or `with_cfg_unpinned` with a deliberate value), or verify and document that reth v2.3.0 sets `false`, plus one factory-path test pinning whichever choice is made.

crates/mega-evm/src/evm/host.rs
```diff
@@ -215,6 +282,188 @@ impl<DB: Database, ExtEnvs: ExternalEnvTypes> Host for MegaContext<DB, ExtEnvs>
...
+    /// read still counted as an oracle access; marking first preserves that.
+    #[inline(never)]
+    fn oracle_sload(
+        &mut self,
+        address: Address,
+        key: StorageKey,
+        skip_cold_load: bool,
+    ) -> Result<StateLoad<StorageValue>, LoadError> {
+        debug_assert!(
+            self.spec.is_enabled(MegaSpecId::MINI_REX) && address == ORACLE_CONTRACT_ADDRESS,
```
flyq, line 383: **[ci — spec-gate FAIL resolution]** The surviving mutant (`MINI_REX → EQUIVALENCE` on this line) is *equivalent*, not a coverage gap: this `debug_assert!` restates the caller's guard (`sload_skip_cold_load`, line 181 — byte-identical to the baseline gate, whose own mutants were killed), and since `is_enabled(EQUIVALENCE)` is tautologically true, the mutation strictly weakens an assert that never fires on correct code — no behavior-observing test can distinguish them, and the assert compiles out of release anyway. The honest fix is a justified entry in `mutants/suppressions.toml` (equivalent mutant: assert-weakening), not a test written to chase the score. Optional hygiene: a `#[should_panic(expected = "oracle_sload is only reachable")]` debug-build unit test in this module calling the private path directly under EQUIVALENCE would kill it.

Cargo.toml
```diff
@@ -127,33 +127,37 @@ mega-state-test = { path = "./crates/mega-state-test", version = "1.7.0", defaul
...
 
 # revm
-op-revm = { version = "8.1.0", default-features = false }
-revm = { version = "27.1.0", default-features = false }
-revm-inspectors = { version = "0.27.3", default-features = false }
+# The OP crates are pinned to the Optimism monorepo revision whose `rust/` workspace is wired to
+# reth v2.3.0 (revm 40.0.3, alloy-evm 0.36.0). crates.io has no `op-revm` / `alloy-op-evm` release
+# built against revm 40, so every OP crate below must share this exact source to keep a single
+# source identity for `OpTxEnvelope`, receipt and hardfork types.
+op-revm = { git = "https://github.com/ethereum-optimism/optimism", rev = "f67d87cd5c9992b62d215c043917b94e3f019629", default-features = false }
```
flyq, line 139: **[minor — release-process check]** Full-SHA pinning is the right hygiene here, but note the operational consequence: `cargo publish` rejects git dependencies, so a crates.io release of `mega-evm` or `mega-state-test` (the latter documented as published) is impossible while these six OP-family git pins stand. If releases are GitHub-tag-only this is dormant — worth confirming against the bump-version-release flow before the next version bump, and a one-line comment here ("blocks crates.io publishing until op-revm ships a revm-40-compatible release") would save the next person the discovery.

crates/mega-evm/src/evm/host.rs
```diff
@@ -149,45 +166,95 @@ impl<DB: Database, ExtEnvs: ExternalEnvTypes> Host for MegaContext<DB, ExtEnvs>
...
+    // bodies that can drift apart; keep it deleted.
+
     fn balance(&mut self, address: Address) -> Option<StateLoad<U256>> {
         self.check_and_mark_beneficiary_balance_access(&address);
         self.inner.balance(address)
     }
 
+    /// Loads `address` and its EIP-7702 delegate, marking beneficiary access for both from Rex6.
+    ///
+    /// The CALL-family opcodes no longer reach this method: revm resolves the delegate inside the
```
flyq: **[minor]** This doc note covers `load_account_delegated`, but the same applies to `balance`, `load_account_code`, and `load_account_code_hash`, which carry no warning: they delegate straight to `self.inner`, i.e. revm 40's journal, whose resident path consults the pre-warmed sets — so a direct trait-method caller gets a *warm* answer for a resident-but-cold precompile/coinbase where the frozen pricing rule (and every interpreter-driven opcode via `load_account_info_skip_cold_load`) says *cold*. Verified not consensus-reachable today (the interpreter prices BALANCE/EXTCODE* through `load_account_info_skip_cold_load`, and SELFBALANCE ignores coldness), so this is future-proofing: either route the other three through the pricing override or extend this warning to them.

crates/mega-evm/src/block/executor.rs
```diff
@@ -670,42 +862,124 @@ where
...
+    /// records the failure, and reports zero gas used, which is accurate: the transaction
+    /// contributed nothing to the block. [`alloy_evm::block::BlockExecutor::finish`] then fails
+    /// the block with the recorded error, so a rejection can never be silently dropped;
+    /// [`MegaBlockExecutor::pending_commit_error`] exposes it earlier for callers that want to
+    /// react before `finish`.
+    ///
+    /// Prefer [`MegaBlockExecutor::commit_tx_result`] when the caller can act on a rejection: it
+    /// is the same commit, returns the error instead of latching it, and leaves the executor
+    /// free to continue building the block without the rejected transaction.
+    fn commit_transaction(&mut self, output: Self::Result) -> GasOutput {
```
flyq, line 922: **[minor]** The latch is correct for every in-crate path (`execute_transaction_with_commit_condition` routes through the fallible `commit_tx_result`, rejection commits nothing, `finish()` errors first thing on the latch — and the new tests pin all three). The bypass risk is one `pub` field away: `receipts` is public, so an embedder driving raw `commit_transaction` calls and harvesting `executor.receipts`/state directly — never calling `finish()` or `pending_commit_error()` — builds a block missing the rejected tx while believing all commits succeeded; `take_pending_commit_error` additionally lets a caller clear the latch without acting on it. Consider privatizing `receipts` behind an accessor, or at minimum moving the contract ("results are only valid once finish()/pending_commit_error is consulted") onto the field's doc.

docs/spec/evm/gas-detention.md
```diff
@@ -145,6 +145,13 @@ If a child call frame triggers detention, the reduced effective compute gas limi
 If volatile access occurs inside a call frame that later reverts, the compute gas already consumed remains consumed.
 The detained compute gas limit MUST remain in effect for the rest of the transaction.
 
+#### Frames That Run Out of Gas on the Triggering Opcode
+
+Detention is triggered by the volatile read itself, not by the successful completion of the opcode that issues it.
+`BALANCE`, `EXTCODESIZE`, `EXTCODECOPY`, `EXTCODEHASH`, `SLOAD` and `SELFDESTRUCT` consume their operands and read state before that access is charged for.
+A frame that reaches one of them holding less gas than the access costs therefore still performs the read, and detention MUST still be applied for the rest of the transaction even though that frame halts out of gas.
```
flyq: **[nit — wording]** "still performs the read" is not literally true for the oracle-SLOAD path: with remaining gas below the cold surcharge, revm 40 requests a skip-cold load and `oracle_sload` marks the tracker *then bails* with `ColdLoadSkipped` without reading the slot (host.rs:436-442). The normative MUST is correct either way (the mark precedes the bail, and the frame OOGs so the skipped value is unobservable) — suggest "the access is still registered, and detention MUST still be applied" so the descriptive sentence matches the code.

crates/mega-evm/src/block/eips.rs
```diff
@@ -168,8 +175,7 @@ pub(crate) fn transact_balance_increments<DB: Database>(
         if balance_increment == 0 {
             continue;
         }
-        let cache_account = db.load_cache_account(address)?;
-        let account_info = cache_account.account_info().unwrap_or_default();
+        let account_info = db.basic(address)?.unwrap_or_default();
```
flyq: **[nit]** `Account::default().with_info(...)` on the next line vs `Account::from(info)` in the deploy/sequencer-registry helpers leaves `original_info` unset on this path only. Inert today: revm 40's `CacheState::apply_evm_state` consults `original_info()` only when the committed account is *absent* from the cache, and every pre-block helper loads via `db.basic()` first, so the entry is always occupied. But a future helper that commits an account it did not first load would be mis-baselined as previously-empty in the bundle. Unifying on `Account::from` (or one comment stating the load-before-commit invariant) removes the trap.

crates/mega-evm/src/evm/context.rs
```diff
@@ -149,8 +177,20 @@ impl<DB: Database, ExtEnvs: ExternalEnvTypes> MegaContext<DB, ExtEnvs> {
...
+            .with_db(db)
+            .with_cfg(CfgEnv::new_with_spec(spec.into_op_spec()));
+
+        // revm 40 flipped `CfgEnv::tx_chain_id_check` default from `false` (revm 27) to `true`.
+        // With the new default, non-legacy txs with `chain_id: None` (and many unit-test
+        // TxEnvBuilder::default() shapes) fail validation with MissingChainId.
+        // Pin revm-27 semantics: the whole chain-id gate is off unless a caller later
+        // enables it via `with_cfg` / factory `EvmEnv`. Note: with the flag off, a
+        // mismatched `Some(wrong_id)` is also accepted — same as revm 27.
+        inner.cfg.tx_chain_id_check = false;
```
flyq, line 193: **[nit]** This constructor pins `tx_chain_id_check` but relies on the upstream default for EIP-8037 (`new_with_spec_and_gas_params` leaves it off only because every `MegaSpecId` maps to pre-Amsterdam ISTHMUS). Every other conversion calls `pin_amsterdam_eip8037_off` belt-and-braces; one more call here makes the constructor robust against a future op-spec base bump.

crates/mega-evm/src/block/result.rs
```diff
@@ -80,6 +134,11 @@ impl InvalidTxError for MegaTxLimitExceededError {
     fn is_nonce_too_low(&self) -> bool {
         false
     }
+
+    fn as_invalid_tx_err(&self) -> Option<&InvalidTransaction> {
```
flyq, line 142: **[minor — needs a mega-reth-side test]** Returning `None` here (and in the second impl below) is defensible — there is genuinely no `InvalidTransaction` counterpart for a Mega limit rejection — but these impls exist purely for reth's error classification and have zero in-repo callers (they are also this file's codecov-0% lines). If any reth path treats `None` as "internal error → abort payload building" rather than "drop this tx", a Mega limit rejection gets misclassified, and nothing in this repo would catch it. Worth one integration test on the mega-reth side pinning the intended classification.

bin/mega-evme/tests/fixtures/test_run_trace_inline.json
```diff
@@ -18,30 +18,28 @@
           "gasCost": 3,
           "depth": 1,
           "stack": [],
-          "memory": []
+          "memory": [],
+          "refund": 0
```
flyq, line 22: **[minor — release note]** Real output drift, not reformatting: every struct log now serializes `"refund": 0` and memory words gain the `0x` prefix (revm-inspectors 0.27→0.40), and the integration test compares parsed JSON. All gas/gasCost values in this fixture are byte-identical to main — execution did not move, serialization did. Anyone diffing `mega-evme run --trace` output against archived traces (or geth) will see it; deserves a line in the release notes.

flyq (commented):
## Review summary

**Verdict: not mergeable yet — one frozen-spec blocker (modexp) and two silently-accepted frozen-spec drifts that need explicit sign-off — but the migration's foundation is excellent and the hard parts are done right.** Everything below was verified against both upstream trees (revm-interpreter 24↔37, revm-handler 8↔20, revm-context 8↔18, revm-precompile 25↔36, revm-state/database 7↔12/15, op-revm 8.1.0 ↔ the pinned git 20.0.0), not just by reading this repo.

Mechanisms verified as faithful re-pins: the zeroed static-gas entries equal the volatile-guarded set exactly per spec, and the add-back arithmetic reproduces revm-27 compute-gas totals with no double-count on any wrapper; `set_halt_action!` preserves remaining gas (vs revm 40's spend-all-on-OOG) through to `try_rescue_gas` on all six limit-halt sites, and both interpreter loops skip the fallback halt when an action is set; the three gas-leakage paths and the resource-limit latch protocol survive the handler rework; `PRE_REX5_SYSTEM_CALL_GAS_LIMIT` correctly freezes 30M against upstream's drift to 31,566,720, and `run_system_call` still skips validate/pre/post_execution; precompile error paths restore revm-27's no-spend halt shape and `bytecode_address` keying reproduces revm-27 dispatch (CALLCODE/DELEGATECALL-to-KZG identical); EIP-8037 is pinned off at every conversion, normalized before the untouched-default classification; the block executor's receipt assembly, state-hook-before-apply ordering, pre-block sequence, and interception contract are byte-equal or provably equivalent; and the resident-entry cold-pricing compensation (`resident_entry_prices_cold` + `access_list_preloads_account`) faithfully restores revm-27 journal pricing across the full warm/cold matrix — upstream's occupied path now consults the warm sets, and the new `rex/precompile_access_coldness.rs` + `rex/access_list_resident_warming.rs` files are what hold that compensation in place. The Rex6 fix commit (88c6b04) is correct on scope (CALL-family RawOperand bracket only, REX6+ only — pre-REX6 never depended on the side effect, and CALLCODE inspecting the caller instead of the target is exactly why the drift surfaced on the CALLCODE spec-arc), idempotency, delegate pricing, and EVM-gas/compute-gas symmetry.

The test migration deserves its own call-out: **zero pre-existing expectations changed.** A token-level sweep of the full ~8k-line tests diff found no lost constant and no deleted test (958→1020, removals = ∅), every assertion edit is a value-identical API reshape (each verified against the upstream definitions, e.g. `gas_used()`→`tx_gas_used()`), and the #358 compute-gas suite + snapshot are byte-identical to main — the source was fixed to match unchanged pins, which is the right direction.

Blocking / decision items are inline comments: the unwrapped modexp OSAKA reorder (blocker, cheap wrapper fix), the CALL-family under-funded-frame detention-mark loss (accepted in a code comment + doc but undisclosed in the PR body — needs sign-off and a replay-corpus scan), the disabled-SELFDESTRUCT halt-reason drift (cheap table fix), and the `tx_chain_id_check` production-path default (overlaps the open codex P1).

CI notes: the spec-gate survivor is an equivalent mutant — see the inline comment at `host.rs:425`; a justified `mutants/suppressions.toml` entry is the honest fix. CodSpeed: the −20% on `revm_pinned`/`op_revm_pinned` measures the upstream 27→40 jump itself — those "control lanes" changed meaning in this one PR, while `.github/scripts/bench_compare.py:278-281` still documents them as immutable and still lists the deleted `revm_latest`/`op_revm_latest` rows (`:410`); the number to actually track across this PR is the mega-vs-pinned overhead ratio. Non-anchorable minor: dual `alloy-eip7928` (0.3.7 via alloy-eips, 0.4.5 via revm-state) in the lockfile — dormant pre-Amsterdam, but it is the one gap in the manifest comment's "single source identity" promise.

Since the PR is [WIP] pending your own semantics review: the per-mechanism list above is exactly what that review would re-derive — suggest spending the remaining effort on the corners still unaudited here: a line-diff of the remaining eth precompiles across revm-precompile 25→36 (the modexp reorder proves upstream slips semantic changes into these files), and `FromRecoveredTx`/`FromTxWithEncoded` field-by-field parity.

Merge-order: land #362 first — this PR does not touch `spec.rs`, and a dry-run cross-merge conflicts only on `executor.rs` + `instructions.rs`, so the cheaper rebase is on this side.

RealiCZ:
@codex review

crates/mega-state-test/src/types/test_unit.rs
```diff
@@ -140,14 +137,17 @@ impl TestUnit {
         } else if let (Some(parent_blob_gas_used), Some(parent_excess_blob_gas)) =
             (self.env.parent_blob_gas_used, self.env.parent_excess_blob_gas)
         {
+            // A fixture may pin the parent's target blob gas per block; only fall
+            // back to the Cancun target when it does not.
+            let parent_target_blob_gas_per_block = self
+                .env
+                .parent_target_blobs_per_block
+                .map_or(eip4844::TARGET_BLOB_GAS_PER_BLOCK_CANCUN, |target| target.to());
```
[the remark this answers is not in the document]
RealiCZ: The unit mismatch is real on its face, but we resolved it by removing the field rather than converting the count, because the field has zero producers: the local fixture corpus never uses it, current EEST does not emit it (EIP-7742, which introduced it, was withdrawn), and it never landed in geth.
The raw gas-unit pass-through also predates this PR — the pre-migration code passed the field into the derivation unchanged with the same Cancun gas-unit fallback; this PR only preserved that output (the swapped first two arguments are symmetric in the excess formula).
Upstream revm's statetest `Env` has since dropped its parent-target field entirely; modern fixtures pin `currentExcessBlobGas` directly.
So `b3ab60c` deletes the field and its wiring: excess blob gas derivation always uses the Cancun target, and since the deserializer is `deny_unknown_fields`, any fixture still carrying the field now fails loudly instead of being silently misread.

crates/mega-evm/src/evm/host.rs
```diff
@@ -133,13 +133,21 @@ impl<DB: Database, ExtEnvs: ExternalEnvTypes> Host for MegaContext<DB, ExtEnvs>
             None
         };
 
-        let result = self.inner.selfdestruct(address, target);
+        // Read before the destruction: it loads `target`, marking its entry warm. The beneficiary
+        // load is the one the interpreter prices, so it follows the resident-entry coldness rule
+        // described on [`Self::resident_entry_prices_cold`] — and the `inspect_account` above is
+        // itself one of the ways `target` ends up resident but cold.
+        let resident_entry_is_cold = self.resident_entry_prices_cold(&target);
+        let mut result = self.inner.selfdestruct(address, target, skip_cold_load);
```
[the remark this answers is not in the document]
RealiCZ: Ordering fact confirmed against both interpreters (24.0.0 vs 37.0.3): revm 27 always performed the beneficiary load before charging the dynamic cost, revm 40 gates the load on affordability.
We are keeping the forwarded skip hint: on a healthy database the two orderings are byte-identical — the frame halts out-of-gas either way, and the revm-27 load's journal effects (warm mark, balance move) are reverted with the halting frame, so gas, receipts, and state roots all match. The mainnet replay corpus validated this surface.
Divergence requires a cold beneficiary, gas below the cold surcharge, and the DB read itself erroring — a broken node outside the consensus model, and under revm 40 such a node now coincidentally produces the same result as healthy nodes, where revm 27 aborted block execution.
The beneficiary-detention marking surface is unaffected: the SELFDESTRUCT wrapper peeks the stack operand and does not depend on the load.
Same class as the SLOAD skip-cold-load thread above (host.rs:175); recorded in the won't-fix registry as part of that family.

crates/mega-evm/src/evm/host.rs
```diff
@@ -149,69 +157,268 @@ impl<DB: Database, ExtEnvs: ExternalEnvTypes> Host for MegaContext<DB, ExtEnvs>
...
+    }
+
+    fn sstore_skip_cold_load(
+        &mut self,
+        address: Address,
+        key: StorageKey,
+        value: StorageValue,
+        skip_cold_load: bool,
+    ) -> Result<StateLoad<SStoreResult>, LoadError> {
+        self.inner.sstore_skip_cold_load(address, key, value, skip_cold_load)
```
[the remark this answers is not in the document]
RealiCZ: Ordering fact confirmed against both interpreters (24.0.0 vs 37.0.3): revm 27 ran `Host::sstore` before the EIP-1706/2200 sentry, revm 40 runs the sentry and static charge first.
We are keeping the forwarded entry point: on a healthy database the orderings are byte-identical — both halt the frame with the sentry out-of-gas, and the revm-27 load's warm mark is reverted with the halting frame, so gas, receipts, and state roots all match. The EQUIVALENCE mainnet replay corpus validated this surface.
Divergence requires gas at or below the stipend plus the storage read itself erroring — a broken node outside the consensus model, and under revm 40 such a node now coincidentally produces the same result as healthy nodes, where revm 27 aborted block execution.
Same class as the SLOAD skip-cold-load thread (host.rs:175); recorded in the won't-fix registry as part of that family.

vincent-k2026:
Reviewed as a targeted pass over the highest-risk seams (`STATIC_GAS_TABLE` / `charge_static_gas!`, `set_halt_action!`, the block-executor latch), not line-by-line across all 195 files. Flagging that up front so the depth of this review isn't overstated.

Genuinely good:

- **The release gate is real work.** Full-history replay 2026-08-10→17, all Rex4+ (~270M txs) plus the pre-REX4 era, `debug_assertions` + `overflow_checks` armed throughout, zero tripwire hits. Guarding an accepted frozen-semantics deviation with a debug-only tripwire and then disproving it against the whole chain is the right shape — and the blind spot (pre-REX4 CALL frames die in `step()`'s table pre-charge before any wrapper runs) is called out by you rather than found by a reviewer.
- `set_halt_action!` (`instructions.rs:550-566`) not using revm's `halt` is correctly argued: `halt` does `spend_all` on `OutOfGas`, which would burn gas MegaETH refunds to the sender, and it refuses to set an action while one is pending, whereas a limit abort right after a CALL/CREATE published its child `NewFrame` has to *replace* it. Both differences are named, and the comment explains why rather than what.
- `test_static_gas_table_is_spec_invariant` (`instructions.rs:3014`) walks `ALL_SPECS` and asserts each builds the same table — a real drift test that goes red if a spec ever remaps its eth spec.

## Blocking

**1. The body says "Marked [WIP]: a manual semantics review of the migration is still in progress", but the title carries no WIP, the PR is not a draft, and it is `MERGEABLE`.** megaeth-labs/stateless-validator#184 is blocked on this getting a tag, so the tag is the unlock condition for downstream repos — it shouldn't happen while the semantics review is open. Please align title/draft state with the body and state the completion condition for that review.

## Should fix

**2. `MegaBlockExecutor::receipts` is `pub`, and its correctness depends on the caller checking `pending_commit_error` first.** Your own doc comment at `block/executor.rs:64-72` writes the failure mode down:

> "A caller harvesting this field (or the equivalent trait accessor) directly must consult `pending_commit_error` first, or let `finish()` fail the block; otherwise a rejected transaction silently vanishes from the block it believes it built."

Documenting it is good; leaving the latch enforceable only by that comment is not. Anywhere in mega-reth that reads `receipts` directly gets a block missing a transaction with no error anywhere. Structural fix: drop `receipts` to `pub(crate)` and expose `fn receipts(&self) -> Result<&[R::Receipt], &BlockExecutionError>` so the latch can't be bypassed.

**3. Ordering note for #347.** Once static gas moves into `step()`'s table pre-charge, any hand-inlined instruction body that keeps `gas!(..., gas::VERYLOW)` in the body double-charges silently. #347 currently has three of those (`add`, `pop`, `push1`). If it is kept, it has to land *after* this PR and be re-derived on top of it — details in my comment there.

crates/mega-evm/src/evm/context.rs
```diff
@@ -311,16 +367,62 @@ impl<DB: Database, ExtEnvTypes: ExternalEnvTypes> MegaContext<DB, ExtEnvTypes> {
...
+
+    /// Shared body of [`with_cfg`](Self::with_cfg) and
+    /// [`with_cfg_unpinned`](Self::with_cfg_unpinned): both adopt the caller's configuration the
+    /// same way, and differ only in whether `tx_chain_id_check` is pinned to the revm-27 `false`
+    /// or taken as the caller provided it.
+    fn apply_cfg(mut self, cfg: CfgEnv<MegaSpecId>, intent: CfgIntent) -> Self {
         self.spec = cfg.spec;
         self.inner = self.inner.with_cfg(cfg.into_op_cfg());
+        if intent == CfgIntent::Pinned {
+            self.inner.cfg.tx_chain_id_check = false;
```
Troublor, line 424: Non-blocking for this PR — but I'd like point 2 tracked and picked up in a follow-up rather than dropped.

`with_cfg` pins `tx_chain_id_check` off unconditionally, and `create_evm` routes through it, so this is the production path too. No live impact that I can find — mega-reth never sets the flag, and its only chain-id enforcement is the pool (`transaction-pool/src/validate/eth.rs:480`).

Two things:

1. An embedder that explicitly sets `true` now loses it silently. `with_cfg_unpinned` isn't reachable from `EvmFactory`, though a post-construction write to `evm.ctx.cfg` does stick — which is what `system_tx_replay.rs` relies on. Pre-PR the field passed through. Is silently overriding an explicit setting the intent here, as opposed to honoring it or rejecting it loudly?

2. Since the pool is the only place this is currently enforced, anything reaching execution without traversing it is unchecked at every layer. Could we enforce it in the EVM from Rex7? It's still the unstable spec, so it can absorb the behavior change that flipping the gate on implies, where the frozen specs can't. I'd leave the shape to you — deposits and the REX5+ system-tx path are the two areas I'd want to see covered.

Troublor, line 424: Following up on point 2 with something concrete.

If the gate is ever enabled, the REX5+ system-tx guard at `execution.rs:113` rejects `chain_id: None` unconditionally — but revm's canonical check, which that guard's comment says it mirrors, exempts legacy and custom types. MegaETH system txs are `TxLegacy`, and mega-reth builds them by inheriting `tx.chain_id()` from the triggering user tx (`payload_executor.rs:1141`), which is `None` for a pre-155 legacy tx the pool accepts. So the guard reads as stricter than intended.

Rex7 looks like the right place to fix both together — the guard and the gate. And if it's spec-gated (`tx_chain_id_check = spec.is_enabled(REX7)`), `CfgIntent` and `with_cfg_unpinned` fall out entirely: the spec answers the "did the caller mean this or is it just the upstream default" question that the bool can't. Worth considering where that gate lives — `apply_cfg` covers all six `with_cfg` call sites in the repo, `on_new_tx` covers everything and matches how `force_amsterdam_eip8037_off` already handles the same problem.

crates/mega-evm/src/evm/context.rs
```diff
@@ -592,6 +695,8 @@ impl<DB: Database, ExtEnvs: ExternalEnvTypes> MegaContext<DB, ExtEnvs> {
     ///
     /// DB-dependent pre-frame usage may still be recorded later during pre-execution.
     pub(crate) fn on_new_tx(&mut self) {
+        force_amsterdam_eip8037_off(&mut self.inner.cfg);
```
Troublor, line 698: Non-blocking — question about the pinning strategy rather than the pin itself.

The doc on `force_amsterdam_eip8037_off` justifies settling this per-transaction rather than at entry points on two grounds I had trouble reproducing:

- The re-enable path (`set_spec_and_mainnet_gas_params` ORs the flag on for Amsterdam+) can't fire while every `MegaSpecId` maps to Prague — and `test_all_specs_map_to_isthmus_and_prague` already forces the conversation if that ever changes. If Amsterdam does get mapped, whether EIP-8037 should be on seems like a decision to make deliberately at that point rather than one pre-answered here.
- "a per-entry-point pin has to be re-derived every time one of them is rewired or a new path appears" — but those entry points are four bounded, in-crate sites, and entry-point pinning is exactly what `tx_chain_id_check` does three lines up.

Meanwhile every production path I can find — factory, the keyless-deploy sandbox (`sandbox/execution.rs:622`, the one construction outside the factory), mega-evme, t8n, state-test — funnels through `MegaEvm::new`, which already pins. Is the `on_new_tx` write covering something that one doesn't, beyond an embedder mutating `evm.ctx.cfg` on a live context? If that's a requirement it's worth stating outright; if not, a single pin would drop a per-tx config write and the second place this value is decided.

Related to the other thread: this flag and `tx_chain_id_check` currently use opposite strategies (choke point vs. entry points). However the Rex7 question lands, one agreed strategy for both would be easier to follow.

crates/mega-evm/src/evm/host.rs
```diff
@@ -149,69 +157,268 @@ impl<DB: Database, ExtEnvs: ExternalEnvTypes> Host for MegaContext<DB, ExtEnvs>
...
+    /// [`Self::access_list_preloads_account`] is the one exemption: an account this transaction's
+    /// access list pre-loads is warm from the transaction's first instruction onwards, and
+    /// re-cooling it here would charge the cold surcharge on top of the access-list fee the
+    /// sender already paid.
+    ///
+    /// Callers OR this into the `is_cold` the journal reports. That is a pricing-only correction:
+    /// the journal still takes its resident-entry branch and still records the `account_warmed`
+    /// entry that re-cools the account if the frame reverts, so nothing else about the load moves.
+    #[inline]
+    fn resident_entry_prices_cold(&self, address: &Address) -> bool {
```
Troublor, line 312: Non-blocking for this PR — a Rex7 follow-up.

The correction itself is right: I checked both versions, and revm 27's resident (`Occupied`) branch prices from the entry alone and never consults `warm_preloaded_addresses`, while revm 40 added that consultation. So this faithfully restores frozen behavior and nothing should change here.

But the behavior it preserves looks like a bug worth fixing later. Precompiles, the coinbase, and address-only access-list entries are never inserted into journal state during pre-execution — they live only in the pre-warm set, and that set is consulted exactly once, on the `Vacant` branch when the entry is first created. MegaETH's own bookkeeping consumes that one chance: `inspect_account` inserts with `mark_cold()`, and it runs before the opcode's own load in the SELFDESTRUCT wrapper (`instructions.rs:2590`), the CALL-family storage wrapper (`:2003`), and CREATE (`:2107`). Net effect: a SELFDESTRUCT to — or CALL into — a precompile or the coinbase pays the ~2,600 cold surcharge purely as an artifact of our pre-inspection, not from any intended rule. revm has a TODO next to `reset_preloaded_addresses` pointing at the same weakness.

The expected rule seems to be that pre-warm membership is final: warm regardless of what the journal holds. That's a real gas change, so Rex7 rather than here — could it be picked up alongside the other Rex7 items? Worth deciding whether the fix is general (drop the resident-entry correction from Rex7 so the pre-warm set wins everywhere) or narrow to the SELFDESTRUCT beneficiary.

tools/mutation/mutate.py
```diff
@@ -0,0 +1,2933 @@
...
+# Crash-recovery journal: written before each mutant apply, removed after restore.
+DEFAULT_JOURNAL = Path("tools/mutation/.mutate-journal.json")
+
+# Linear MegaSpecId order (must match crates/mega-evm/src/evm/spec.rs enum).
+# Enumeration hard-fails if the source enum diverges from this list.
+SPEC_ORDER: Tuple[str, ...] = (
+    "EQUIVALENCE",
+    "MINI_REX",
+    "REX",
+    "REX1",
```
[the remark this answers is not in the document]
RealiCZ: Fixed in 0072855. `SPEC_ORDER` stays the behavior-introducing ladder: an `is_enabled` gate compares behavior projections, so a swap onto `MINI_REX_1` would be a double shift in disguise (same rationale as `mutants/operators/spec-gate/generate.py`). `verify_spec_order` now parses the alias projections from `MegaSpecId::behavior()` and removes them before comparing the enum against `SPEC_ORDER`; a product gate on an alias rung stops enumeration with a pointed message instead of guessing. `--list` enumerates again (509 bodies, 103 spec-gate sites).

crates/mega-evm/src/limit/limit.rs
```diff
@@ -778,8 +778,13 @@ impl AdditionalLimit {
...
         // If frame-local, absorb it — clear the exceed flag and change to Revert so
-        // remaining gas returns to the caller. State changes are reverted by revm's
-        // Revert handling. This works at any depth including the top-level frame.
+        // remaining gas returns to the caller. This works at any depth including the
+        // top-level frame.
+        //
+        // The rewrite changes the reported result, not the journal. revm decides
+        // commit-or-revert from the frame's original instruction result, before the
+        // `FrameResult` ever reaches this hook, so a frame that ran to a successful exit is
+        // already committed and stays committed under the rewritten Revert.
```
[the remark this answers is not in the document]
RealiCZ: This is the frozen ≤REX6 behavior, not something revm 40 introduced. On both bases the child's journal decision is made inside `process_next_action` (commit on success, revert otherwise; `return_create` / `return_call` in revm-handler 18 as in 20), and only then does `run_exec_loop` hand the `FrameResult` to `frame_return_result`, where this hook runs; mega-evm's hook sat after `process_next_action` on revm 27 too. So a parent-lane exceed first seen at frame return has always rewritten the result without touching the journal. Moving the check ahead of the commit decision would change results for stable specs and break replay, so it stays as is here; the comment above the block records exactly this. The REX7 line closes the window by folding the parent-lane check into the frame verdict before the commit decision, in the REX7 branch stacked on this PR, where behavior is allowed to change.

crates/mega-evm/src/evm/instructions.rs
```diff
@@ -182,50 +191,88 @@ impl<DB: Database, ExtEnvs: ExternalEnvTypes> core::fmt::Debug for MegaInstructi
...
+                gas_table,
+                eth_spec,
+            ),
         };
         Self { spec, inner: instruction_table }
     }
 }
 
+/// Returns the static gas table the interpreter pre-charges from for `spec`.
+///
```
Troublor, line 251: **Rex7 proposal: can the volatile-access guard concede the underfunded-frame case?**

This whole mechanism — the per-spec zeroing here, `charge_static_gas!`, the `_self_charged` handler variants, and the hand-maintained `volatile_guarded_opcodes()` list — exists to keep the `disableVolatileDataAccess` guard reachable by a frame that can't afford revm 40's static pre-charge. Correct for the frozen specs, and the reasoning in the doc comments is good.

But it leaves a correspondence that has to be maintained by hand in two places, and a mismatch is silent and consensus-visible in both directions: an entry zeroed without a guard makes the opcode free, a guard whose entry stayed makes it charge twice. The test pins it today, but the list is manual.

Worth asking whether Rex7 should just let the pre-charge OOG a frame that is too poor to afford the opcode, and only run the guard when gas suffices. The conceded case is a frame holding less than 2–100 gas that was going to die anyway; in exchange the guarded opcodes go back to revm's vanilla table and the mechanism disappears for new specs.

The per-spec modules already give us the seam to do that without disturbing the frozen specs: `mod rex7` could stop delegating to `rex6::instruction_table()` and wire its own table with `gas_table(table) = table`. The existing machinery stays where it is and becomes append-never, and the duplication is scoped to the handful of opcodes that actually differ.

Not for this PR — just flagging while Rex7 is still open, since it's unfixable once frozen. There's a broader version of this too: `Gas` is a concrete field on `Interpreter`, not an associated type on `InterpreterTypes`, which is why every opcode has to be wrapped to measure compute gas at all. Might be worth an upstream ask to revm.

crates/mega-evm/src/evm/instructions.rs
```diff
@@ -949,33 +1333,95 @@ pub mod volatile_data_ext {
...
+    /// The check over-approximates on purpose — it does not reconstruct how far revm 27 would
+    /// have gotten — with one exception: a `MemoryOOG` halt is never routed here, because memory
+    /// expansion was charged before the load under revm 27 as well, so that halt shape cannot
+    /// diverge. An already-marked transaction is skipped for the same reason: the mark is
+    /// idempotent, so losing a duplicate cannot change replay.
+    #[cfg(debug_assertions)]
+    fn debug_check_frozen_detention_window<H: HostExt + ?Sized>(
+        host: &mut H,
+        opcode: u8,
+        raw_target: Option<Address>,
```
Troublor, line 1357: **Blocking: this debug-only check mutates the journal, making gas depend on the build profile.**

`best_effort_resolve_eip7702_delegate_address` takes `&mut self` and reaches `inspect_account`, whose `Vacant` arm does `entry.insert(account)` after `mark_cold()`. It pushes no journal entry, so a frame revert can't undo it — `JournalEntry::AccountWarmed`'s revert only re-cools, it never removes a state entry. The account stays resident-and-cold for the rest of the transaction.

`resident_entry_prices_cold` then prices exactly that state as cold, exempting only access-list entries that carry storage keys. So:

- **debug**: check runs → target materialized resident-cold → a later touch of it costs 2,600
- **release**: the whole check is `#[cfg(debug_assertions)]` and absent → fresh load consults `warm_precompiles` / `warm_coinbase_account` → the same touch costs 100

REX6+ only (the `||` short-circuits below that), and only for an address that would otherwise be warm — a precompile, the coinbase, or an address-only access-list entry. But a REX6 `CALL` to a precompile that runs out of gas on the static or value-transfer charge is an ordinary path, and it leaves the two builds 2,500 gas apart in both EVM gas and compute gas.

Two consequences:

1. `cargo test` is a debug build, so the whole suite exercises an EVM that charges differently from the shipped one.
2. It defeats the check's stated purpose. The doc comment says this is "the tripwire such a replay must run with" — but `debug_assert!` only fires in a debug build, and a debug build is the one whose gas is wrong. The full-history replay meant to prove no transaction sits in the window would itself diverge from history, on unrelated transactions.

The check is already documented as deliberately over-approximate, so it doesn't need to resolve anything it can't see: peeking `journal.state` and skipping when the target isn't resident would keep the detection and drop the side effect. Worth a test pinning the same gas value under both profiles either way.

Separately, three lines down: the assert message points at `REVM_40_REVIEW_GUIDE.md (wontfix #20)`, which isn't in the repo or anywhere in its git history. That message is the only instruction anyone gets at the moment this fires.

PR #366, opened 2026-08-04
feat(mega-evme): batch replay and verification tooling

RealiCZ:
## Summary

Turns `mega-evme replay` into a self-contained equivalence-verification tool: replay whole blocks or transaction lists in one process, verify every replay against its on-chain receipt, and persist reusable fixtures and RPC caches safely under concurrency.

- **Batch replay**: `--block <N>` / `--tx-file <LIST>` replay many transactions in a single process (one provider, one cache, each block executed once), emitting one NDJSON entry per target; single-transaction output stays byte-identical.
- **Receipt verification**: `--verify-receipt` fetches each target's on-chain receipt and compares status / gasUsed / logs, reporting a structured diff; mismatches drive a dedicated exit code.
- **Fixture sweep**: `--dump-fixture-dir <DIR>` bulk-dumps self-validating state-test fixtures (bench corpus format) with per-target fidelity gating.
- **Concurrent-safe caching**: cache persistence takes a file lock and merges with on-disk state, so multiple processes can share one `--rpc.cache-dir`; `cache merge` consolidates existing per-worker caches; capture envelopes get optimistic-concurrency protection for the external-env snapshot.
- **Cache capacity**: `--rpc.cache-size` is replaced by `--rpc.cache-max-entries` (`0` = effectively unlimited), so long verification runs stop silently evicting early entries.
- **Rate-limit clarity**: `--rpc.rate-limit` is renamed to `--rpc.cu-per-sec` (old name kept as a visible alias), with a warning for self-throttling values.
- **Request timeout**: `--rpc.request-timeout` (default 30 s, `0` disables) bounds every HTTP request, so a stalled endpoint fails fast and retries instead of hanging the process.
- **Exit codes and structured errors**: one documented taxonomy (`0` success / `1` execution or input error / `2` verification mismatch / `3` RPC failure); `--json` runs always end with a machine-readable error object on failure.

All changes are confined to `bin/mega-evme` and `docs/mega-evme`; no consensus code is touched.

## Testing

- Full suite (`cargo test -p mega-evme`) plus envelope-gated offline batch integration tests.
- Live mainnet verification: whole-block batch replays with `--verify-receipt` (29/29 and 25/25 receipt matches on fresh blocks), two-process concurrent cache sharing with no lost entries, and fixture sweeps self-validated through the state-test runner.
- CI-grade lint: workspace clippy under `-D warnings`, rustfmt, cargo-sort, Prettier on docs.

## Follow-up fixes (2026-08-11)

Rebased onto the current `cz/chore/upgrade-revm-40` tip (repository rules forbid force-pushes, so the base sync landed as merge commit `62e1dc0`; zero conflicts), plus three fixes:

- **Replay correctness — RPC account existence** (`36081e7`): JSON-RPC cannot express "this account was never created" (`eth_getBalance` / `eth_getTransactionCount` / `eth_getCode` all answer zero), so the forked backend materialized never-created accounts as existing empty accounts. That flipped the EIP-7702 per-authorization refund condition: a brand-new authority was judged already-in-trie and each replayed authorization refunded 12,500 gas the chain did not, making replayed `gasUsed` under-report the on-chain receipt. The forked state now maps the all-zero answer back to `None`. This is safe because an existing-but-empty account cannot occur post-EIP-161, which every chain this tool replays has had from genesis. Tests: DB-level boundary matrix (all-zero → nonexistent; balance-only / nonce-only / code-only → still existing) plus an execution-level mock-transport test asserting a type-4 transaction with a fresh authority costs exactly 12,500 gas more than with an existing authority (verified to fail without the fix).
- **Batch replay cache policy** (`409de9a`): online batch replay (`--tx-file` / `--block`) now engages the on-disk RPC cache only when `--rpc.cache-dir` is passed explicitly, and otherwise behaves as `--rpc.no-cache-file`. Rationale: a linear history scan's request keys are block-scoped and essentially never repeat across runs, so a shared cache file buys almost no hits — while the clean-exit persist re-reads, merges, and rewrites the whole file under a cross-process lock, a tail that grows linearly with the file (minutes at multi-GB sizes) and serializes concurrent batch workers into hour-long queues. Defaulting batch to no persistence makes the exit cost zero and independent of cache size, which dominates the alternative (incremental/sharded persist) that would still pay the per-process load cost and add format complexity. Single-transaction replay, `run`/`tx --fork`, and capture mode keep their previous defaults; `--rpc.no-cache-file`, `--tx-file` / `--block` / `--verify-receipt` semantics, and the exit-code taxonomy are unchanged. Tests: a batch run against a seeded default-path cache file leaves it byte-identical and creates no cache file anywhere (so a 26k-target batch exits with no cache tail at all), an explicit `--rpc.cache-dir` still persists, and single-transaction replay still persists by default.
- **`tx --raw` decoding** (`a77ed86`): `DecodedRawTx` held the real envelope yet hand-mapped every variant into a `TxEnv`; it now derives the transaction through the upstream `FromTxWithEncoded` impl (which also fills the deposit parts and the enveloped bytes for L1 fee calculation), so new transaction types are picked up with the dependency instead of a manual mapping. The CLI-flag path (`TxArgs`) has no real envelope and is unchanged. Tests: EIP-155 signed-vector decode, deposit-envelope decode, and CLI-override behavior.

## Review-fix batch + coherent `--override.spec` (2026-08-12)

22 commits (16 changes + 2 test-hardening + 4 merges), all in `bin/mega-evme/` and `docs/mega-evme/`; workspace suite 1846 green, clippy/fmt/prettier/cargo-sort clean.

- **Batch target classification is now total** over the `(blockNumber, blockHash)` resolution space: a mined answer without an inclusion hash and the contradictory null-number-with-hash shape both fail the target as `rpc` instead of being queued or misread as pending; a block-body hash that resolves to null aborts as a typed RPC inconsistency (exit 3) that still names the vanished transaction for the abort sweep.
- **The single-transaction path now matches batch on fetch coherence**: target metadata is classified exhaustively before any block fetch, the parent block must link to the fetched block, the target must anchor to it (reported inclusion hash and body membership), block-body nulls reuse the same typed rpc variant (the user-supplied hash keeps its definitive not-found, exit 1), and a pending replay fetches the latest block once and reuses it for both roles.
- **`replay --override.spec` is now a coherent what-if**: the executor receives a schedule synthesized from the forced spec (activation from the spec, per-fork parameters delegated to the chain config), so pre-block predeploys, EIP-2935/4788 gating, block-level limits, and EVM semantics all derive from one spec; without an override the behavior is byte-identical (verified by a binary-output diff probe). Replaying an old block under a newer spec deliberately installs predeploys that never existed at that height. Note: mainnet/testnet schedules currently publish only the Rex5 registry parameters, so `--override.spec Rex6`+ on those chains fails closed with a message naming the missing config (previously it crashed with a code-hash mismatch).
- **Every cache-file write now takes the sidecar lock**: `cache merge` locks its output and folds the current on-disk file in under the lock (two-process serialization tests), lock-acquisition failure fails closed instead of degrading to an unlocked write, `--rpc.clear-cache` unlinks and loads inside one critical section, and the provider persist classifies the on-disk shape before writing so it can no longer replace a capture envelope. Safeguard diagnostics (chain-identity, output replacement) now reach stderr at default verbosity.
- **Capture hygiene**: the capture transport no longer bakes `result: null` answers into fixtures — offline replay reports a cache miss naming the request instead of a frozen not-found.
- **Robustness**: the panic hook writes fallibly, so a closed stdout (`--json | head`) ends with the documented exit 1 instead of a SIGABRT; the fixture `pre`-map's absent-means-nonexistent shape is pinned by tests; raw-tx decoding gains signed EIP-2930/1559/7702 vectors with full field assertions.

Review threads fixed by this batch are resolved; the batch-mode `--rpc.clear-cache` thread stays open pending a semantics decision.

## Exit-taxonomy unification + batch-cache opt-in (2026-08-12, second batch)

11 commits, all in `bin/mega-evme/` and `docs/mega-evme/`; workspace suite 1891 green, clippy/fmt/prettier/cargo-sort clean.

- **`--rpc.clear-cache` is a disk-cache opt-in for batch replay**: the documented recovery flag now engages the cache (delete under the sidecar lock, start empty, persist on exit) instead of being silently dropped; the `--rpc.no-cache-file` combination's actual behavior (clear does not run) is documented as is.
- **Receipt handling is one contract across all modes**: dump and verify classify a null, divergent, or unfetchable receipt identically (rpc, exit 3) with byte-identical error objects, and every receipt must belong to the transaction it was requested for — a receipt served for another transaction is an RPC inconsistency in single verify, single dump, and batch alike.
- **Batch classification is order-independent and truthful**: per-target inclusion validation against the fetched block replaces the first-seen anchor; an anchored target absent from the block body is an endpoint contradiction (rpc); a non-target transaction aborting the block drives the run's exit class through a separate floor without corrupting the per-target totals; one unavailable receipt under `--verify-receipt --dump-fixture-dir` is counted once while both result fields are emitted; NDJSON output follows the documented `(block, tx_index)` order with absent targets last; a block containing none of its targets is answered without being executed.
- **Failed receipt fetches keep the replayed result**: the target's execution facts stay on its NDJSON line with the failure under `verification.error` / `fixture.error`, per the documented keep-your-result policy; the documented jq selectors cover both failure shapes.
- **Pre-block RPC failures classify correctly**: an unanswered state read inside the EIP-2935/EIP-4788 system calls exits 3 (was 1), matched at the cause boundary through named wrapper constants so an execution error merely embedding "RPC error:" text cannot misclassify. Known residue, deliberately out of scope: a keyless-deploy sandbox DB failure is collapsed to a selector-only revert inside the executor and cannot be classified from the tool side.
- The `--block 0` (invalid request, exit 1) vs endpoint-reported block-0 inclusion (contradictory endpoint data, exit 3) asymmetry is now documented rationale rather than an accident.

All four previously open review threads are fixed by this batch and resolved.

## Served-answer authentication + classification consistency (2026-08-12, third batch)

Four fixes from the Codex review round on `8a8cad0`, each with red-green regression tests:

- **Fetched transactions are authenticated before execution**: all three `eth_getTransactionByHash` consumers (batch loop, single-path preceding loop, single-path target) recompute the hash from the served consensus encoding and re-derive the sender from the signature, refusing mismatches as an inconsistent-endpoint failure (exit 3). The response's own `hash`/`from` fields are never trusted — notably, alloy's `trie_hash()`/`tx_hash()` return the *cached* server-supplied hash for RPC-deserialized transactions, so authentication hashes the encoding explicitly.
- **Hashless block aborts attribute to the in-flight transaction**: a rejection that names no hash (block-gas admission, for one) now lands on the transaction whose iteration raised it instead of sweeping the aborter itself as an unanswered peer.
- **A null endpoint-resolved block is an RPC failure**: on the single path, the replayed block and its parent are fetched at heights the endpoint itself resolved, so a null answer is the divergent-views class (exit 3), matching batch; `BlockNotFound` (exit 1) stays reserved for user-supplied heights.
- **Local clear-cache failures exit 1**: lock-acquisition and unlink failures classify as input errors, not RPC failures — retrying or switching the endpoint cannot fix the local filesystem.

Docs: the exit-code taxonomy in `overview.md` now records the authentication and resolved-block-null conventions; `state-management.md` records the clear-cache failure class. All four review threads from this round are fixed and resolved.

## Refactor: consistency infrastructure + single-format RPC cache (2026-08-14, fourth batch)

Two refactor waves, every step gated by a committed golden-equivalence baseline that was never regenerated.

**A-phase (consistency infrastructure):**
- Typed capture-doctor test helpers (`tests/common/doctor.rs`): every envelope-tampering operation carries an exactly-one-hit contract, parsed-field locators (a trap entry pins that raw-marker matching cannot come back), and a full contract matrix.
- Golden equivalence gate (`tests/replay_equivalence.rs` + `tests/golden/`): 22 argv rows × human/json over the committed captures, full-stderr pinning, exit-2/mixed-batch/keep-result rows, negative controls that run under bless too, and a bless flow that re-derives each snapshot from a second run before accepting it.
- Shared target-coherence judgments (`src/replay/coherence.rs`): the metadata four-shape classification, genesis guard, parent-linkage, inclusion-anchor, and body-membership checks now exist once as pure predicates with typed errors; both drivers adapt them locally, and every contract message is byte-identical to before.
- Single receipt admission (`verify::ReceiptEvidence`): dump+verify on the single path fetch/authenticate/anchor the receipt once instead of twice.

**B-phase (single-format RPC cache):**
- The online per-chain cache is now served from the transport layer in the same envelope format capture writes (plus a `kind: "cache"` marker; capture fixtures stay byte-identical), with a typed caching policy: JSON-RPC errors, null results, `eth_blockNumber`, moving-block-tag requests, and pending transaction metadata are never persisted — fixing the open review thread where a cached pending answer froze `--verify-receipt` until a manual `--rpc.clear-cache`.
- Warm replays now cache block bodies (previously never cached): a warm single-transaction verify drops from 3 RPC requests to 1.
- The capacity flag is an eviction threshold rather than a preallocation (a huge `--rpc.cache-max-entries` no longer balloons memory), `0` keeps its published 1,048,576-entry ceiling.
- Cache files self-identify: the body's `chain_id` is verified against the endpoint (hard error on mismatch), files the tool cannot identify are never deleted or overwritten, and the retired provider-array format self-heals with one warning (it stores only hashed keys under a different algorithm, so no migration is possible).
- The provider-array support surface (alloy `CacheLayer` persistence, `SharedCache`, dual merge paths, filename chain-id heuristics) is deleted; `cache merge` is envelope-only, checks chain identity from the body, and refuses non-envelope input with actionable guidance. Its output carries no `kind` marker, so consolidating worker caches into a seedable per-chain file is retired (documented in `cache.md`).
- Docs and `--help` text describe the single format and the real memory behavior.

## Refactor: shared mined-block execution kernel (2026-08-14, fifth batch)

The single-transaction and batch replay drivers previously each carried their own copy of the block execution walk. This batch extracts one kernel and migrates both onto it, in four equivalence-gated steps:

- **Kernel extraction** (`src/replay/kernel.rs`): fork parent state → authenticated per-transaction walk with in-flight abort attribution → early stop at the last target → finish and receipt harvest. Guards, fetching, entry assembly, and error adaptation stay driver-side.
- **Typed lifecycle**: `TargetLifecycle` (inspector as a construction generic, fallible transaction-transform and post-execution hooks) and a two-phase materialization guarantee enforced by the type system — a fixture draft can only be redeemed against a `CleanRun` token the kernel mints after a clean finish, so a draft cannot be published for a block that aborted.
- **Single mined path migrated**: trace, state dump, `--override.*`, `--dump-fixture`, and `--verify-receipt` all plug into the kernel's lifecycle points. Zero test modifications, zero golden regenerations; the surfaces the goldens did not cover were verified by a 239-file binary diff against the pre-migration build (only the execution-time line differs), and RPC wire-request counts are unchanged. The pending path deliberately keeps its dedicated adapter: its fetch-once-fill-both-roles anti-divergence semantics are a special case, not a kernel behavior.
- **Regression matrix committed**: the equivalence gate grows from 48 to 79 cases — tracer modes (opcode/call/prestate/prestate-diff, inline and to-file), state dumps, spec and transaction overrides, preceding-transaction failure, and mid-block abort — all offline, with large trace bodies pinned by normalized digest and a negative control proving the digest tracks the body it stands for.

There is now exactly one body-walk implementation in the tool. Honest accounting: `src/replay` grew 13.5% (typed API surface, the now-standalone pending adapter, and reconciliation rationale in comments — about half the growth is prose); the structural goal, not the line count, was the point, and the same-fix-twice class of drift this PR repeatedly hit during review is structurally gone for the execution core.

Two deliberate, non-behavioral deltas: the human-output execution-time span now includes fork and finish, and the fixture-refusal path now finishes the block (provably no state read, no RPC, same surfaced error).

## Review response: block-header authentication + planning-time target authentication (2026-08-18, sixth batch)

Three fixes and two documented acknowledgements from the Codex round on the refactor batches:

- **Fetched block headers are now authenticated**: the consensus-header hash is recomputed (`hash_slow`, not the served field) and compared against the endpoint's claim at every block-retrieval site in both drivers. Previously, a tampered capture or inconsistent backend could alter timestamp/gasLimit/baseFee while keeping the claimed hash and every guard passed — verified operationally against the pre-fix binary (forged gasLimit exited 0; now exits 3 naming both hashes). Mock chains in tests are now sealed bottom-up like real blocks; the capture doctor gained field-rewrite-and-reseal operations under its contract tests.
- **Batch target resolution authenticates before classifying**: a pending-looking answer served under a different transaction's hash was previously stamped as definitively pending (resolution is the one lookup that never reaches the kernel's authenticated walk); it is now an inconsistent-endpoint rpc failure.
- **`--overwrite` requires `--dump-fixture-dir`** at parse time instead of parsing successfully and silently doing nothing.
- Documented: the per-chain cache persists mined/numbered answers as final (MegaETH's single-sequencer blocks do not reorg; `--rpc.clear-cache` / `--rpc.no-cache-file` are the outs on endpoints where that assumption fails), and `--override.spec` synthesizes the schedule but never rewrites forked parent state (contracts deployed by later forks remain reachable as plain bytecode under a downgrade).

All five review threads from this round are addressed and resolved.

## Review response: height authentication + in-place merge concurrency (2026-08-18, seventh batch)

- **Fetched headers must match the requested height**: an authentic, self-consistent block served for the wrong number — including a consistently shifted parent/block pair, which passed every existing guard and silently replayed another block's body under the requested number's stamp (verified against the pre-fix binary: exit 0 with cross-stamped receipts; now exit 3 naming both heights) — is rejected at the fetch by the shared coherence judgment.
- **In-place `cache merge` can no longer roll back a concurrent writer**: an input that names the output is excluded from the pre-lock read and contributes only through the locked re-read, pinned by a two-process regression test that fails against the pre-fix binary.
- Documented: the replay trust boundary — transaction bodies and block headers are authenticated by recomputable hash; the block body listing (rebuilding the transactions root would need trailing bodies the walk deliberately never fetches after the last target) and state reads remain trusted, with `--verify-receipt` as the cross-check for both.

All three review threads from this round are addressed and resolved.

## Review response: hard-link aliases + served-hash-field invariant (2026-08-18, eighth batch)

- **Hard-linked aliases of the merge output are now detected**: file identity is compared by `(dev, ino)` rather than canonical path, closing the remaining alias route by which an in-place `cache merge` could read a stale pre-lock copy of the output.
- **The served `hash` field joins the transaction authentication invariants**: an honest payload carrying a lying `hash` field previously passed authentication (which deliberately recomputes) while the batch dump path filed the fixture under the lied name — result line reporting `H`, artifact written as `X.json`, `--overwrite` able to replace an unrelated target's fixture (verified against the pre-fix binary). Every `tx_hash()` read downstream of authentication is now a verified value.

Both review threads from this round are addressed and resolved.

## --dump commit semantics (2026-08-25, ninth batch)

`--dump` serialized the raw `EvmState` without interpreting revm's account status flags, so it described a world the commit never produces:

- **An account destroyed by `SELFDESTRUCT`** (created and erased in one transaction, EIP-6780) was printed as a live account with full code and storage, contradicting the prestate-diff tracer's post side. It is now the bare marker `{"selfdestructed": true}`, and `--prestate` reads such an entry as "this address does not exist" — so a dump stays safe to feed straight back, and the round-tripped world equals the committed one.
- **An address only observed as nonexistent** (read, or touched without ever gaining balance, nonce, or code) was printed as an existing empty account; a round-tripped prestate would then answer `EXTCODEHASH` with the empty-code hash where the chain answers zero. Such addresses are now omitted; the tombstone outranks the omission.

A class sweep found no further members: the fixture pre-state closure already maps a missing account to absence, the prestate-diff tracer is upstream-owned and correct, storage slots have no observable zero-vs-absent distinction, and the commit paths use revm's own `commit`. The tombstone predicate is the same one revm's `CacheDB::commit` uses, so dump and commit cannot drift apart on it.

Live accounts serialize byte-identically to before; the equivalence goldens are untouched.


Troublor (approved):

bin/mega-evme/src/cache/merge.rs
```diff
@@ -0,0 +1,776 @@
...
+                    path.display(),
+                    loaded_env = loaded_c,
+                    ours_env = ours_c,
+                    on_disk_env = disk_c,
+                )))
+            }
+        }
+        (Some(_), _) => Ok(ours_c),
+        (None, disk) => Ok(disk.clone()),
+    }
```
flyq, line 372: **[MAJOR] `ours == loaded` is not a conflict — this discards the whole run's captured RPC responses.**

This branches only on `ours != on_disk`, then accepts only when `on_disk == loaded`. But a capture run given no `--bucket-capacity` carries the loaded snapshot forward verbatim (`replay/cmd.rs:557-580`: `effective` falls back to `pctx.external_env`), so `ours == loaded` — this process expressed **no opinion**. If a sibling legitimately refreshed the on-disk snapshot between our load and our persist, we hit `ours != disk` *and* `disk != loaded` and hard-fail.

I confirmed this with a throwaway test; the message is self-refuting:

```
Conflicting external_env snapshots when merging 'capture.json':
  loaded  Some(ExternalEnvDoc { bucket_capacities: [(1, 10)] }),
  ours    Some(ExternalEnvDoc { bucket_capacities: [(1, 10)] }),
  on-disk Some(ExternalEnvDoc { bucket_capacities: [(1, 20)] })
```

`loaded` and `ours` are identical. Because capture persistence is a hard error by design (`cache_store.rs:188`), this propagates through `CacheFileEnvelope::save` → `RpcCacheStore::persist` and **every RPC response the run captured is thrown away**. Bucket capacities are external-env metadata; the cache entries were never in conflict with anything.

Suggested fix — add an arm before the OCC check:

```rust
match (&ours_c, &disk_c) {
    // We carried the loaded snapshot forward, i.e. expressed no opinion:
    // a sibling's refresh wins and our cache entries still merge.
    (Some(_), Some(_)) if ours_c == loaded_c && ours_c != disk_c => Ok(disk_c),
    (Some(o), Some(d)) if o != d => { /* existing OCC check */ }
    (Some(_), _) => Ok(ours_c),
    (None, disk) => Ok(disk.clone()),
}
```

No test covers `ours == loaded != disk` — the closest, `test_merge_envelope_for_persist_rejects_true_concurrent_conflict`, uses three distinct values. Please add that case with the fix.

bin/mega-evme/src/replay/batch.rs
```diff
@@ -0,0 +1,1548 @@
...
+/// than a construction failure.
+fn is_unsupported_fixture_shape(message: &str) -> bool {
+    message.contains("does not support deposit")
+        || message.contains("does not support EIP-7702")
+        || message.contains("has no fixture mapping")
+        || message.contains("reports no gas price")
+        // Fidelity is normally gated before `build_draft`; keep the classification
+        // if a fidelity check inside the builder ever surfaces here.
+        || message.contains("does not reproduce on-chain execution")
+}
```
flyq: **[MEDIUM] Substring-matching error prose to classify skip-vs-fail — and it already misses one of the three messages it claims to cover.**

The matched strings live in `fixture.rs:187`, `:192`, `:214`, `:482`, `:105`/`:113`. But `check_fidelity`'s **third** rejection (logs-root mismatch, `fixture.rs:120-125`) reads *"the local replay emits different logs than the chain"* and contains none of them — so it would classify as a run-failing construction error, not the skip the comment on line 1084-1086 says this entry exists to guarantee. Not live today because batch pre-gates fidelity at `:998-1002`, but it is exactly the failure mode this pattern invites, already present as written.

This is the one place the PR reaches for stringly-typed classification; the envelope path deliberately went the other way (`EnvelopeReread`, `cache/merge.rs:182-190`, with `test_envelope_persist_degrades_on_corrupt_disk_path_containing_chain_id` pinning the anti-pattern).

Suggested fix: give `build_draft` a typed rejection (`Err(FixtureBuildError::Unsupported(reason))` vs `::Construction(e)`), or at minimum hoist the messages into shared `const`s so source and test cannot diverge.

bin/mega-evme/src/replay/batch.rs
```diff
@@ -0,0 +1,1548 @@
...
+    /// Unsupported shapes remain skips; database/construction failures become
+    /// fixture errors so the run exits non-zero.
+    #[test]
+    fn test_fixture_build_err_classifies_skips_vs_construction_errors() {
+        let deposit = fixture_report_from_build_err(ReplayError::Other(
+            "--dump-fixture does not support deposit transactions".into(),
+        ));
+        assert!(deposit.skipped.is_some(), "deposit is a skip: {deposit:?}");
+        assert!(deposit.error.is_none());
+
```
flyq: **[MEDIUM] These tests cannot catch drift in the classification they pin.**

The expected messages are hand-constructed here rather than obtained by driving `build_draft`. Reword the source message in `fixture.rs` and this test still passes while the classification silently flips — at which point every `--block N --dump-fixture-dir` sweep exits `1` on the index-0 deposit that every OP-stack block carries, instead of skipping it.

Either drive the real builder, or share the strings as `const`s between `fixture.rs` and `is_unsupported_fixture_shape` (paired with the typed-error fix above).

bin/mega-evme/src/replay/fixture.rs
```diff
@@ -79,6 +79,63 @@ const DEPOSIT_TX_TYPE: u8 = 0x7e;
...
+        ));
+    }
+    let actual_logs_root = state_test::utils::log_rlp_hash(result.logs());
+    if actual_logs_root != anchor.logs_root {
+        return Err(format!(
+            "replay logs root {actual_logs_root} != on-chain receipt logs root {}: the \
+             local replay emits different logs than the chain for chain {chain_id} \
+             (same gas/status, different log contents)",
+            anchor.logs_root
+        ));
```
flyq, line 159: **[MEDIUM] This message is not covered by `is_unsupported_fixture_shape`.**

The other two `check_fidelity` rejections (`:105`, `:113`) both say *"does not reproduce on-chain execution"*, which `batch.rs:1086` matches on. This one says *"emits different logs than the chain"* instead, so if a fidelity failure ever surfaces from inside `build_draft` (the case `batch.rs:1084-1086` explicitly says it is defending against), a logs-root divergence would be reported as a run-failing construction error rather than a skip.

Either add this wording to the match list, or — better — replace the whole substring match with a typed rejection.

bin/mega-evme/src/common/provider/mod.rs
```diff
@@ -75,31 +76,33 @@ pub struct RpcArgs {
...
-    /// Maximum number of items to keep in the in-memory RPC LRU cache.
-    /// Set to 0 to disable the cache layer entirely.
-    #[arg(id = "cache_size", long = "rpc.cache-size", default_value_t = 10_000)]
-    pub cache_size: u32,
+    /// Maximum number of items in the in-memory RPC LRU cache (and therefore what
+    /// gets persisted to the cache file). `0` = effectively unlimited (caps at
+    /// 1,048,576 entries; the cache index is preallocated proportional to the
+    /// cap). Default is `0`.
+    #[arg(id = "cache_max_entries", long = "rpc.cache-max-entries", default_value_t = 0)]
+    pub cache_max_entries: u32,
```
flyq, line 110: **[MEDIUM] Breaking CLI change, but the PR is labeled `api:compatible` and there is no migration note.**

`--rpc.cache-size` is removed with no alias. Failing loudly is the **right** call — `0` inverted meaning (was "disable the cache", is now "unlimited"), so an alias would have silently done the opposite of what a user's script asked — and `test_rpc_args_rejects_removed_cache_size_flag` pins the deletion properly.

But: (a) the `api:compatible` label should be `api:breaking`, and (b) `docs/mega-evme/configuration/state-management.md` gained the new row without a "removed flags / migration" note, so a user whose script dies with a clap usage error gets no guidance on what to change it to.

bin/mega-evme/src/common/provider/mod.rs
```diff
@@ -152,38 +164,27 @@ impl RpcArgs {
...
             Some(resolve_cache_path(self.cache_dir.as_deref(), chain_id)?)
         };
 
-        // 4. Build the cache layer and (optionally) the disk store.
-        let cache_layer = CacheLayer::new(self.cache_size);
+        // 3. Build the cache layer and (optionally) the disk store.
+        // Cache layer is always installed; 0 max entries maps to
+        // EFFECTIVELY_UNLIMITED_CACHE_ENTRIES.
+        let max_items = cache_max_entries_capacity(self.cache_max_entries);
+        let cache_layer = CacheLayer::new(max_items);
```
flyq: **[MEDIUM] The "disable the in-memory cache" capability is gone with no replacement.**

The old `build_provider` had a `cache_size == 0` fast path that skipped the cache layer entirely. The layer is now always installed, and `--rpc.no-cache-file` only disables *disk* persistence — the `--rpc.no-cache-file` doc row was edited to drop the old `--rpc.cache-size 0` pointer without noting that the capability disappeared along with it.

Either restore a way to disable the layer, or state explicitly in the docs that it cannot be disabled.

bin/mega-evme/src/common/provider/mod.rs
```diff
@@ -429,12 +512,35 @@ fn parse_non_empty_path(s: &str) -> std::result::Result<PathBuf, String> {
...
+/// Cap used when `--rpc.cache-max-entries 0` ("effectively unlimited") is requested.
+///
+/// Alloy's `SharedCache` preallocates its LRU hash table to full capacity
+/// (`lru::LruCache::with_hasher` → `HashMap::with_capacity_and_hasher`), so a true
+/// `u32::MAX` mapping would preallocate ~2^33 hash buckets and balloon RSS by multi-GB
+/// on every default-config online run. 2^20 entries preallocates ~2^21 pointer-sized
+/// buckets (tens of MB) while covering ~2,600 blocks' worth of RPC entries per process
+/// (a full mainnet block is ~200 entries; the largest real merged corpus to date was
+/// 15,294).
+const EFFECTIVELY_UNLIMITED_CACHE_ENTRIES: u32 = 1_048_576;
```
flyq, line 570: **[MINOR / judgment call] This default is now paid by every online run, not just the sweeps it was chosen for.**

Default moved from a 10,000-entry LRU to `0` → 1,048,576. Since `SharedCache` preallocates the LRU index, that is ≈32 MiB of bucket array plus ≈2 MiB of eagerly-memset control bytes on **every** `run --fork` / `tx --fork` / online `replay`, not only long verification runs.

The rationale here is honest and `test_cache_layer_constructs_at_effectively_unlimited_capacity` proves it constructs fast, so this is a judgment call rather than a defect — but consider defaulting to ~100k with `0` still meaning unlimited for sweeps.

bin/mega-evme/Cargo.toml
```diff
@@ -48,6 +48,7 @@ op-alloy-rpc-types.workspace = true
 # misc
 clap = { workspace = true, features = ["default", "env"] }
 dirs.workspace = true
+fs2 = { version = "0.4", default-features = false }
```
flyq: **[MINOR] `fs2` is unmaintained and this workspace's MSRV makes it unnecessary.**

`fs2 0.4.3` was last released in 2018 and its repository is archived. It is pulled in for exactly one call (`cache_store.rs:257`) and drags `winapi 0.3` plus its two `*-pc-windows-gnu` stub crates into `Cargo.lock`.

Workspace MSRV is `1.94` (root `Cargo.toml:20`) and `std::fs::File::lock()` has been stable since **1.89** with the same blocking-exclusive-advisory semantics. Dropping this line and using std removes one unmaintained dependency and four lock-file entries with no behavior change.

(Also: `default-features = false` is inert here — `fs2` has no default features.)

bin/mega-evme/src/common/provider/cache_store.rs
```diff
@@ -189,40 +235,106 @@ impl fmt::Debug for RpcCacheStore {
...
+fn acquire_exclusive_lock(target: &Path) -> std::io::Result<ExclusiveFileLock> {
+    let lock_path = lock_sidecar_path(target);
+    if let Some(parent) = lock_path.parent() {
+        fs::create_dir_all(parent)?;
+    }
+    // truncate(false): the sidecar is only a flock target; keep any existing bytes.
+    let file =
+        OpenOptions::new().create(true).read(true).write(true).truncate(false).open(&lock_path)?;
+    file.lock_exclusive()?;
+    Ok(ExclusiveFileLock { _file: file })
```
flyq: **[MINOR] This is the only `fs2` call site — `std::fs::File::lock()` covers it.**

```rust
-    use fs2::FileExt as _;   // line 28
-    file.lock_exclusive()?;
+    file.lock()?;
```

Stable since Rust 1.89; workspace MSRV is 1.94. Same blocking exclusive advisory lock, and it lets `fs2` + `winapi` come out of the dependency tree entirely.

bin/mega-evme/src/replay/batch.rs
```diff
@@ -0,0 +1,1548 @@
...
+    P: Provider<op_alloy_network::Optimism> + Clone + std::fmt::Debug,
+{
+    let BlockJob { number, block, targets } = job;
+    let verify_receipt = report.verify_receipt;
+    let dump_dir = report.dump_fixture_dir.as_deref();
+    let overwrite = report.overwrite;
+
+    if number == 0 {
+        return fail_all(&targets, BatchErrorKind::Rpc, "Block 0 has no parent block to fork from");
+    }
```
flyq: **[MINOR] `--block 0` silently succeeds with completely empty output.**

For block 0, `targets` is empty (genesis has no transactions), so `fail_all` (`:1189-1191`) returns `[]` — nothing is emitted, the tally stays clean, and the run exits `0`. The error message on this line never reaches the user.

The same shape applies to any empty block: `--block N --json` on a zero-transaction block prints nothing and exits 0, which is awkward next to the taxonomy's promise that a machine-readable run always ends with something on stdout.

Suggest rejecting `--block 0` in `validate()`, and emitting a notice for the genuinely-empty-block case.

bin/mega-evme/src/replay/batch.rs
```diff
@@ -0,0 +1,1548 @@
...
+//! that way pays the provider/cache setup once per transaction, which dominates
+//! the actual EVM work. This module reuses one provider and one RPC cache for the
+//! whole run, groups the requested transactions by their containing block, and
+//! executes each block exactly once while recording the result of every target it
+//! passes through.
+//!
+//! Every RPC call issued here has the same shape as the single-transaction path
+//! (`eth_getTransactionByHash`, `eth_getBlockByNumber` with hash-only bodies, and
+//! the state reads behind [`EvmeState::new_forked`]), so an offline envelope
+//! captured by single-transaction replays serves batch runs without a miss.
```
flyq: **[MINOR] This claim doesn't hold once `--verify-receipt` / `--dump-fixture-dir` are in play.**

True for plain replay. But those two flags fetch `eth_getTransactionReceipt` for **every** transaction of the block (`fetch_target_receipts`, `:578-583`), which single-transaction captures never fetched for non-targets. Offline, those come back as `rpc` error entries and the run exits `3`.

Worth one qualifying sentence here and in `docs/mega-evme/commands/replay.md`, which repeats the same claim.

bin/mega-evme/src/common/provider/cache_store.rs
```diff
@@ -8,19 +8,36 @@
...
+    cache::{
+        lock_sidecar_path, merge_envelope_for_persist, merge_kv_entries, read_provider_cache,
+        reread_envelope_for_merge, write_bytes_atomic, write_envelope_atomic, CacheKv, EnvelopeDoc,
+        EnvelopeReread, ExternalEnvDoc, ENVELOPE_VERSION,
+    },
+    common::{EvmeError, Result},
+};
 
 /// Clean-exit cache persistence handle.
 ///
```
flyq, line 53: **[MINOR] Stale reference on line 45 below.**

The doc says a store may be no-op because of "`--rpc.cache-size 0`". That flag no longer exists, and `--rpc.cache-max-entries 0` does not produce a no-op store — it means *unlimited*. Worth updating to just "non-fork run or `--rpc.no-cache-file`".

bin/mega-evme/src/common/exit.rs
```diff
@@ -0,0 +1,435 @@
+//! Central exit-code taxonomy for the `mega-evme` CLI.
+//!
+//! Verification pipelines branch on the process status, so every failure the
+//! CLI can reach maps onto exactly one documented code, and every exit flows
+//! through this module: the binary hands its top-level result to
+//! [`report_command_result`] and returns the code it produces. No other code
+//! path calls `std::process::exit`.
```
flyq: **[MINOR] `set_thread_panic_hook` also calls `std::process::exit`** (`lib.rs:53`), so "No other code path calls `std::process::exit`" isn't quite accurate. Reword to exempt the panic hook — the invariant it is really asserting (every *command result* becomes a status here) still holds.

bin/mega-evme/tests/replay_batch.rs
```diff
@@ -0,0 +1,628 @@
...
+//!
+//! These run against an RPC capture envelope large enough to replay whole
+//! blocks, which is too big to commit; point `MEGA_EVME_TEST_ENVELOPE` at one
+//! and run them explicitly:
+//!
+//! ```bash
+//! MEGA_EVME_TEST_ENVELOPE=<path> cargo test -p mega-evme -- --ignored
+//! ```
+//!
+//! They are `#[ignore]`d so CI, which has no envelope, skips them.
```
flyq: **[MINOR] Test coverage: every CI-visible batch test is a single-target run.**

CI coverage is genuinely good — the 20 `replay_verify.rs` tests run fully offline against the committed capture and do exercise the deferred-fixture lifecycle, the abort-before-finish path, both `--overwrite` behaviors, the fidelity gate, and (because the committed capture's target sits at block index 1) the receipt `offset`/`commit_index` mapping and the `last_target_index` early break.

But the committed capture holds exactly one receipt, so all CI batch tests are single-target `--tx-file` runs, and these paths live only in this `#[ignore]`d file:

- more than one target in one block (`pending` with >1 entry, per-target `commit_index` fan-out, entry ordering)
- `--block N` whole-block mode at all
- `resolve_targets` grouping across multiple blocks (`BTreeMap<u64, Vec<B256>>` with >1 key)
- the unreported/swept-target append logic with targets on both sides of a mid-block abort
- `first_log_index` accumulation against real logs (the committed capture's target is log-less, as its own test notes)

A committed 3–5-transaction capture over one small block would close most of this — worth its bytes for a 1,500-line driver whose entire premise is multi-transaction execution.

bin/mega-evme/src/replay/batch.rs
```diff
@@ -0,0 +1,1548 @@
...
+
+/// The kind reported for a target swept up by an abort caused elsewhere.
+///
+/// The abort says nothing about this target: whatever class caused the block to
+/// stop (unknown hash, RPC failure, executor/setup error on another
+/// transaction), a non-aborting swept target is unanswered (`rpc`). Only the
+/// transaction that caused the abort keeps its own classified kind.
+fn swept_kind(_err: &ReplayError) -> BatchErrorKind {
+    BatchErrorKind::Rpc
+}
```
flyq: **[MINOR] Unused parameter, plus a consequence worth surfacing in the PR body.**

The parameter is always ignored. It's documented and tested, but a reader still hits a function whose argument does nothing — either drop it (the call site at `:918` already has the error for the message) or add one line saying the signature exists to keep the decision reviewable.

The consequence is the more interesting part: when a **non-target** transaction aborts the block with an execution-class failure, `aborting_tx_hash` doesn't match any target, so every target is swept as `rpc` and the run exits `3` — "the question went unanswered, retrying may help" — for a deterministic failure that retrying will not fix. Defensible under the "nothing was established about this target" rule, but it means the exit code can understate. Worth a sentence in the docs' exit-status section.

flyq (commented):
## Review — batch replay and verification tooling

Reviewed the full diff against the branch checkout. Locally: `cargo clippy -p mega-evme --lib --tests --all-features` clean, `cargo test -p mega-evme` **all green** (114 unit + 20 `replay_verify` + 12 `exit_codes` + `provider`/`state`/`replay_dump`; the 13 `replay_batch` tests are `#[ignore]`d), `cargo fmt --all --check` and `cargo sort` clean.

**Verdict: close to mergeable.** One real defect (finding 1) throws away a whole run's captured RPC responses in a case that is neither a conflict nor rare; one fragility (finding 2) is already half-broken as written; one labeling/migration gap (finding 3) needs a decision. Everything else is minor. Specifics are inline on the relevant lines.

### What I verified

1. **Deferred-fixture lifecycle is correct on every path.** Drafts are built pre-commit (same moment as the single-tx dump), carried as `DeferredFixture::Ready`, and written only after `finish()` succeeds. A `finish()` failure drops ready drafts unwritten; a mid-block abort after a target built a draft turns it into an error report rather than publishing; `persist_noclobber` enforces `--overwrite=false` against a concurrent creator at publish time, with the prep-time `exists()` check correctly documented as a fast path only. A failed dump stays on the target's own result line so its receipt and verdict survive, and fails the run through the tally instead.
2. **Receipt index mapping holds.** `offset = receipts.len().saturating_sub(committed)` indexes from the end so a pre-transaction receipt can't shift the mapping, and the `first_log_index` slice is provably in bounds because `receipts.get(offset + commit_index)` is checked first. The single-tx path's `.rev().skip(1)` sum is the same quantity computed the other way.
3. **Parent-linkage and inclusion guards are real, not decorative.** The parent-hash check before forking, and the shared `check_inclusion` between `--verify-receipt` and `--dump-fixture`, mean a reorged or unanchored receipt is an `rpc` entry and never a mismatch — the distinction the whole exit taxonomy rests on.
4. **Verification is a pure function over `ReceiptFacts`, read through one accessor set for both sides**, so the two receipts can't be read with different semantics. Argument order (`onchain` first) is correct at both call sites. All mismatched dimensions are reported at once rather than short-circuiting.
5. **Exit taxonomy is exhaustive and can't produce 0 from a failure.** `from_evme_error` matches every variant explicitly; an all-zero `BatchFailureCounts` maps to `ExecutionError` rather than falling through to success. Batch precedence is resolved from typed counts, not by parsing a message. `database_cause`'s downcast chain recovers an RPC failure from inside a block execution error without inspecting rendered text — nice.
6. **Deferring cache persistence past the run error is deliberate and right**: a run that failed is exactly the one you want to debug offline, and a persist failure is reported on stderr without stealing the exit code.
7. **`TestExternalEnvs` sharing is benign.** It's `Rc<RefCell<…>>`-backed but `bucket_capacities()`/`oracle_storage()` return *configured* values, not accumulated accesses — so cloning it per block can't leak block N's accesses into block N+1's `megaEnv`, and computing `mega_env` before execution is correct in both paths.

### Suggested path to merge

1. Fix finding 1 (`ours == loaded` is not a conflict) and add the missing OCC test — the only finding that loses user data.
2. Address finding 2 with a typed `build_draft` rejection (or shared consts + a test that drives the real builder), and close the logs-root gap while you're there.
3. Settle finding 3: relabel `api:breaking`, add the migration note, decide whether "disable the in-memory cache" comes back.
4. Land the remaining minors — all one-to-four-line changes.
5. Commit a small multi-transaction capture to lift the batch driver's core paths out of the `#[ignore]`d suite.
6. Merge order: this is stacked on `cz/chore/upgrade-revm-40` (#365), still `[WIP]` with open frozen-spec findings. Nothing here depends on those, but this can't land before its base — worth confirming whether it should be retargeted at `main` if #365 slips.

bin/mega-evme/src/replay/batch.rs
```diff
@@ -0,0 +1,1548 @@
...
+                requested = hashes.len(),
+                blocks = jobs.len(),
+                unresolved = failures.len(),
+                "Batch replay of a transaction list",
+            );
+            for failure in failures {
+                let entry = BatchEntry::Failed(failure);
+                tally.record(&entry);
+                emit(&entry, report.json);
+            }
```
flyq, line 632: **[MINOR] Unresolved hashes are emitted before all block results, so the stream isn't globally ordered by input position.**

`docs/mega-evme/commands/replay.md` says results come in ascending `(block, tx_index)` and that a hash the block does not contain "is reported last, in input order" — true *within* a block, but these `resolve_targets` failures are emitted ahead of every block. One clarifying sentence in the docs.

bin/mega-evme/src/replay/batch.rs
```diff
@@ -0,0 +1,1548 @@
...
+/// hashes that could not be resolved (in the order they were requested).
+async fn resolve_targets<P>(provider: &P, hashes: &[B256]) -> (Vec<BlockJob>, Vec<FailedTx>)
+where
+    P: Provider<op_alloy_network::Optimism>,
+{
+    let mut grouped: BTreeMap<u64, Vec<B256>> = BTreeMap::new();
+    let mut failures = Vec::new();
+
+    for hash in hashes {
+        match provider.get_transaction_by_hash(*hash).await {
```
flyq, line 699: **[MINOR] Every `--tx-file` target is fetched twice.**

`resolve_targets` fetches each requested hash here, then `replay_block` re-fetches every transaction of the block — including these same targets — at `:680`. Alloy's provider `CacheLayer` covers a fixed method list; if `eth_getTransactionByHash` isn't on it, this doubles the calls against a live endpoint for the whole tx-list path.

Worth confirming, and if so, threading the already-fetched `Transaction` through `BlockJob` so the second fetch is skipped.

bin/mega-evme/src/common/outcome.rs
```diff
@@ -81,17 +86,24 @@ pub fn op_receipt_to_tx_receipt(
...
     let inner = receipt.clone().map_logs(|log| {
         let log = alloy_rpc_types_eth::Log {
             inner: log,
-            block_hash: None,
+            block_hash,
             block_number: Some(block_number),
             block_timestamp: Some(block_timestamp),
-            transaction_hash: Some(B256::ZERO),
-            transaction_index: Some(0),
+            transaction_hash: stamped_tx_hash,
```
flyq, line 105: **[MINOR] This changes `replay --json` output shape for any transaction with logs.**

Inner logs now carry the real `block_hash` / `transaction_hash` / `transaction_index` and a **block-global** `log_index`, where before they were `null` / `B256::ZERO` / `0` / per-transaction index. This is a fix and it's well covered by the two new unit tests, but it changes what an existing `replay --json` consumer parses and isn't called out in the PR body or the docs' output section. Worth a release-note line.

bin/mega-evme/src/cache/merge.rs
```diff
@@ -0,0 +1,776 @@
...
+    tmp.flush()?;
+    tmp.persist(path).map_err(|e| {
+        std::io::Error::other(format!(
+            "failed to rename temp file into {}: {}",
+            path.display(),
+            e.error,
+        ))
+    })?;
+    Ok(())
+}
```
flyq, line 529: **[MINOR] No `fsync` before rename.**

`flush()` clears the userspace buffer but doesn't force the data out, so a crash between write and rename can publish a truncated file under the target name. Fine for a cache; for a fixture that a benchmark corpus depends on (`fixture.rs:320`, same pattern), a `tmp.as_file().sync_all()` before persist is cheap insurance.

bin/mega-evme/src/replay/batch.rs
```diff
@@ -0,0 +1,1550 @@
...
+        parent_block.hash(),
+        block.header.parent_beacon_block_root(),
+        block.header.extra_data().clone(),
+        block_limits,
+    );
+
+    info!(block = number, fork_block = parent_block.header.number(), "Forking state for block");
+    let mut database = match EvmeState::new_forked(
+        provider.clone(),
+        Some(parent_block.header.number()),
```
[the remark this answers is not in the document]
RealiCZ: Acknowledged, and left as a documented limitation rather than fixed here.

The guard does only validate the headers, as you say: the fork opens state by `parent_block.header.number()`, so an endpoint serving headers and state from different backends can still hand back state for a different block at that height.

Anchoring the state reads would mean forking by hash instead of number, and that has a blast radius well beyond this change. alloy hashes the block id into its cache key — `params_hash` formats `BlockId::Hash` and `BlockId::Number` differently (`alloy-provider-2.0.5/src/layers/cache.rs`) — so switching would change every cached RPC key and stop every committed offline capture from resolving, including the whole-block one this PR adds. It also depends on the endpoint supporting EIP-1898 state queries by hash, and `EvmeState::new_forked` is shared with `run --fork` / `tx --fork`.

The comment above the two guards now states what they do and do not cover, so the next reader does not mistake header validation for state anchoring.

bin/mega-evme/src/common/provider/cache_store.rs
```diff
@@ -189,40 +234,107 @@ impl fmt::Debug for RpcCacheStore {
...
+            warn!(
+                path = %target.display(),
+                error = %err,
+                "Failed to re-read on-disk RPC cache during merge; persisting our entries only",
+            );
+            Vec::new()
+        }
+    };
+
+    let merged = merge_kv_entries(disk_entries, our_entries);
```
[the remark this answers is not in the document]
RealiCZ: Confirmed and fixed in 6f1f8b5.

The union is now truncated to this run's cap before writing, keeping this run's entries and letting on-disk entries fill whatever room is left — they are already LRU-bounded by the same cap and are the ones the run just proved it needs.

Worth recording how nearly this slipped through: the first version only had a unit test calling the merge with an explicit cap, which passes even if the call site stops passing one. The store-level test added alongside it does catch that — with the cap removed at the call site it fails with 24 entries against a configured maximum of 4.

bin/mega-evme/src/replay/cmd.rs
```diff
@@ -100,24 +156,77 @@ pub(super) struct ReplayOutcome {
...
+        // Execute, report, and (for --dump-fixture) finalize/write — but defer
+        // error propagation until the cache store has persisted: in capture mode
+        // an execution or dump-gate failure is exactly the case you'd want to
+        // debug offline, so the captured RPC responses must not be discarded.
+        let run_result = match &mode {
+            ReplayMode::Single(tx_hash) => self.run_single(&mut pctx, *tx_hash).await,
+            ReplayMode::Batch(batch_mode) => self.run_batch(&mut pctx, batch_mode).await,
+        };
+
+        let persist_result = pctx.cache_store.persist();
```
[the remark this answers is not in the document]
RealiCZ: Checked this one against the alloy source and the premise does not hold for the version in the lock file, so no change.

`CacheLayer::get_transaction_by_hash` only writes to the cache when the response is `Some` (`alloy-provider-2.0.5/src/layers/cache.rs`):

```rust
let result = client.request(req.method(), req.params()).await?;

if let Some(ref tx) = result {
    let json_str = serde_json::to_string(tx).map_err(TransportErrorKind::custom)?;
    let hash = req.params_hash()?;
    let _ = cache.put(hash, json_str);
}
```

A `null` result is never stored, so replaying a hash before the endpoint can serve it does not record the miss and later runs re-ask rather than reporting `TransactionNotFound` from disk.

Persisting past a failed run is deliberate for the other half of the reasoning: a run that failed is exactly the one worth debugging offline, and for `--rpc.capture-file` the capture is the requested artifact.

bin/mega-evme/src/replay/batch.rs
```diff
@@ -0,0 +1,1550 @@
...
+                kind: BatchErrorKind::Rpc,
+                message: format!("Failed to fetch transaction: {e}"),
+            }),
+            Ok(None) => failures.push(FailedTx {
+                tx_hash: *hash,
+                kind: BatchErrorKind::NotFound,
+                message: "Transaction not found".to_string(),
+            }),
+            Ok(Some(tx)) => match tx.block_number {
+                Some(number) => grouped.entry(number).or_default().push(*hash),
```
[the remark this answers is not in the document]
RealiCZ: Confirmed and fixed in 003c2e2.

`resolve_targets` now carries the inclusion hash reported by `eth_getTransactionByHash` into the job, and the block is rejected before it runs when its hash does not match — the same shape as the parent-linkage guard, reported as `rpc` so it stays an unanswered question rather than a wrong answer.

One adjacent case fell out of it: two targets grouped under the same block number can themselves report different inclusion hashes, which is the same divergence seen one step earlier. The disagreeing target is now reported as unanswered instead of being replayed against whichever view happened to be recorded first.

flyq (approved):

vincent-k2026:
No blocking findings from this pass. 34k lines / 137 files, all under `bin/mega-evme/` and `docs/mega-evme/`, and for a verification tool the fail-loud discipline holds up.

Things I verified myself rather than taking from the PR body:

| Claim | Result |
|---|---|
| Nothing outside `bin/mega-evme/` + `docs/mega-evme/` | ✅ all 137 paths, filter comes back empty |
| `--rpc.cache-max-entries 0` has an actual ceiling | ✅ `provider/mod.rs:570`, `EFFECTIVELY_UNLIMITED_CACHE_ENTRIES = 1_048_576`, with the derivation in the constant's own comment (~200 entries per mainnet block → ~2,600 blocks) |
| Docs match the code | ✅ `tx.md:207` and `state-management.md:326` both say "capped at 1,048,576 entries" — `0` is not documented as truly unlimited |
| `--rpc.request-timeout` covers every HTTP request | ✅ exactly one `reqwest::Client::builder()` in the tool (`provider/mod.rs:457`), with `.timeout()` and `.connect_timeout()` both set |

Also checked the two headline fixes for red-green discrimination and the mechanisms hold up: reverting the EIP-7702 fresh-authority `None` mapping makes both scenarios materialize the authority as existing, so the `+12_500` assertion fails; reverting the served-hash authentication lets the batch dump write a fixture under `<LYING_TX_HASH>.json`, so the file-existence assertion fails. I reasoned these from the code rather than executing them — if you want it as merge evidence, a revert-then-red run pasted here would close it.

One design decision worth calling out as right: when `--rpc.cache-size` became `--rpc.cache-max-entries`, `0` inverted meaning (disable → effectively unlimited), and you chose to **reject the old flag rather than alias it** so a script fails loudly instead of silently doing the opposite. That's the correct handling of an inverted flag.

## Non-blocking, process

This PR accumulated four rounds of "review-fix batch" in the body (22 commits, then 11, then 4 fixes) across 137 files. At this size it has stopped being a reviewable unit. For the next comparable piece of work, four PRs — batch replay / receipt verification / concurrent-safe caching / exit-code taxonomy — would each be independently reviewable; the coupling between them is low.

PR #369, opened 2026-08-25
feat: expose sequencer registry pending changes helper

milando12:
## Summary

- Expose the existing typed SequencerRegistry pending-change helper for external pre-block executors.
- Document that callers invoke it only for known-due changes and commit its returned state through their normal pipeline.
- Add an integration test for applying a v1-scheduled rotation under Rex6 before the separate v2 deployment.

## Testing

- cargo fmt --all --check
- cargo check -p mega-evm
- cargo test -p mega-evm
- cargo clippy --workspace --lib --examples --tests --benches --all-features --locked
- cherry-pick and focused test on current upstream main

The branch is rooted at v1.7.0 so downstream users can pin this API without Rex7 changes.


milando12:
Closing because Tenderly implemented the adapter entirely in tdly-megaeth against released v1.7.0, so this upstream change is no longer needed.

PR #370, opened 2026-09-03
feat: add a hook channel into keyless sandbox execution

RealiCZ:
## Summary

`KeylessDeploy` (`0x6342…0003`) runs the wrapped deployment in a nested sandbox EVM, which is invisible to the parent EVM's inspector: tracers stop at the intercepted CALL. This PR adds a hook channel into that sandbox, `SandboxInspector`, whose signatures match revm's `Inspector` on the sandbox EVM, plus the tracing glue that lets `mega-evme` and node tracing RPCs render sandbox frames under the `KeylessDeploy` CALL.

The no-hook path is unchanged: with nothing attached, sandbox execution is byte-for-byte the same as before. Attaching a non-intervening hook leaves result, state, gas, and resource usage identical to the unattached run. Interventions are node-local and non-consensus.

## What changed

- `sandbox::inspector` — `SandboxInspector<E>`: object-safe, all hooks default-empty / `None`, nine hooks that mirror revm-inspector 8.1.0's `Inspector` exactly (`&mut` inputs, override return values) so `CALL`/`CREATE` can be short-circuited and outcomes rewritten inside the sandbox as on a top-level EVM, plus a paired `sandbox_start` / `sandbox_end` lifecycle carrying `SandboxStartInfo` (signer, deploy address, gas limit override / effective / tx, and `OuterCallInfo`: the intercepted call's depth, caller, gas limit, remaining gas, value, and calldata) and `SandboxEndOutcome` (`Applied { Deployed | EmptyCode | ExecutionFailed, gas_used }` or `NotApplied { Rejected | PostAccountingHalt | ApplyFailed | AddressMismatch }`). A hook that returns `None` from `call`/`create` and leaves interpreter and context state alone observes without intervening, the same read-only notion the outer EVM's inspector has. `EmptyCode` is reported on every spec; the wire shape stays the frozen one (pre-REX5 drops the constructor's logs). Lifecycle events are delivered on the slot that also receives the sandbox's opcode-level hooks (`EmptyExternalEnv` pre-REX4, the parent env from REX4 on).
- Blanket impl: any type that implements `Inspector` for every sandbox context lifetime is a `SandboxInspector`. Local types that are not inspectors implement the trait by hand; coherence is pinned by unit tests.
- One type-erased slot on `MegaContext` (`Rc<RefCell<dyn SandboxInspector<E>>>`, held twice: parent env type and `EmptyExternalEnv`). Setting a hook replaces the previous one.
- Public API, forwarded identically from `MegaContext`, `MegaEvm`, and `MegaBlockExecutor`: `set_keyless_sandbox_hook(Rc<RefCell<I>>)` and `clear_keyless_sandbox_hook` (the only detach path). No take/drain API: recorded data stays in the caller's handle.
- External-env invariance: pre-REX4 sandboxes keep `EmptyExternalEnv`, REX4+ keep sharing the parent env, whether or not a hook is attached. `with_external_envs` resets the slot and debug-asserts it was empty.
- `AddressMismatch` is now checked before `apply_sandbox_state`. Unreachable on the no-hook path (the deploy address derives from the signer and the sandbox nonce is fixed); on the inspector channel a `create` short-circuit to another address must not leave applied state in the parent journal. A `create` short-circuit that returns no address is charged the same way (crate-private `SandboxRun::NoContractCreated`: REX5+ books the sandbox's gas and usage, nothing is applied, the outer call reverts with `NoContractCreated`); the no-hook path never produces it, and the public `SandboxOutcome` and `execute_keyless_deploy_call` keep their shapes.
- `inspectors` feature (optional `revm-inspectors` dependency): `sandbox::trace` ships `SharedTracingInspector` (outer EVM inspector), `SandboxTracer` (a hand-written `SandboxInspector` that never intervenes and records each sandbox execution into an arena of its own, keyed by the intercepted call from `OuterCallInfo`), `paired` to build both, and `splice_sandbox_traces`, which grafts every recorded sandbox under the outer `KeylessDeploy` CALL that started it and leaves the tracer empty. Splicing is idempotent; a sandbox whose call is not in the outer arena is dropped; a sandbox aborted by a database error inside a nested frame is closed as `FatalExternalError`; pre-REX5 empty-code deployments carry no logs (the receipt has none). The synthetic CREATE step on the intercepted CALL reports the outer frame's remaining gas and the sandbox reservation as `gasCost`. `mega-evme` `trace`/`replay` use it, and mega-reth's tracing RPCs use the same implementation.
- `test_utils::keyless`: shared corner-case init codes (CREATE then REVERT, three nesting levels with reverting calls and a log, sandbox out of gas, a call back into `KeylessDeploy`) used by the in-process tests, the offline state-test corpus, and the e2e suite.
- `benches/mega_bench.rs`: `keyless_sandbox_hook` rows (`rex5/no_hook`, `rex5/hook`) put the `InspectorBridge` → hook forwarding path under CodSpeed.

## Contract (main)

1. No hook attached: the sandbox path is unchanged.
2. Hook attached, not intervening: result, state, gas, and usage identical to the unattached path.
3. Interventions take effect inside the sandbox per revm's top-level semantics; the parent frame records the post-intervention gas and usage as-is and does not check conservation.
4. The hook is node-local and non-consensus.
5. Later specs may measure interventions and refuse some shapes; integrators must not depend on this base being permissive.

## Tests

- `tests/rex2/keyless_sandbox_hook.rs`: parity matrix REX2–REX6 × success/revert, REX5/REX6 resource halt, pre-REX4 crowded parent env, split-CREATE through the real interceptor; event ordering, reverted frames, short-circuited precompile frames balanced, lifecycle exactly once; all `SandboxEndOutcome` variants incl. `ApplyFailed` via an error-injecting database; a defaults-only hook attached to a logging, calling, creating, and self-destructing deployment leaves result, state, and usage identical to the no-hook run; `EmptyCode` is pinned on every spec together with the `EmptyCodeDeployed` wire shape; a hook with one impl per env sees a sandbox's lifecycle and opcode hooks on the same impl on every spec.
- `tests/rex2/keyless_sandbox_inspector.rs`: no-intervention parity (local no-op hook and `TracingInspector`) across specs and outcomes; interventions each with an unintervened control arm: `call` short-circuit, `create` short-circuit to `AddressMismatch` and to `Rejected`, `call_end` rewrite, `step` gas surcharge (sandbox `gas_used` exactly +1000; outer delta 0 pre-REX5, +1000 REX5+), journal writes, caller prank (`inputs.caller` rewritten on `call`, observed by the callee's `SSTORE(CALLER)`), creator prank (nested CREATE lands at `pranked.create(0)`, parent nonce untouched); setting a hook replaces the previous one and `clear` restores parity; seven outcomes reached through the hook; a defaults-only hook is inert on a self-destructing constructor; a `create` short-circuit without an address books the same outer gas and usage as the address-mismatch arm.
- `tests/rex2/keyless_sandbox_extreme.rs`: the four corner-case shapes, each run with a recording hook and with no hook, with byte-identical results.
- `tests/rex2/keyless_sandbox_trace.rs` (feature `inspectors`): the paired tracer grafts the three-level tree with consistent arena links, step depths one level below their frame, and a synthetic CREATE step carrying the outer remaining gas and the reservation, with result parity against the no-hook run; splicing is idempotent and `fuse` resets the outer; one tracer pair across two transactions grafts only the transaction just traced; a recorded sandbox is dropped when spliced into an empty outer, an unrelated transaction, or after the outer was fused; a database error inside a nested sandbox frame grafts the CREATE and the helper CALL as `FatalExternalError` (both tracer configs); empty-code logs follow the receipt (stripped REX2–REX4, kept REX5+); a parity-config leaf CREATE is still grafted with the synthetic step; no extra step when the parent already pairs one call-like step per child; a sandbox `SELFDESTRUCT` reaches the tracer; the `MegaBlockExecutor` forwarders attach a plain `TracingInspector`, detach, and attach a `SandboxTracer` across a three-transaction block; a validate-rejected sandbox claims its frame and grafts nothing; the outer adapter forwards create/log/selfdestruct and the blanket impl forwards log/selfdestruct to a plain `TracingInspector`.
- `bin/mega-evme`: geth call / opcode / prestate rendering with sandbox frames, parity / `flatCallTracer` trace addresses (`[] [0] [0,0] [0,0,0] [0,0,1] [0,1]`), splicing into an outer inspector that recorded no frames drops the sandbox (the arena's default root is never used as a parent), three offline replay tests over captured mainnet REX2 transactions (the synthetic CREATE step pins `depth`, the remaining gas, and the reservation).
- Corpus: `bench/replay/fixtures/keyless_deploy_rex2_{small,medium,large}.json` (mainnet blocks 7487457 / 7560053 / 8064149, manifest case `keyless_deploy`) and eight `keyless_sandbox_*_rex{5,6}.json` corner-case fixtures filled with `state-test --fill`.
- Verification: `cargo test -p mega-evm --all-features` (1275 passed), `cargo test -p mega-evme` (83 passed), `cargo test -p mega-state-test --test replay_corpus`, `cargo fmt --all --check`, `cargo clippy --workspace --lib --examples --tests --benches --all-features --locked`, `cargo sort --check …`, `cargo check -p mega-evm --target riscv64imac-unknown-none-elf --no-default-features`, `cargo doc -p mega-evm --no-deps` (no new broken links).
- Real chain: three mainnet REX2 keyless deployments replayed on a mainnet RPC node with identical results and sandbox frames rendered; devnet end-to-end (mega-reth + mega-e2e branches) renders the sandbox in every tracing entry point across 13 sub-scenarios.

## Docs

`crates/mega-evm/src/sandbox/mod.rs` and `sandbox::inspector` module docs and `crates/mega-evm/src/evm/AGENTS.md` ("Keyless sandbox hook") describe the hook, the contract, the env invariant, and the lifecycle slot rule. `docs/mega-evme/tracing/` documents the synthetic CREATE entry and the log rule for pre-REX5 empty-code deployments. `docs/spec/` is untouched: nothing here changes protocol behavior.

## Review follow-ups

The first review round (15 findings at `ce43e5b`) is addressed as follows:

- Fixed: stale / duplicated sandbox trees across transactions and non-idempotent splicing (per-execution arenas keyed by the intercepted call, consuming splice); `as_sandbox_observer` aliasing the outer handle (removed; `paired` builds the two halves); unclosed sandbox root rendered as a success (closed as `FatalExternalError`); pre-REX5 empty-code logs shown but absent from the receipt (stripped); lifecycle hooks landing on a different env slot than the opcode hooks (same slot); `SandboxCompletionKind` docs wrong pre-REX5 (uniform `EmptyCode`, pinned on every spec); `create` override with `address: None` refunding the whole reservation (charged like `AddressMismatch`); undocumented synthetic CREATE gas placeholder (real remaining gas and reservation, documented, `depth` pinned); deep clone of the sandbox arena (moved); no benchmark for the hooked arm (`keyless_sandbox_hook`); `Option` setters (by value, `clear_keyless_sandbox_hook` detaches).
- Deferred to follow-ups: the `AddressMismatch` hoist relying on ordering rather than a journal checkpoint (design note); `notify_sandbox_end` pairing enforced at seven sites (labeled-block refactor); test helper duplication across the keyless test files and `mega-evme`.

The second round (maintainer review at `f511680`) asked to drop the read-only `SandboxObserver` channel and keep `SandboxInspector` only: the read-only guarantee was partial (`step` and the context were mutable on both traits), the outer EVM's inspector has no read-only variant and can rewrite every frame of canonical execution, and enforcing read-only-ness cost a clone of every `CALL`/`CREATE` input and outcome on the tracing path. Done: one trait, one setter (`set_keyless_sandbox_hook`), `ReadOnlyHook` and the blanket observer impl deleted, the lifecycle types live in `sandbox::inspector`, and `SandboxTracer` forwards without copies.

## Notes for reviewers

- `inspectors` is off by default and does not affect `no_std` builds.
- Downstream: mega-reth (`cz/feat/keyless-sandbox-trace`) and mega-e2e (`cz/feat/keyless-sandbox-trace`) consume this branch; mega-reth pins a git rev and switches to the released tag once this lands.


crates/mega-evm/src/sandbox/trace.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    let parent_idx = outer
+        .traces()
+        .nodes()
+        .iter()
+        .rposition(|node| {
+            node.trace.address == KEYLESS_DEPLOY_ADDRESS && !node.trace.kind.is_any_create()
+        })
+        .unwrap_or(0);
```
flyq: **[MAJOR] correctness** · `CONFIRMED`

**`splice_sandbox_traces` grafts stale or duplicated sandbox trees for any consumer that keeps handles across transactions.**

splice_sandbox_traces falls back to unwrap_or(0) when no KeylessDeploy CALL exists in the outer arena, nothing resets the sandbox SharedTracingInspector between transactions (the blanket SandboxObserver impl leaves sandbox_start a no-op and only the outer handle is fused at replay/cmd.rs:528), and the function is not idempotent, so multi-tx consumers get stale sandbox trees grafted under unrelated roots or duplicated under the wrong parent.

**Suggested fix:** Early-return on `None` instead of `unwrap_or(0)`; make the splice consume the sandbox arena (`mem::take` + `fuse`, see finding 10) so a second splice is a no-op; and give the sandbox arena a per-transaction reset the caller cannot forget — either document next to the outer fuse that both handles must be fused, or replace the blanket impl with a concrete adapter that can observe `sandbox_start`. A two-transaction test plus a double-splice test would also kill the `trace.rs:48` mutant survivor.

<details>
<summary>Failure scenario</summary>

Consumer keeps one outer + one sandbox handle across a block and fuses only the outer (the pattern evm/AGENTS.md:31 advertises to node tracing RPCs): tx1 = keyless deploy using nested_keyless_call_init; tx2 = plain transfer to 0xaaaaaa -> rposition finds no KeylessDeploy node -> unwrap_or(0) -> callTracer renders CALL 0xaaaaaa { calls: [CREATE { calls: [CALL KeylessDeploy reverted] }] } although tx2 executed no CREATE (reproduced). If tx2 is another keyless deploy, push_trace's depth-0 branch overwrites only arena[0].trace and keeps arena[0].children/ordering, so tx1's reverted nested CALL appears under tx2's CREATE and root.ordering holds dangling Step(10..14) against steps.len()==10 (reproduced). Calling splice twice on the nested-keyless shape picks the reverted nested CALL as parent and grafts a duplicate subtree plus a synthetic CREATE step (reproduced: 3 -> 5 nodes). Codex's P2 on this line covers only the None fallback slice. The mutation gate agrees: replacing SharedTracingInspector::fuse with () at trace.rs:48 survives.

</details>

crates/mega-evm/src/sandbox/trace.rs
```diff
@@ -0,0 +1,239 @@
...
+        self.0.borrow()
+    }
+
+    /// Mutably borrows the inner inspector.
+    pub fn borrow_mut(&self) -> RefMut<'_, TracingInspector> {
+        self.0.borrow_mut()
+    }
+
+    /// Observer handle for the read-only keyless sandbox channel.
+    pub fn as_sandbox_observer(&self) -> Rc<RefCell<Self>> {
```
flyq: **[MAJOR] correctness** · `CONFIRMED`

**`as_sandbox_observer(&self)` on the outer handle silently overwrites the `KeylessDeploy` CALL root, then the splice panics.**

SharedTracingInspector::as_sandbox_observer(&self) returns an observer aliasing the same Rc<RefCell<TracingInspector>> arena, so calling it on the handle that is also the outer EVM inspector (the natural reading of the struct doc at 34-36) silently overwrites the KeylessDeploy CALL root with the sandbox CREATE and the follow-up splice panics; nothing guards or documents against it.

**Suggested fix:** Make the pairing a constructor (e.g. `SharedTracingInspector::paired() -> (outer, sandbox)`), or at minimum `debug_assert!(!Rc::ptr_eq(..))` in `splice_sandbox_traces` / `generate_trace_with_sandbox`, and say in the doc that the observer must not be the outer handle.

<details>
<summary>Failure scenario</summary>

MegaEvm::new(ctx).with_inspector(outer.clone()); evm.set_keyless_sandbox_observer(Some(outer.as_sandbox_observer())); run a REX5 keyless deploy (reproduced): no BorrowMutError during execution because the interceptor runs from inspect_frame_init after the outer call hook returned; the sandbox's depth-0 CREATE hits push_trace's arena[0].trace = new_trace branch, so the arena ends with ONE node: kind=Create, address=<deployed contract>, caller=<keyless signer>, gas_limit=<sandbox limit>, steps=<constructor steps>, but gas_used/status stamped by the outer call_end -- the KeylessDeploy CALL vanishes with no signal. splice_sandbox_traces(&mut outer.borrow_mut(), &outer.borrow()) then panics 'already mutably borrowed' at trace.rs:53. Both in-repo callers happen to build two separate inspectors.

</details>

crates/mega-evm/src/sandbox/trace.rs
```diff
@@ -0,0 +1,239 @@
...
+    let depth_offset = outer.traces().nodes()[parent_idx].trace.depth + 1;
+    let base = outer.traces().nodes().len();
+
+    let remapped: Vec<_> = sandbox
+        .traces()
+        .nodes()
+        .iter()
+        .cloned()
+        .enumerate()
+        .map(|(old_idx, mut node)| {
```
flyq: **[MEDIUM] correctness** · `CONFIRMED`

**A sandbox root CREATE that was never closed is grafted as-is and renders as a success.**

splice_sandbox_traces grafts a sandbox arena whose root CREATE was never closed (status None, gas_used 0) when a DB error aborts the sandbox inside a nested frame, so the geth call tracer renders the CREATE with no error and the parity tracer renders it as a successful Create under an outer CALL that reverted with InternalError; sandbox_has_frames (line 119, doc 'at least one completed frame') is inaccurate and used by only one unit test.

**Suggested fix:** Close dangling nodes in the remap closure (`status = Some(FatalExternalError)`, `gas_used = gas_limit`). `sandbox_has_frames`'s doc ("at least one completed frame") is inaccurate and it is used by one unit test; drop it or fix the doc.

<details>
<summary>Failure scenario</summary>

Traced keyless deploy whose constructor CALLs a helper that executes BALANCE on an address for which the backing DB errors (reproduced with ErrorInjectingDatabase{fail_on_account=sentinel}): the inner CALL closes with FatalExternalError, then EthFrame::return_result on the parent runs take_error(ctx.error())? (revm-handler frame.rs:454) before the parent's frame_end, so the sandbox root keeps status None; transact_raw returns Err -> Rejected(InternalError) -> outer CALL reverts with 0xfe835e35; has_frames(sandbox) is true (steps=8, children=[1]) so the arena is grafted unchanged. Observed: callTracer CREATE child has no error field and gasUsed==gas; parity [0] = error: None, result: Some(Create(CreateOutput{address: 0x53c3.., code: 0x, gas_used: 0})) while [0,0] shows FatalExternalError and [] shows Reverted. When the failing opcode is in the top-level constructor create_end does fire and it renders correctly.

</details>

crates/mega-evm/src/sandbox/trace.rs
```diff
@@ -0,0 +1,239 @@
...
+///
+/// Geth-style struct logs only descend into a child when the parent frame has a matching
+/// call-like opcode in its steps. An intercepted `KeylessDeploy` CALL records no bytecode, so
+/// a CREATE step is grafted onto the parent; without it the sandbox constructor steps would
+/// be omitted from struct-log output instead of nested in execution order.
+///
+/// An `outer` that recorded no frames at all (already fused, or never run) is also left
+/// untouched: its arena holds only the default root node, and grafting the sandbox under
+/// that phantom root would invent a call tree the outer EVM never executed.
+pub fn splice_sandbox_traces(outer: &mut TracingInspector, sandbox: &TracingInspector) {
```
flyq: **[MEDIUM] correctness** · `CONFIRMED`

**REX2–REX4 empty-code deploys show logs in the spliced call trace that the receipt dropped.**

For REX2-REX4 keyless deploys whose constructor emits a LOG and returns empty runtime code, the spliced call trace (with_log) shows the log under the grafted sandbox CREATE while the receipt contains zero logs, violating the geth withLog invariant (logs shown iff in receipt) that revm-inspectors encodes via call_or_parent_failed; PR-introduced because before the splice no tracer could see the sandbox.

**Suggested fix:** Fix in the tracing layer, not consensus: `sandbox_end` delivers `Applied { ExecutionFailed }` for exactly this case while the CREATE node is not `is_error()`, so `SharedTracingInspector`/splice can strip or mark those logs. Add a REX2–REX4 logging-empty-code trace test.

<details>
<summary>Failure scenario</summary>

mega-evme run/replay at REX2-REX4 with --trace --tracer call --trace.call.with-log on a KeylessDeploy call whose init code is LOG1(..); RETURN(0,0) (reproduced at REX3 and REX4: child CREATE logs=1, receipt logs=0; REX5 control: 1/1). process_sandbox_transact_result (execution.rs:929-941) takes the !is_rex5_enabled arm and collapses to ExecutionFailed{EmptyCodeDeployed}, so the dispatch arm at 540 never calls ctx.log() (frozen, intentional, pinned by tests/rex5/keyless_empty_code_logs.rs); but the sandbox TracingInspector already recorded the log on a CREATE node with status Return, and the outer CALL is built by make_execution_failure! with InstructionResult::Return, so call_or_parent_failed() is false for both nodes and geth_call_traces renders frame.calls[0].logs = [log]. Only terminal arm where a success-status child carries logs the receipt dropped.

</details>

crates/mega-evm/src/sandbox/execution.rs
```diff
@@ -433,14 +432,29 @@ pub fn execute_keyless_deploy_call<DB: AlloyDatabase, ExtEnvs: ExternalEnvTypes>
...
-            "Rex5+ sandbox pre-debit must succeed: gas_limit_override is capped to gas.remaining()",
+            "Rex5+ sandbox pre-debit must succeed: effective_gas_limit is capped to gas.remaining()",
         );
     }
 
     // Step 9: Execute sandbox and apply state changes.
+    //
+    // INVARIANT: when a hook is attached, `sandbox_start` and `sandbox_end`
+    // fire exactly once on this path, covering every terminal arm below.
+    let hook = ctx.keyless_sandbox_hook.clone();
```
flyq: **[MEDIUM] api-contract** · `CONFIRMED`

**Lifecycle hooks and opcode hooks can land on different `SandboxObserver<E>` impls pre-REX4, and no doc says so.**

sandbox_start/sandbox_end are always dispatched through the parent-env slot (ctx.keyless_sandbox_hook) regardless of spec, while pre-REX4 opcode-level hooks are dispatched through the EmptyExternalEnv slot, so a hook type with hand-written per-env impls receives the lifecycle pair on SandboxObserver<ParentEnv> and every step/call/create on SandboxObserver<EmptyExternalEnv> at REX2/REX3; no doc states this dispatch rule.

**Suggested fix:** Fire the lifecycle on the same spec-selected slot as the opcode hooks (lines 691/694), or document the rule on both setters.

<details>
<summary>Failure scenario</summary>

Parent ExtEnvs != EmptyExternalEnv, spec REX2/REX3 (the historical-replay window). An integrator writes a full impl SandboxObserver<EmptyExternalEnv> for T and an empty stub impl SandboxObserver<MyEnv> for T {} to satisfy the setter's dual bound. Replaying a REX2/REX3 keyless deploy delivers every opcode event but silently drops sandbox_start/sandbox_end (they land on the stub). Reproduced with a divergent dual-impl observer: REX2/REX3 lifecycle=[(parent,start),(parent,end)] opcode_impls={empty}; REX4+ all on parent. Same split for set_keyless_sandbox_inspector (context.rs:558-563). Every doc site (observer.rs:52-56, context.rs:83-84/94-97, sandbox/mod.rs:87-92) scopes the dual-slot split to 'opcode-level hooks' and never says where the lifecycle lands.

</details>

crates/mega-evm/src/sandbox/observer.rs
```diff
@@ -0,0 +1,537 @@
...
+
+/// Completion kind for a sandbox whose state was applied to the parent.
+#[non_exhaustive]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum SandboxCompletionKind {
+    /// Inner CREATE succeeded with non-empty runtime bytecode.
+    Deployed,
+    /// Inner CREATE succeeded but returned empty runtime bytecode.
+    EmptyCode,
+    /// Sandbox EVM execution failed (revert or halt) after producing mergeable state.
```
flyq: **[MEDIUM] documentation** · `CONFIRMED`

**`SandboxCompletionKind` docs are wrong for pre-REX5 empty-code deploys, and no test pins the pre-REX5 kind.**

SandboxCompletionKind's public docs are wrong for pre-REX5 empty-code deploys: they are delivered as Applied { ExecutionFailed } (documented as 'revert or halt', though the constructor returned successfully) rather than EmptyCode (documented with no REX5+ qualifier, unlike its internal twin at execution.rs:824-828); the Applied.completion doc 'wire-shape completion reported to the outer caller' is also inaccurate since both cases share the same ABI wire shape, and no test pins the pre-REX5 kind.

**Suggested fix:** The observer is off-consensus, so either report `EmptyCode` uniformly, or correct both doc lines and add a REX2–REX4 arm asserting the actual kind.

<details>
<summary>Failure scenario</summary>

An observer keyed on SandboxCompletionKind::EmptyCode silently misses every such deploy on REX2-REX4 blocks, and one treating ExecutionFailed as 'the sandbox EVM reverted or halted' misclassifies a Return-status constructor. Reproduced by temporarily adding MegaSpecId::REX4 to the [REX5, REX6] loop of test_sandbox_end_applied_empty_code_on_rex5 (tests/rex2/keyless_sandbox_observer.rs:539): 'REX4: expected Applied(EmptyCode), got End(Applied { completion: ExecutionFailed, gas_used: 92534 })'. Chain: execution.rs:923-941 bytecode.is_empty() && !is_rex5_enabled -> SandboxCompletion::ExecutionFailed{EmptyCodeDeployed} -> dispatch arm 540-551 -> notify_sandbox_end(Applied{ExecutionFailed}).

</details>

crates/mega-evm/src/sandbox/execution.rs
```diff
@@ -502,8 +560,12 @@ pub fn execute_keyless_deploy_call<DB: AlloyDatabase, ExtEnvs: ExternalEnvTypes>
             // upfront is intentionally retained, mirroring the upfront
             // `KEYLESS_DEPLOY_OVERHEAD_GAS` charge.
             if ctx.spec.is_enabled(MegaSpecId::REX5) {
-                gas.erase_cost(gas_limit_override_u64);
+                gas.erase_cost(effective_gas_limit);
```
flyq, line 606: **[MEDIUM] api-contract** · `CONFIRMED`

**An inspector `create` override with `address: None` fully refunds the sandbox reservation and drops its usage, contradicting contract item 3.**

An inspector create override returning Some(CreateOutcome { result: Return, address: None }) on the top-level sandbox CREATE lands in the pre-existing Success{Output::Create(_, None)} -> Rejected(NoContractCreated) arm (903-912), so on REX5+ the Rejected arm fully refunds the reservation via gas.erase_cost(effective_gas_limit) and discards the sandbox's limit_usage/volatile_accesses although the sandbox tx ran and consumed intrinsic gas, contradicting inspector contract item 3 and the Rejected docs, and asymmetric with the AddressMismatch path where Some(OTHER_ADDR) is charged and merged.

**Suggested fix:** Either carve out synthetic `address: None` explicitly in contract item 3, the `Rejected` docs, and the arm comment, or route `Output::Create(_, None)` through `apply_sandbox_post_accounting` like `AddressMismatch`. Make `test_inspector_create_short_circuit_without_address_is_rejected` actually compare against `_control`.

<details>
<summary>Failure scenario</summary>

REX5+ keyless deploy with a rewriting SandboxInspector whose create returns Some(CreateOutcome{result: Return, address: None}): frame_start returns it as the top-level FrameResult::Create (Output::Create(_, None)); last_frame_result reports gas_used = sandbox intrinsic gas and evm/execution.rs:702 already recorded it into the sandbox additional_limit; process_sandbox_transact_result returns Rejected(NoContractCreated) dropping gas_used/limit_usage/volatile_accesses; the Rejected arm (556-570) erases the whole reservation -> the outer call reverts with NoContractCreated but charges zero sandbox gas and omits the sandbox footprint from tx-level counters. test_inspector_create_short_circuit_without_address_is_rejected (tests/rex2/keyless_sandbox_inspector.rs:807-849) computes _control and never compares outer gas_used or usage, so the refund is not pinned. Node-local only.

</details>

crates/mega-evm/src/sandbox/trace.rs
```diff
@@ -0,0 +1,239 @@
...
+    parent.children.push(base);
+
+    // Struct-log rendering pairs call-like parent steps with `children` in order. A missing
+    // CREATE step here would leave the grafted child unreachable.
+    let calllike = parent.trace.steps.iter().filter(|s| is_calllike_opcode(s.op.get())).count();
+    if calllike < parent.children.len() {
+        let step_idx = parent.trace.steps.len();
+        let create_depth = parent.trace.depth as u64 + 1;
+        let contract = parent.trace.address;
+        let gas_remaining = parent.trace.gas_limit;
```
flyq: **[MINOR] correctness** · `CONFIRMED`

**The synthetic CREATE step reports the outer frame's inbound gas and `gasCost: 0`, undocumented.**

The synthetic CREATE step stamps gas = parent.trace.gas_limit (the intercepted CALL's full inbound gas, before the 100K overhead / REX5 materialization / REX5 pre-debit were charged) and hardcodes gasCost = 0, so struct logs show a CREATE whose gas overstates the outer frame's real remaining gas and whose gasCost hides the frame's largest charge; pre-REX5 the next step's gas can exceed it; the placeholder is pinned by replay_keyless.rs:96-102 and not mentioned in docs/mega-evme/tracing/opcode-tracer.md, which also does not say a CREATE entry the contract never executed is inserted.

**Suggested fix:** Expose the real `gas.remaining()` via an `outer_gas_remaining` field on `SandboxStartInfo` that the tracing observer stashes, or at minimum document the placeholder and the synthetic entry in `opcode-tracer.md`. Pin `depth` in the `replay_keyless.rs` fixture assertion to kill the `trace.rs:196` survivors.

<details>
<summary>Failure scenario</summary>

Replay the pinned REX2 fixture with --trace --tracer opcode (reproduced): structLogs[0] = {op: CREATE, pc: 0, gas: 636184, gasCost: 0, depth: 1}, structLogs[1] = {op: PUSH1, gas: 25510737, gasCost: 3, depth: 2}. Real remaining gas at sandbox_start was 636184 - 100000 = 536184 (overhead debited at execution.rs:176-177); a consumer reading opcode-tracer.md:49-50 literally or differencing consecutive gas values sees +24.9M gas materialize inside a child frame. Cosmetic: geth builder copies step.gas_cost verbatim. The two surviving + -> - / + -> * mutants at trace.rs:196 (create_depth = depth + 1) show the synthetic step's depth is not pinned either.

</details>

flyq (commented):
## Review — observer/inspector channels and sandbox trace splicing

Reviewed at `ce43e5b22c20` · xhigh effort · 15 findings

Reviewed the full diff against the branch checkout at `ce43e5b`. Locally: `cargo fmt --all --check`, `cargo clippy --workspace --lib --examples --tests --benches --all-features --locked`, `cargo sort --check`, Prettier on `docs/`, and the riscv `no_std` check all clean. Every correctness finding below was reproduced with a throwaway test, since reverted. Line numbers are from the post-change files at that commit.

**Verdict: the core is sound; the tracing glue is not merge-ready yet.** With no hook attached the sandbox path is semantically unchanged, the `AddressMismatch` hoist is unreachable on that path, and every wrapper forwards all nine revm `Inspector` hooks. The defects are in `sandbox/trace.rs`, which is the piece the PR advertises to mega-reth's tracing RPCs, and on exactly that shared-handle, multi-transaction usage it produces wrong output: stale sandbox trees under unrelated transactions (1), a silent root overwrite plus a `BorrowMutError` panic when the observer aliases the outer handle (2), a never-closed CREATE rendered as a success under a reverted CALL (3), and receipt-inconsistent logs on REX2–REX4 (4). None of these are reachable from `mega-evme`'s own CLI, which builds fresh handles and splices once, so the PR's own tests are green. Findings 5–7 and 12 are public-contract mismatches an integrator will trip on; the rest is cleanup. Specifics are inline on the relevant lines.

### What I verified

1. **The `AddressMismatch` hoist does not change frozen-spec behavior.** The sandbox pins the signer nonce to 0 via `with_nonce_override`, so on the no-hook path the CREATE address is a pure function of the signer and the pre-`apply_sandbox_state` check is structurally unreachable; only an inspector `create` short-circuit to another address can reach it. The no-hook path was restructured into the `None` arm of `run_sandbox_ctx` with the same body, matching contract item 1 and the `spec:stable` / `spec:unchanged` labels.
2. **Every wrapper forwards all nine revm `Inspector` hooks.** The blanket `SandboxObserver`/`SandboxInspector` impls, `ReadOnlyHook`, `InspectorBridge`, and `SharedTracingInspector` each cover `initialize_interp`, `step`, `step_end`, `log`, `call`, `call_end`, `create`, `create_end`, `selfdestruct`.
3. **Two candidates were refuted and are not reported.** The manifest note's 25.5M gas figure is correct once the 10,000-gas/byte code-deposit charge is included; the state-test runner's inability to trace sandbox frames predates this PR.
4. **Lint and portability gates are clean**, `--locked` metadata resolves, all new `#[test]` functions carry the `test_` prefix, and `docs/` passes Prettier.

### Ties to existing signal

The `cargo-mutants gate` failure has two survivors in `sandbox/trace.rs` that map directly onto findings here: `fuse` replaced with `()` at `trace.rs:48` (nothing observes the reset, finding 1) and the `+ 1` on `create_depth` at `trace.rs:196` (the synthetic step's depth is unpinned, finding 8). The tests for those two findings should also turn the gate green. Codex's P2 on `trace.rs:159` covers only the `rposition → None` fallback slice of finding 1; the missing per-transaction reset, the non-idempotence, and the aliasing footgun in finding 2 are new.

### Inline findings

| # | Where | Finding |
|---|---|---|
| 1 | `crates/mega-evm/src/sandbox/trace.rs:159` | splice grafts stale/duplicate sandbox trees across txs |
| 2 | `crates/mega-evm/src/sandbox/trace.rs:62` | as_sandbox_observer on outer handle clobbers root, panics |
| 3 | `crates/mega-evm/src/sandbox/trace.rs:169` | Unclosed sandbox root CREATE grafted and rendered as success |
| 4 | `crates/mega-evm/src/sandbox/trace.rs:147` | REX2-REX4 empty-code deploy trace shows logs receipt dropped |
| 5 | `crates/mega-evm/src/sandbox/execution.rs:446` | Lifecycle and opcode hooks dispatched on different env slots |
| 6 | `crates/mega-evm/src/sandbox/observer.rs:125` | SandboxCompletionKind docs wrong pre-REX5 for empty code |
| 7 | `crates/mega-evm/src/sandbox/execution.rs:563` | create override with address None refunds whole reservation |
| 8 | `crates/mega-evm/src/sandbox/trace.rs:198` | Synthetic CREATE step gas placeholder is undocumented |
| 9 | `crates/mega-evm/src/sandbox/execution.rs:494` | AddressMismatch hoist relies on ordering, not a checkpoint |
| 10 | `crates/mega-evm/src/sandbox/trace.rs:167` | Splice deep-clones the whole sandbox arena including stacks |
| 11 | `crates/mega-evm/src/sandbox/execution.rs:719` | No Criterion/CodSpeed bench for the hooked sandbox arm |
| 12 | `crates/mega-evm/src/sandbox/observer.rs:324` | Observer read-only guarantee structural only for call/create |
| 13 | `crates/mega-evm/src/sandbox/execution.rs:444` | notify_sandbox_end pairing enforced by convention at 7 sites |
| 14 | `crates/mega-evm/src/evm/context.rs:515` | Option setter keeps a dead None arm on 3 public types |
| 15 | `crates/mega-evm/tests/rex2/keyless_sandbox_support.rs:185` | Test helpers re-derived instead of using test_utils::keyless |

### Suggested path to merge

1. Fix findings 1 and 2 together in `sandbox/trace.rs`: early-return on `None`, consume-and-fuse the sandbox arena in the splice (finding 10's `mem::take` shape gives idempotence for free), a `paired()` constructor or `ptr_eq` guard, and a two-transaction test.
2. Close dangling sandbox roots in the remap closure (3) and strip receipt-inconsistent logs for the REX2–REX4 empty-code arm (4) — both tracing-layer only, no frozen-spec risk.
3. Pin the synthetic CREATE step's `depth` in `replay_keyless.rs` and either surface real remaining gas or document the placeholder (8). Steps 1 and 3 together should clear the mutation gate.
4. Settle the three public-contract items before mega-reth pins a release: lifecycle-slot dispatch rule (5), `SandboxCompletionKind` docs or uniform `EmptyCode` (6), and the `address: None` refund carve-out (7). Finding 14's signature change is cheapest in the same pass.
5. Findings 9, 11, 12, 13, 15 can follow up separately; 12 deserves at least the negative test now so the read-only promise is not purely documentary.

**Not reported inline (lower value, all confirmed):** unused `PartialEq` derive on `LimitUsage` (`crates/mega-evm/tests/rex2/keyless_sandbox_support.rs:339`); the `Some`/`None` arms of `run_sandbox_ctx` duplicate the sandbox construction (`crates/mega-evm/src/sandbox/execution.rs:718`); in `bin/mega-evme/src/replay/cmd.rs:475` the outer `SharedTracingInspector` is built unconditionally while the sandbox handle is gated on `is_tracing_enabled()`.

crates/mega-evm/src/sandbox/execution.rs
```diff
@@ -451,15 +465,41 @@ pub fn execute_keyless_deploy_call<DB: AlloyDatabase, ExtEnvs: ExternalEnvTypes>
...
+            // state on this path.
+            if let SandboxCompletion::Deployed { deploy_address: deployed, .. } = &completion {
+                if *deployed != deploy_address {
+                    notify_sandbox_end(
+                        &hook,
+                        SandboxEndOutcome::NotApplied {
+                            reason: SandboxRejectKind::AddressMismatch,
+                        },
+                    );
+                    return make_error!(KeylessDeployError::AddressMismatch);
```
flyq, line 518: **[DESIGN NOTE] robustness** · `CONFIRMED`

**The `AddressMismatch` hoist protects the parent journal by ordering, not by mechanism — latent, not live at HEAD.**

The AddressMismatch hoist protects the parent journal by ordering only: the interceptor's synthetic FrameResult is delivered with no journal checkpoint around it, so any Revert-shaped make_error! issued after apply_sandbox_state (498) would leave applied sandbox state in ResultAndState.state next to ExecutionResult::Revert and the block executor would commit it; nothing structural prevents a future post-apply reject from reintroducing it.

**Suggested fix:** Not a merge blocker. Suggestion: wrap interceptor dispatch in a journal checkpoint reverted whenever the synthetic result is not ok, so every interceptor revert is journal-clean by construction.

<details>
<summary>Failure scenario</summary>

Latent, not live at HEAD. Hypothetical follow-up: a REX7 shape check after line 503 (contract item 5 invites one) that returns make_error!(X) on a REX5+ keyless deploy whose sandbox deployed code: apply_sandbox_state_journaled has already checkpoint_commit'ed the deployment, signer nonce bump and storage into the depth-0 journal (state_merge.rs:77-86 commits upward; pre-REX5 apply_sandbox_state_legacy writes journal.state directly); evm/execution.rs:1260-1268 returns FrameInitResult::Result before the inner frame_init that would create a checkpoint; both exec loops return the first-frame Result directly; block/executor.rs:638 commits state regardless of result -> the chain ends up with a deployed contract from a call that reverted.

</details>

crates/mega-evm/src/sandbox/trace.rs
```diff
@@ -0,0 +1,239 @@
...
+        })
+        .unwrap_or(0);
+    let depth_offset = outer.traces().nodes()[parent_idx].trace.depth + 1;
+    let base = outer.traces().nodes().len();
+
+    let remapped: Vec<_> = sandbox
+        .traces()
+        .nodes()
+        .iter()
+        .cloned()
```
flyq: **[MINOR] efficiency** · `CONFIRMED`

**The splice deep-clones the whole sandbox arena, including every step's stack snapshot.**

splice_sandbox_traces deep-clones every sandbox CallTraceNode, including every CallTraceStep with its stack/push_stack Vec<U256>, into an intermediate Vec (163-184) before extend (186), although the sandbox arena is never read again after the graft.

**Suggested fix:** Take `sandbox: &mut TracingInspector`, `core::mem::take(sandbox.traces_mut().nodes_mut())` then `sandbox.fuse()`, and `extend` from the mapped iterator. Minimal no-signature fix: drop the `collect()`.

<details>
<summary>Failure scenario</summary>

mega-evme traces use TracingInspectorConfig::all() (bin/mega-evme/src/common/trace.rs:103, full stack snapshots per step); tracing a keyless deploy with a compute-heavy constructor transiently holds two full copies of the sandbox arena plus per-step stack clones while the original stays alive in the observer handle.

</details>

crates/mega-evm/src/sandbox/execution.rs
```diff
@@ -646,18 +715,49 @@ fn run_sandbox_ctx<DB: AlloyDatabase, ExtEnvs: ExternalEnvTypes>(
...
-        sandbox_evm.ctx.volatile_data_tracker.borrow().get_volatile_data_accessed();
-    process_sandbox_transact_result(
-        result,
-        limit_usage,
-        volatile_accesses,
-        is_rex5_enabled,
-        is_rex6_enabled,
-    )
+    match hook {
+        Some(hook) => {
```
flyq, line 759: **[PROCESS] benchmark-coverage** · `CONFIRMED`

**No Criterion/CodSpeed benchmark for the hooked `run_sandbox_ctx` arm.**

The PR adds the Some(hook) arm of run_sandbox_ctx (per-opcode Rc<RefCell<dyn>> dispatch through InspectorBridge -> ReadOnlyHook -> observer) and reworks execute_keyless_deploy_call, but adds no Criterion/CodSpeed benchmark; the only bench addition is a bench/replay/manifest.json case that runs state-test --bench with no hook and is not wired into the per-PR CodSpeed workflow.

**Suggested fix:** One `mega_bench` case for a keyless deploy with a `SharedTracingInspector` attached, so per-PR CodSpeed sees regressions in the `InspectorBridge → ReadOnlyHook` forwarding path.

<details>
<summary>Failure scenario</summary>

AGENTS.md rule: 'Changes on the EVM execution hot path must be accompanied by benchmarks. This includes ... system contract interception'. git diff --stat -- crates/mega-evm/benches .github is empty; grep set_keyless_sandbox over benches/, bench/, .github/ finds nothing; codspeed.yml:63 builds only crates/mega-evm bench targets (mega_bench.rs:432 interception benches hit only 0x..0004/0005); replay-bench.yml runs weekly / manual only. A future per-opcode regression in InspectorBridge/ReadOnlyHook/SharedTracingInspector forwarding produces no signal on any PR. The hooked arm is opt-in tooling, so this is process-level; the manifest entry still improves on main.

</details>

crates/mega-evm/src/sandbox/observer.rs
```diff
@@ -0,0 +1,537 @@
...
+        context: &mut MegaContext<SandboxDb<'_>, E>,
+        log: Log,
+    ) {
+        Inspector::log(self, interp, context, log);
+    }
+
+    #[inline]
+    fn call(&mut self, context: &mut MegaContext<SandboxDb<'_>, E>, inputs: &CallInputs) {
+        // The inspector works on a copy: its mutations and any override are discarded.
+        let mut inputs = inputs.clone();
```
flyq, line 367: **[MEDIUM] api-contract** · `CONFIRMED`

**The observer channel's read-only guarantee is structural only for `call`/`create`; `step`-family hooks hand `&mut` through, and nothing pins that.**

The observer channel's structural read-only guarantee covers only CALL/CREATE inputs and overrides (clone-and-discard at 324/336/342/353 plus ReadOnlyHook::call/create returning None at 431/451); step/step_end/initialize_interp/log forward &mut Interpreter and &mut MegaContext verbatim, so a step-mutating type is accepted by set_keyless_sandbox_observer and rewrites sandbox state, no negative test pins that, and the tracing path pays a three-deep RefCell/dyn chain per hook that as_sandbox_observer could shorten by one layer.

**Suggested fix:** At minimum a negative test that a step-mutating observer is rejected or documented as unsupported; ideally one trait with a read-only mode on `InspectorBridge`. Related: `as_sandbox_observer` could return `Rc::clone(&self.0)` directly, saving one `RefCell`/`dyn` layer per hook.

<details>
<summary>Failure scenario</summary>

A caller attaches through the 'read-only' channel a type whose step calls interp.gas.record_cost(..) or context.journal_mut().sstore(..) (a one-token trait rename SandboxInspector -> SandboxObserver on the test-side StepGasSurcharge/JournalWriter yields a compiling mutating observer): ReadOnlyHook::step (observer.rs:402) hands the real &mut through, the mutation lands in the sandbox journal, is merged into the parent, and the node diverges from consensus; the only guard is the doc sentence at observer.rs:24-31, and test_observer_channel_drops_call_overrides (keyless_sandbox_observer.rs:702) only proves a call override is dropped. Per sandbox opcode the observer channel runs three borrow_mut/dyn hops, twice (step + step_end); as_sandbox_observer (trace.rs:62-64) could return Rc::clone(&self.0) since TracingInspector: Inspector<CTX> already satisfies the blanket bound.

</details>

crates/mega-evm/src/sandbox/execution.rs
```diff
@@ -433,14 +432,29 @@ pub fn execute_keyless_deploy_call<DB: AlloyDatabase, ExtEnvs: ExternalEnvTypes>
...
         debug_assert!(
             ok,
-            "Rex5+ sandbox pre-debit must succeed: gas_limit_override is capped to gas.remaining()",
+            "Rex5+ sandbox pre-debit must succeed: effective_gas_limit is capped to gas.remaining()",
         );
     }
 
     // Step 9: Execute sandbox and apply state changes.
+    //
+    // INVARIANT: when a hook is attached, `sandbox_start` and `sandbox_end`
```
flyq, line 461: **[MINOR] robustness** · `CONFIRMED`

**`notify_sandbox_end` pairing is enforced by convention at seven sites.**

notify_sandbox_end is hand-placed at seven terminal sites (472, 488, 499, 515, 528, 545, 565) with three early returns after sandbox_start (478, 494, 502), so the 'fires exactly once' INVARIANT comment at 444-445 is enforced only by convention; all four frame-result macros evaluate to FrameResult values with no return and the match at 458 is the tail expression, so a labeled block yielding (SandboxEndOutcome, FrameResult) once would make the pairing compiler-enforced.

**Suggested fix:** A labeled block yielding `(SandboxEndOutcome, FrameResult)` with one `notify_sandbox_end` call after it makes the pairing compiler-enforced.

<details>
<summary>Failure scenario</summary>

A future terminal arm (e.g. a REX7 post-sandbox rejection per contract item 5) added without its own notify_sandbox_end silently breaks the sandbox_start/sandbox_end pairing for every hook; it is caught only if a test drives that arm (the pairing is pinned solely by assert_single_start_end_pair, keyless_sandbox_observer.rs:316-322, on the paths tests happen to exercise). Ordering is unobservable to hooks: sandbox_end receives only &SandboxEndOutcome and ctx.log at 511/524 is a journal push that fires no Inspector::log hook.

</details>

crates/mega-evm/src/evm/context.rs
```diff
@@ -452,6 +494,88 @@ impl<DB: Database, ExtEnvs: ExternalEnvTypes> MegaContext<DB, ExtEnvs> {
...
+    ///
+    /// Attaching an observer does not change sandbox external-env semantics at
+    /// any spec. Replaces any attached inspector.
+    ///
+    /// Use [`Self::clear_keyless_sandbox_hook`] to detach. Passing `None`
+    /// also clears both slots, but type inference for `O` often fails on that
+    /// path. Observation is read-only: mutating interpreter or context state
+    /// through the hooks is undefined and may diverge consensus. There is no
+    /// take/drain API; recorded data stays in the caller's observer.
+    pub fn set_keyless_sandbox_observer<O>(&mut self, observer: Option<Rc<RefCell<O>>>)
```
flyq: **[MINOR] api-design** · `CONFIRMED`

**`Option<Rc<RefCell<O>>>` setter signature carries a dead `None` arm and a self-described inference footgun, on three public types.**

set_keyless_sandbox_observer<O>(Option<Rc<RefCell<O>>>) and set_keyless_sandbox_inspector<I>(Option<..>), public on MegaContext, MegaEvm (evm/mod.rs:246/259) and MegaBlockExecutor (block/executor.rs:161/177), carry an Option solely for a None => clear arm whose own docs say type inference fails on it; no call site passes a literal None, all 18 direct callers pass Some, clear_keyless_sandbox_hook already exists on all three types, and this signature becomes a semver break once mega-reth pins a released tag.

**Suggested fix:** Take `Rc<RefCell<O>>` by value now and keep `clear_keyless_sandbox_hook` as the single detach path, before mega-reth pins a released tag.

<details>
<summary>Failure scenario</summary>

A downstream caller writing evm.set_keyless_sandbox_observer(None) hits a type-inference error and must either turbofish a phantom observer type (the 12 test sites do exactly None::<Rc<RefCell<RecordingObserver>>>, e.g. tests/rex2/keyless_sandbox_inspector.rs:670) or discover the separate clear method; the dead None => clear arms (context.rs:532, :565) and the 'type inference for O often fails' caveat are repeated across six doc blocks on three public types.

</details>

crates/mega-evm/tests/rex2/keyless_sandbox_support.rs
```diff
@@ -0,0 +1,400 @@
...
+pub(crate) fn create_pre_eip155_deploy_tx_with_value(
+    init_code: Bytes,
+    value: U256,
+) -> (Bytes, Address) {
+    create_pre_eip155_deploy_tx_with_value_and_gas_limit(init_code, value, SIGNED_TX_GAS_LIMIT)
+}
+
+/// Like [`create_pre_eip155_deploy_tx_with_value`] with an explicit inner gas limit, for
+/// shapes that need the sandbox to run out of gas.
+pub(crate) fn create_pre_eip155_deploy_tx_with_value_and_gas_limit(
```
flyq, line 191: **[MINOR] reuse** · `CONFIRMED`

**Test-side duplication the PR's own `test_utils::keyless` was created to end.**

The PR's new test code re-derives helpers the PR itself created a shared home for: the r=s=0x2222 pre-EIP-155 signer / keylessDeployCall builder / funded_db / TEST_CALLER and LARGE_* constants are copied again into keyless_sandbox_support.rs and bin/mega-evme/src/common/trace.rs:319 (now 11 signer copies in 9 files) instead of test_utils/keyless.rs; the keyless run harness is spelled three times (support.rs:271-386, keyless_sandbox_inspector.rs:106-169, keyless_sandbox_extreme.rs:156) differing only in the setter; three recorders with overlapping event enums exist (RecordingObserver is defined twice with different bodies); and test_utils/keyless.rs keeps call_and_pop / return_stop_runtime / INTERNAL_CALL_GAS private so tests re-spell them with a bare 50_000_u32 that push_number emits as PUSH4 instead of the u16 helper's PUSH2.

**Suggested fix:** Move the signer/call-tx/`funded_db` helpers and constants into `test_utils::keyless`, make the bytecode fragments `pub`, and keep one harness taking `attach: impl FnOnce(&mut MegaContext)` with one recorder type.

<details>
<summary>Failure scenario</summary>

A change to the keyless signing scheme, the keylessDeployCall ABI, or default gas constants must be edited in 11 signer copies and ~10 call-tx builders; a run-config field change touches 3 harness bodies and 32 seven-field struct literals; retuning INTERNAL_CALL_GAS leaves five 50_000 literals silently diverging, and the copies are not byte-identical to the shared helper (PUSH4 vs PUSH2); hex!("60006000fd") is still spelled at support.rs:66 / observer.rs:610 / inspector.rs:657 despite the new REVERTING_RUNTIME.

</details>

crates/mega-evm/src/sandbox/observer.rs
```diff
@@ -0,0 +1,589 @@
...
+/// [`call`](Self::call) / [`create`](Self::create) cannot rewrite inputs or
+/// override the frame; [`step`](Self::step) and context arguments remain
+/// contractually read-only. `ReadOnlyHook` never answers a `call`/`create`
+/// override on the observer's behalf.
+///
+/// Types that already implement [`Inspector`] for every sandbox context
+/// lifetime receive a blanket impl that forwards each hook on a temporary
+/// copy of the inputs. Local types that are not inspectors implement this
+/// trait by hand.
+pub trait SandboxObserver<ExtEnvs: ExternalEnvTypes> {
```
Troublor, line 209: **Requesting a change before merge:** drop `SandboxObserver` entirely and keep only `SandboxInspector`.

The two traits declare the same hook set; the only difference is that the observer removes the write channels (`&CallInputs` instead of `&mut CallInputs`, no `Option<CallOutcome>` return). What that buys is that a *hand-written* read-only hook cannot intervene. Three things make that thinner than it looks:

1. **The guarantee is already partial.** `step`, `step_end`, and the `&mut MegaContext` argument stay mutable on both traits — as the module docs say, that half is contractual. An observer that mutates interpreter or context state changes results regardless.

2. **The outer EVM already takes the opposite position.** `MegaEvm::with_inspector` / `create_executor_with_inspector` accept any `Inspector` with no read-only variant, and `inspect_frame_init` honours an override outright ("Inspector intercepted — `frame_init()` is skipped entirely"). That inspector can rewrite every frame of every transaction in canonical block execution. A read-only channel only for the sandbox applies a stricter policy to the inner EVM than to the outer one, which is the wrong way round — the outer hook can do strictly more damage.

3. **It costs per-frame work on the path the feature exists for.** The blanket impl enforces read-only-ness by cloning `CallInputs` / `CallOutcome` and discarding the copy (`observer.rs:367,379,385,396`), and `SandboxTracer` repeats the same dance (`trace.rs:279-311`) because the observer hands it `&` while `TracingInspector` wants `&mut`. On one trait it forwards directly and both sets of clones disappear.

Concretely:

- Delete the trait and its blanket impl (`observer.rs:209-411`) and `ReadOnlyHook` (`412-522`).
- Keep `SandboxStartInfo`, `OuterCallInfo`, `SandboxEndOutcome`, `SandboxCompletionKind`, `SandboxRejectKind` — fold them into `inspector.rs` and drop `observer.rs`.
- Collapse to one setter, which can then be named for what it is: `set_keyless_sandbox_hook` / `clear_keyless_sandbox_hook`. The `ExtEnvs: 'static` bound that only the observer setter carries (`context.rs:517`) goes with it, so that "Notes for reviewers" item disappears too.
- `SandboxTracer` implements `SandboxInspector` and returns `None` from `call` / `create`.

Read-only then means what it means in revm: your hooks return `None`, which is the trait's default body anyway. The exclusivity rule, the "attaching one replaces the other" semantics, and the aliasing hazard between the two channels all stop being things a reader has to learn.

This is public API that mega-reth and mega-e2e already track, so collapsing it after a released tag is a breaking change for them — I'd rather it not ship in this shape. The change is mechanical (about 310 lines deleted, and the ~34 `set_keyless_sandbox_observer` call sites are almost all tests), so I'm happy to turn a follow-up push around quickly.

Troublor (changes requested):

vincent-k2026 (commented):
Read this at `7041fecf` after @flyq's 15 findings and @Troublor's change request. Both rounds look closed to me, so this is mostly confirmation plus two things neither round covered.

## Confirmed closed

`sandbox/observer.rs` is gone — the directory is now `error/execution/inspector/mod/state/state_merge/trace/tx`, so `SandboxObserver`, its blanket impl and `ReadOnlyHook` are actually deleted rather than deprecated, and there is one setter. That is @Troublor's request in full.

flyq's finding 5 (lifecycle and opcode hooks on different env slots) is structurally closed rather than closed by convention: `set_keyless_sandbox_hook` stores one `Rc` into both slots (`context.rs:519-522`), so `SandboxHookHandle<ExtEnvs>` and `SandboxHookHandle<EmptyExternalEnv>` are two views of the same `RefCell` and `notify_sandbox_start`/`_end` (`execution.rs:816,824-825`) cannot land on a different impl than the bridge does.

I also went looking for a `BorrowMutError` on the nested path, since flyq hit one in the aliasing scenario and there is now a `keyless_sandbox_nested_keyless_call_rex{5,6}` fixture. It is not reachable: every `InspectorBridge` hook is a single-expression borrow (`inspector.rs:434-501`), so the guard is dropped before revm acts on the return value and before an inner `KeylessDeploy` interception can start a nested sandbox. Same for the two lifecycle notifications. Worth stating explicitly since two slots sharing one `RefCell` plus nested sandboxes reads like a re-entrancy hazard and isn't one.

## `with_external_envs` drops an attached hook silently in release

`context.rs:389-392` guards the reset with `debug_assert!`, then unconditionally sets both slots to `None`. The reason given ("Hook is parameterized by `ExtEnvs`; a type change cannot keep it") is sound, but the handling splits by build profile: debug panics, release silently detaches. `test_with_external_envs_panics_in_debug_when_hook_is_attached` is `#[cfg(debug_assertions)]`, which acknowledges that.

The failure mode in release is the one this PR exists to prevent: an integrator attaches a `SandboxTracer`, calls `with_external_envs` afterwards, and gets a node that traces nothing, with no log line to explain it. `with_external_envs` is public, takes `self` and returns the new type, so it reads as a builder step that can sit anywhere in the chain — and there is already an internal call site at `context.rs:910`.

Cheapest fix is `assert!` instead of `debug_assert!`: it is a programming error either way, and failing the same way in both profiles is better than a silent no-op on the observability path. An `error!` would do if a panic is too strong. Making it unrepresentable (a type state that cannot hold a hook before the envs are wired) is nicer but not worth the churn here.

## Question: why does `MegaBlockExecutor` carry the setter?

Contract item 3 says the parent frame records post-intervention gas and usage as-is and "does not check conservation", and item 4 says the hook is node-local and non-consensus. Those two together mean an intervening hook attached to a block executor changes canonical execution output with nothing to catch it.

I take @Troublor's symmetry argument — the outer `Inspector` can already rewrite every frame via `create_executor_with_inspector`, so a stricter policy for the inner EVM would be backwards. So this isn't a request to remove it. But since the block-level forwarder is new surface in this PR and the tests exercise it (`SandboxTracer` across a three-transaction block), it's worth recording the reasoning: is `debug_traceBlock` the intended consumer, and was gating the setter behind `inspectors` or a tracing-only executor type considered? Whatever the answer, it belongs next to contract items 3 and 4 rather than in a review thread.

## Deferred findings need issue numbers

The description lists flyq's 9 (`AddressMismatch` hoist ordering vs a journal checkpoint), 13 (`notify_sandbox_end` pairing at seven sites) and 15 (test helper duplication) as "deferred to follow-ups", with no links. 9 is the one that matters — it is the guard standing between a `create` short-circuit and applied state in the parent journal, currently held by statement order. Please file them and reference the issues here, otherwise "deferred" is indistinguishable from dropped once this merges and mega-reth pins the tag.

## What I did not do

Static reading of the PR head and the two prior rounds; I did not build, run the suite, or reproduce anything. flyq's round was reproduced locally with throwaway tests, so treat that as the load-bearing verification and this as a pass over what changed since.

Troublor (approved):

flyq (approved):

PR #371, opened 2026-09-03
fix(chain): schedule Rex6 on the canonical mainnet and testnet hardfork schedules

RealiCZ:
## Summary

Schedule Rex6 on the canonical mainnet / testnet hardfork schedules in `block/chain.rs`, so `mega-evme replay` (which resolves the spec through `hardfork_schedule`) stops applying Rex5 rules to blocks after the activation timestamps; drop the stale "Rex6 is unscheduled" prose; and fix a wrong module-doc sentence in `block/limit.rs` about deposit DA accounting.

## Why

Both networks have been executing Rex6 since the published timestamps — testnet `1786330800` (2026-08-10), mainnet `1787626800` (2026-08-25); the on-chain `SequencerRegistry` (`0x6342…0006`) reports `version() == "2.0.0"` and `minRotationDelay() == 21600` on both. #364 published the timestamps in `docs/`, but `chain.rs` still pinned Rex6 as `Never` and a test asserted it, so every replay of a post-activation block ran the wrong spec. Consensus is unaffected (mega-reth derives forks from genesis); this is replay/tooling correctness.

Found by the Fenz.AI 2026-09-02 audit (F-016 / `ME-FE-20260902-016`, F-023 / `ME-FE-20260902-023`).

## Changes

- `block/chain.rs`: `Rex6` entries for mainnet and testnet, each with `SequencerRegistryRex6Config { rex6_min_rotation_delay: 21600 }` (the governance value the operator genesis attaches; readable live via `minRotationDelay()`). Module doc mentions the Rex6 params.
- Tests: `test_mainnet_/test_testnet_schedule_resolves_specs_by_timestamp` and `test_schedule_dispatch_by_chain_id` pin the Rex5→Rex6 boundary; `test_canonical_schedules_do_not_activate_rex6_or_rex7` becomes `test_canonical_schedules_activate_rex6_and_pin_rex7_never` (Rex6 at the published timestamps with delay 21600, Rex7 still `Never`, Rex6 terminal).
- `AGENTS.md`: Rex6 is frozen **and** activated; the "frozen ≠ scheduled" rule stays, without naming Rex6 as the unscheduled example.
- `block/limit.rs`: module doc said deposit DA sizes are "still tracked and accumulated in block DA usage counters"; `post_execution_update` skips deposits, so the doc now says so.

No behaviour change for any spec; no `docs/` change (already published in #364).

## Verification

- `cargo test -p mega-evm --lib` — 262 passed
- `cargo test -p mega-evm --test mutation -- hardfork_schedule` — 2 passed
- `cargo clippy --workspace --lib --examples --tests --benches --all-features --locked`, `cargo fmt --all --check` — clean

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -64,8 +64,8 @@ Progression: `EQUIVALENCE` → `MINI_REX` → `MINI_REX_1` → `MINI_REX_2` → 
   - _`REX7` is the current unstable spec under active development._
     When a new spec is introduced, this line should be updated to indicate the unstable spec.
   - Frozen and activated are separate properties.
-    `REX6` is frozen and activated on both networks (testnet `1786330800`, mainnet `1787626800`); `REX7` has no activation timestamp yet.
-    Freezing forbids further semantic change; scheduling is a later, separate decision, recorded in `block/chain.rs`.
+    `REX6` is frozen and activated on both networks (see `block/chain.rs`); `REX7` has no activation timestamp yet.
+    Freezing forbids further semantic change; scheduling is a later, separate decision made in the node chainspecs and mirrored here in `block/chain.rs` for replay tooling.
   - Specifications of each behavior-introducing spec can be found in the upgrade pages under `docs/spec/upgrades/`; alias rungs have no page of their own and are recorded in the upgrade overview and `docs/spec/hardfork-spec.md`.
 - **Hardfork** (`MegaHardfork`) defines network upgrade events (when specs activate).
   Every hardfork schedules a spec rung of its own — the fork→spec mapping is 1:1.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -64,8 +64,8 @@ Progression: `EQUIVALENCE` → `MINI_REX` → `REX` → `REX1` → `REX2` → `R
   - _`REX7` is the current unstable spec under active development._
     When a new spec is introduced, this line should be updated to indicate the unstable spec.
   - Frozen and activated are separate properties.
-    `REX6` is frozen but has no activation timestamp on mainnet or testnet, so both chains still execute `REX5`.
-    Freezing forbids further semantic change; scheduling is a later, separate decision.
+    `REX6` is frozen and activated on both networks (testnet `1786330800`, mainnet `1787626800`); `REX7` has no activation timestamp yet.
+    Freezing forbids further semantic change; scheduling is a later, separate decision, recorded in `block/chain.rs`.
   - Specifications of each spec can be found in the upgrade pages under `docs/spec/upgrades/`.
 - **Hardfork** (`MegaHardfork`) defines network upgrade events (when specs activate).
   Multiple hardforks can map to one spec.
@@ -307,7 +307,7 @@ When the agent is requested to implement a new feature or bug fix, it should con
   If editing nearby tests in the same module, align names to the same `test_` style when reasonable.
 - **Do NOT modify behavior for existing stable specs.**
   All specs through `REX6` are frozen; `REX7` is the unstable spec under active development.
-  A spec being frozen is independent of whether any network has scheduled it — `REX6` is frozen and unscheduled, and is still off-limits to behavior changes.
+  A spec being frozen is independent of whether any network has scheduled it — a frozen spec is off-limits to behavior changes whether or not a timestamp has been published.
   New EVM behavior, gas cost changes, or opcode modifications for stable specs **must** introduce a new spec and be gated with `spec.is_enabled(MegaSpecId::NEW_SPEC)`.
   Never change what an existing stable spec does.
 - **System contract changes require a new spec.**
--- a/crates/mega-evm/src/block/chain.rs
+++ b/crates/mega-evm/src/block/chain.rs
@@ -5,8 +5,9 @@
 //! `mega-evme` replay command, Foundry-based tooling, integration tests) should
 //! use [`hardfork_schedule`] rather than re-declaring timestamps locally.
 //!
-//! Per-fork parameters that are chain-specific data — currently the
-//! [`SequencerRegistryConfig`] seeded at Rex5 activation — are attached here too.
+//! Per-fork parameters that are chain-specific data — the
+//! [`SequencerRegistryConfig`] seeded at Rex5 activation and the
+//! [`SequencerRegistryRex6Config`] seeded at Rex6 activation — are attached here too.
 
 use alloy_hardforks::ForkCondition;
 use alloy_primitives::address;
@@ -34,6 +35,7 @@ pub fn mainnet_hardforks() -> MegaHardforkConfig {
         .with(MegaHardfork::Rex3, ForkCondition::Timestamp(1771639200))
         .with(MegaHardfork::Rex4, ForkCondition::Timestamp(1776659200))
         .with(MegaHardfork::Rex5, ForkCondition::Timestamp(1780632000))
+        .with(MegaHardfork::Rex6, ForkCondition::Timestamp(1787626800))
         // Seeded values read from the on-chain SequencerRegistry (0x6342…0006) at
         // Rex5 init: INITIAL_SEQUENCER (slot 5) and the admin (slot 2). At replay
         // time these only satisfy the activation guard — the live system address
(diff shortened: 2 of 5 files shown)


crates/mega-evm/src/block/limit.rs
```diff
@@ -107,8 +107,9 @@
 //!   block-level Data Availability size limit checks during pre-execution validation
 //! - **Rationale**: Deposit transactions are trustless L1→L2 messages that cannot be censored. They
 //!   must be included in blocks regardless of their DA size to maintain bridge integrity
-//! - **Tracking**: While exempt from limit checks, deposit DA sizes are still tracked and
-//!   accumulated in block DA usage counters for monitoring purposes
+//! - **Tracking**: Deposit DA sizes are **not** accumulated into the block DA usage counter
+//!   (`post_execution_update` skips them), so `block_da_size_used` reflects non-deposit
+//!   transactions only; operators estimating L1 posting cost must add deposits separately
```
flyq: **correctness** · `CONFIRMED`

The rewritten deposit-DA bullet is correct up to the semicolon, but the added closing clause — "operators estimating L1 posting cost must add deposits separately" — is an editorial claim the code does not support, and it runs backwards for L1-derived deposits.

<details>
<summary>Failure scenario</summary>

An operator sizing an L1 batch reads this module doc, adds the block's deposit DA sizes on top of `block_da_size_used`, and double-counts: OP-Stack L1→L2 deposits are derived from L1 by the node, not posted by the batcher, so they contribute zero L1 posting cost.

The code's own rationale is narrower and correct — `limit.rs:989-990` says the block DA limit is "only appliable for non-deposit transactions", i.e. the counter skips deposits because the *limit* exempts them. `block_da_size_used` is the budget tracker for the block-level DA limit, not an L1 cost estimate.

The one class where the answer is genuinely unclear is not the class this bullet describes: sequencer-originated system transactions are promoted to deposit type (`system/tx.rs:94`, invariant asserted at `:187`), so `post_execution_update` counts them as deposits — but whether those synthetic deposits reach L1 depends on MegaETH's derivation, which is not decidable from this repo. That is precisely why a claim about L1 posting cost does not belong in this module doc.

**Ask:** delete the clause after the semicolon. The sentence is complete and correct without it — it already states the fact (`block_da_size_used` excludes deposits) and the mechanism (`post_execution_update` skips them).

</details>

crates/mega-evm/src/block/chain.rs
```diff
@@ -34,6 +35,7 @@ pub fn mainnet_hardforks() -> MegaHardforkConfig {
         .with(MegaHardfork::Rex3, ForkCondition::Timestamp(1771639200))
         .with(MegaHardfork::Rex4, ForkCondition::Timestamp(1776659200))
         .with(MegaHardfork::Rex5, ForkCondition::Timestamp(1780632000))
+        .with(MegaHardfork::Rex6, ForkCondition::Timestamp(1787626800))
```
flyq, line 46: **test-coverage** · `CONFIRMED`

The new Rex6 timestamps are typed twice in the same file with nothing tying them to the published schedule, and the `rex6_min_rotation_delay: 21600` entry plus its four-line comment is copy-pasted between the two schedules with no shared constant.

<details>
<summary>Failure scenario</summary>

**Unpinned literals.** `chain.rs:38` / `:66` and the test literals at `:187` are the same numbers typed twice in the same file, so a mistyped digit is self-consistent: change `1787626800` to `1787626000` in both places and `cargo test -p mega-evm --lib` stays green, while every `mega-evme replay` of a mainnet block in that 800-second window silently resolves REX5 instead of REX6 and a replay of the activation block seeds the wrong `_minRotationDelay`. This is the same convention the existing Rex5 entries use, so it is not a regression — but the values now also live in `docs/spec/upgrades/overview.md:135,138` and in `AGENTS.md:67`, three unsynchronized copies of a number that `AGENTS.md:83` declares this file the single source of truth for.

*Optional ask:* a test that reads the two timestamps out of `docs/spec/upgrades/overview.md` and asserts they equal the schedule entries. (Note `crates/mega-evm/tests/compute_gas/claims.rs` from #358 asserts *rules* with hand-copied constants and does not itself parse `docs/`, so this would be a new mechanism rather than an existing one to reuse.) Accepting the convention is also defensible.

**Duplicated delay.** `chain.rs:47-51` and `:75-79` are identical down to the wording. The Rex5 blocks above them duplicate their comment too, but their *values* genuinely differ per chain, which justifies the split; here both the value and the prose are the same.

**Ask:** hoist a named `const REX6_MIN_ROTATION_DELAY: u64 = 21_600;` carrying the doc comment once, and let both schedules reference it. If governance ever sets a different delay on one network, that turns a silent divergence into a deliberate edit.

</details>

AGENTS.md
```diff
@@ -64,8 +64,8 @@ Progression: `EQUIVALENCE` → `MINI_REX` → `MINI_REX_1` → `MINI_REX_2` →
   - _`REX7` is the current unstable spec under active development._
     When a new spec is introduced, this line should be updated to indicate the unstable spec.
   - Frozen and activated are separate properties.
-    `REX6` is frozen but has no activation timestamp on mainnet or testnet, so both chains still execute `REX5`.
-    Freezing forbids further semantic change; scheduling is a later, separate decision.
+    `REX6` is frozen and activated on both networks (testnet `1786330800`, mainnet `1787626800`); `REX7` has no activation timestamp yet.
```
flyq: **documentation** · `CONFIRMED`

Two nits in the rewritten bullet: it hard-codes the activation timestamps into agent-facing instructions, and "recorded in `block/chain.rs`" reads as more authority than that file has.

<details>
<summary>Failure scenario</summary>

The rewrite is right on the substance — Rex6 is scheduled now, so the old text had to go.

**Third unpinned copy of the timestamps.** Line 67 joins `chain.rs:38,66` and `overview.md:135,138`, a few bullets above `:83`'s "`block/chain.rs` is the single source of truth … look there to find or change when a fork activates". Agent-facing instructions that hard-code a schedule are the copy most likely to go stale unnoticed, which is the drift that SSOT rule exists to prevent. *Ask:* drop the literals and point at the file — "`REX6` is frozen and activated on both networks (see `block/chain.rs`); `REX7` has no activation timestamp yet" carries the same information with nothing to keep in sync.

**"scheduling … recorded in `block/chain.rs`" (`:68`).** "Recorded" is accurate for this repo, and `:83` already frames the file correctly, but a reader arriving at `:68` first can take it to mean that scheduling a fork *is* editing `chain.rs`. It is not: a network's schedule lives in the node chainspecs (mega-reth's `crates/megaeth/chainspec`, mega-kona's fork registry) and this table mirrors it for replay tooling — as this PR's own description says. *Ask:* "…is a later, separate decision made in the node chainspecs and mirrored here in `block/chain.rs` for replay tooling."

*Not a finding, noted because it looks like one:* `:66-68` now illustrates "frozen and activated are separate properties" with two cases where the properties agree (Rex6 frozen *and* activated, Rex7 unstable *and* unscheduled), so read alone the bullet no longer demonstrates its own rule. The second `AGENTS.md` hunk handles that correctly — `:314` drops the now-false Rex6 example for the general statement — so no change is needed.

</details>

flyq (commented):
## Code review — Rex6 canonical schedule entries

Reviewed at `2b0e97544671` · xhigh effort · 3 findings

Reviewed at `2b0e9754` (the merge of `main` at #362; the PR's own diff against the new base is unchanged at 3 files, +50/−16). Line numbers below are from the post-change files at that commit.

The schedule change is correct and should merge. Verified: the two timestamps and the `21600` delay match `docs/spec/upgrades/overview.md:135,138` and the canonical fork registry; the new tests exercise the boundary rather than restating the entry (`chain.rs:147-148`, `:162-163`, `:193-195`); the committed replay fixture (`replay_offline.cache.json`, chain `4326` at `1780969472`) sits between mainnet Rex5 and Rex6 and still resolves REX5, so no re-capture is needed; and the replay-only claim holds — `hardfork_schedule` has no callers in `mega-reth`. `cargo test -p mega-evm --lib` passes at the merge commit (288 passed, including all 6 `block::chain` tests).

One finding is worth fixing before merge; the other two are cleanup.

### Inline findings

| # | Where | Finding |
|---|---|---|
| 1 | `crates/mega-evm/src/block/limit.rs:112` | New L1-posting-cost clause is unsupported and backwards |
| 2 | `crates/mega-evm/src/block/chain.rs:38` | Rex6 literals unpinned; 21600 duplicated verbatim |
| 3 | `AGENTS.md:67` | Bullet hard-codes timestamps, overstates chain.rs |

flyq (approved):

Troublor (approved):

